PR #1518, opened 2023-07-12
added room.isChat

gitsper:
Adds a way for library users to check if a room is a chat room. Without this, chat clients might list non-chat rooms.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -332,6 +332,16 @@ class Room {
     });
   }
 
+  /// Checks if the `m.room.create` state has a `type` with no value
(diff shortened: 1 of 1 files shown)


nico-famedly:
We discussed this internally and we think this is better solved in each application (by writing an extension for example to provide this function). Specifically we have rooms that have a room type, but are still used for chatting and the other way around, as such this would be misleading. We do however consider providing easier access to the roomtype by just having a getter for that, which would simplify your logic to just checking for an empty room type in your application instead of checking for isChat.

Thank you for the effort however and sorry it took us so long to get back to you!

PR #1534, opened 2023-07-31
refactor: simplify UIA stage selection logic

nico-famedly:
(Also prevents a potential out of bounds exception)

### DIFF
--- a/lib/src/utils/uia_request.dart
+++ b/lib/src/utils/uia_request.dart
@@ -102,17 +102,11 @@ class UiaRequest<T> {
       List<AuthenticationFlow> flows, List<String> completed) {
     final nextStages = <String>{};
     for (final flow in flows) {
-      final stages = flow.stages;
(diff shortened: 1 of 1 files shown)


lib/src/utils/uia_request.dart
```diff
@@ -102,18 +102,8 @@ class UiaRequest<T> {
       List<AuthenticationFlow> flows, List<String> completed) {
     final nextStages = <String>{};
     for (final flow in flows) {
-      final stages = flow.stages;
-      final nextStage = stages[completed.length];
-      var stagesValid = true;
-      for (var i = 0; i < completed.length; i++) {
```
nico-famedly, line 108: Actually, this does a prefix check as well, which I removed, which might be bad, since the stages are stored in a set, which reorders them? But then a client might pick the wrong flow just to end up in one it can't do...

PR #1537, opened 2023-08-02
fix: fixed hardcoded historyCount

PhantomRay:
Hi there
Found a bug in _requestEvents. `historyCount` is supposed to be applied to `limit`, however it is using Room.DefaultHistoryCount.

Also, is it possible to filter event types? Otherwise, when retrieving historical chat, there might be events not supposed to be shown.

### DIFF
--- a/lib/src/timeline.dart
+++ b/lib/src/timeline.dart
@@ -116,7 +116,7 @@ class Timeline {
           : await room.client.database?.getEventList(
               room,
               start: events.length,
-              limit: Room.defaultHistoryCount,
+              limit: historyCount,
             );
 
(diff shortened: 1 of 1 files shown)


krille-chan (approved):
Thank you so much 👍👍👍

krille-chan:
Hey, you must use conventional commits for this repository

nico-famedly:
Our CI requires commit messages to be formatted as conventional commits. We would really like to merge your contribution, however this requires the CI to pass. Can you rewrite your commit message to use a lower-case "fix" so that we can merge this? Thank you very much for the contribution though and I hope we can merge this soon!

PR #1545, opened 2023-08-07
feature: Delay onCallCandidates to execute before createNewCall

CristianKoh1:


nico-famedly:
So for us to proceed with this PR, I think we need to clear a few hurdles:

1. Contributions here require you to have signed the CLA. Can you click on the link above and do so, if you want to continue with this PR? If you don't want to, that is fine, but please give us a heads-up, so that we can close this.
2. After that I would like to also understand the reasoning behind this change. Usually adding a delay to something to fix race conditions is not how we want to solve a problem and we prefer to explicitly wait for specific events to happen instead. Could you provide an explanation, what exactly is fixed by this and if this can't be solved differently?

But in general thank you very much for your contribution. As you might have noticed our call code is not as stable as we would like and as such we very much appreciate any improvements in that area. Sorry that it took us so long to get back to you, but there has been a lot of stuff going on lately!

td-famedly:
Looks abandoned closing since, calls rewrite soon

PR #1547, opened 2023-08-13
feat: Add option to not cache users in memory when requesting all of …

krille-chan:
…a room

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1535,13 +1535,16 @@ class Room {
   /// List `membershipFilter` defines with what membership do you want the
   /// participants, default set to
   /// [[Membership.join, Membership.invite, Membership.knock]]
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -1535,13 +1535,16 @@ class Room {
...
   Future<List<User>> requestParticipants(
       [List<Membership> membershipFilter = const [
         Membership.join,
         Membership.invite,
         Membership.knock,
       ],
-      bool suppressWarning = false]) async {
+      bool suppressWarning = false,
+      bool cache = true]) async {
     if (!participantListComplete && partial) {
```
mErrenst, line 1548: Is the indentation here correct? It seems to me like this should be more indented? Wouldn't be different than before but looks wrong to me

mErrenst (approved):

PR #1548, opened 2023-08-23
build: Bump version to 0.22.3

krille-chan:


pubspec.yaml
```diff
@@ -39,7 +39,6 @@ dev_dependencies:
   lints: ^2.0.0
   test: ^1.15.7
   #flutter_test: {sdk: flutter}
-
```
TheOneWithTheBraid, line 42: That does not look like an intentional change.

TheOneWithTheBraid (approved):

pubspec.yaml
krille-chan: Hm, looks like vs code is doing this automatically with the formatting plugin. So I would tend to just leave it in if it's okay for you :D

TheOneWithTheBraid: Sounds good, I approved, just wanted to point out on it.

PR #1550, opened 2023-09-04
fix: startDirectChat might return an unjoined room

nico-famedly:
If you create DM room, but the other party doesn't join it, they might be unable to create a new DM using startDirectChat. Since creating a new DM is pretty much the same as joining the invite, we try to join the pending DM invite first and fall back to creating a new room if that fails.


mErrenst (approved):
LGTM, thanks for fixing this! <3

PR #1553, opened 2023-09-11
fix: Wait for rate limit to pass while sending messages

nico-famedly:
Otherwise the user can't handle that exception themselves, since we don't rethrow.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1100,7 +1100,16 @@ class Room {
           txid: messageID,
         );
       } catch (e, s) {
-        if (e is MatrixException || DateTime.now().isAfter(timeoutDate)) {
+        if (e is MatrixException &&
+            e.retryAfterMs != null &&
+            !DateTime.now()
+                .add(Duration(milliseconds: e.retryAfterMs!))
+                .isAfter(timeoutDate)) {
+          Logs().w(
+              'Ratelimited while sending message, waiting for ${e.retryAfterMs}ms');
+          await Future.delayed(Duration(milliseconds: e.retryAfterMs!));
+        } else if (e is MatrixException ||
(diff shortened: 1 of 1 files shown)


techno-disaster (dismissed):

lib/src/room.dart
```diff
@@ -1100,17 +1100,26 @@ class Room {
           txid: messageID,
         );
       } catch (e, s) {
-        if (e is MatrixException || DateTime.now().isAfter(timeoutDate)) {
+        if (e is MatrixException &&
+            e.retryAfterMs != null &&
+            !DateTime.now().isAfter(timeoutDate)) {
+          Logs().w(
+              'Ratelimited while sending message, waiting for ${e.retryAfterMs}ms');
+          await Future.delayed(Duration(milliseconds: e.retryAfterMs!));
```
krille-chan, line 1108: But after waiting this is not resending the message right? Because it will not go into the other else if block where it tries to send it again but just go to the end of the method if I see this correctly

krille-chan (changes requested):

lib/src/room.dart
nico-famedly: the whole method is wrapped in a while true already, the else branch retries as well, afaik? Did I misread this? :D

krille-chan: Sorry, yes you are right. Haven't expanded the focus. What happens when the `e.retryAfterMs` is larger than the `timeoutDate`?

nico-famedly: I guess then it waits for a too long period and then fails. I guess that could be prettier, if we just immediately fail in that case? However that then again quietly doesn't send the event, so maybe we should propagate the exceptions at some point?

krille-chan (approved):

PR #1557, opened 2023-09-14
fix: storing the end of history pagination

nico-famedly:
At least in our CI this throws a null assertion error and since we explicitly expect this to sometimes be null in our code, we should store it as such.


nico-famedly:
See https://github.com/famedly/app/actions/runs/6182935067/job/16794025458#step:4:2031

krille-chan (approved):
Sounds reasonable 👍 thanks

PR #1558, opened 2023-09-15
Apply state event before decryption on leaved room

mrmcoding88:
jira-132

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -947,6 +947,21 @@ class Client extends MatrixApi {
                         .toList() ??
                     []));
 
+        leftRoom.prev_batch = room.timeline?.prevBatch;
+        room.state?.forEach((event) {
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -975,6 +962,19 @@ class Client extends MatrixApi {
           ));
         });
```
nico-famedly, line 963: I only noticed it now, but the room.state.forEach needs to be after the room.timeline.events.forEach for the setState, since the .state is the state events from before the timeline in the spec. Can you fix that too?

nico-famedly (changes requested):

nico-famedly (dismissed):

nico-famedly:
Thank you :)

nico-famedly:
fails formatting though, you need to run dart format on it :D

nico-famedly (approved):

PR #1560, opened 2023-09-18
fix: Search event not return when result not found

drminh2807:
- Fix return empty result if result not found
- Fix duplicate events when first load online event


hoangdat (approved):

nico-famedly:
Hi there, sorry it has taken us some time to respond to this. From your description it is not quite clear to me, what the issue is, that you are fixing. Could you expand on that a little bit?

It would also be very useful if you could write a test for this issue to make sure we don't break it again in the future. Please ask questions, if you need help with that. This would also work as an explanation for what issue you encountered.

I'll try to get back to you asap, once you have resolved those 2 concerns!

drminh2807:
@nico-famedly 
Number 1: when not found any event, you still need `yield found` in the last, so I can know it empty and show Empty UI. Hope it easy to understand
Number 2: Duplicate events when first load online event
- `sinceEventId` should be the last of local or database eventId
- Should not include `afterEvents` in `EventContext`, because it already in the database
IMG_6878

I will write tests when I have free time

nico-famedly:
Okay, number 1 makes sense to me now. For number 2 I guess it is about that when searching, we shouldn't query events from the server again, that we already have locally, since we already searched those, so we start searching from the last event in the db to skip those? I think your logic is a bit weird in that case.

Specifically if the sinceEventId is null, we shouldn't be yielding duplicate events, since we are only searching from the prevBatch token (unless the prevBatch token is empty!). So that only leaves the case, where we have an eventId to start the search from. In that case we should make sure on the first search iteration to not yield duplicate events, since the context will have returned events before the startEventId as well. I guess that is a bit tricky, I'll try to leave some inline comments.

lib/src/timeline.dart
```diff
@@ -590,9 +591,15 @@ class Timeline {
...
+    var ignoreEventIds = <String>[];
     if (sinceEventId != null) {
-      prevBatch =
-          (await room.client.getEventContext(room.id, sinceEventId)).end;
+      final eventContext =
+          await room.client.getEventContext(room.id, sinceEventId);
+      final eventIdsAfter =
+          eventContext.eventsAfter?.map((event) => event.eventId).toList() ??
+              [];
+      ignoreEventIds = eventIdsAfter..add(sinceEventId);
```
nico-famedly, line 601: I think this logic shouldn't be necessary, if you added all the yielded event ids to the `ignoreEventIds` already.

krille-chan:
This should now be fixed by using the prev_batch for the next search I geuss 🤔 please ping me if I'm wrong. Just use `timeline.startSearch` from now on

PR #1562, opened 2023-09-18
fix: wait for online key backup key to be cached on reset

nico-famedly:
Otherwise uploading the keys won't work, since the key backup is "disabled".

### DIFF
--- a/lib/encryption/utils/bootstrap.dart
+++ b/lib/encryption/utils/bootstrap.dart
@@ -584,12 +584,10 @@ class Bootstrap {
       Logs().v('Wait for secret to come down sync');
 
       if (!await encryption.keyManager.isCached()) {
-        await client.onSync.stream
-            .firstWhere((syncUpdate) =>
-                syncUpdate.accountData != null &&
-                syncUpdate.accountData!
-                    .any((accountData) => accountData.type == megolmKey))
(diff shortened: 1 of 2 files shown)


lib/encryption/utils/bootstrap.dart
```diff
@@ -581,6 +581,17 @@ class Bootstrap {
...
       await newSsssKey?.store(megolmKey, base64.encode(privKey));
+      Logs().v('Wait for secret to come down sync');
+
+      if (!await encryption.keyManager.isCached()) {
+        await client.onSync.stream
+            .firstWhere((syncUpdate) =>
+                syncUpdate.accountData != null &&
+                syncUpdate.accountData!
+                    .any((accountData) => accountData.type == megolmKey))
+            .then((_) => Logs().v('New backup key was created'));
```
krille-chan, line 592: Hm, I see no reason for the `.then()` here. Am I overseeing something? Just do the log command after the `await client.onSync.stream.firstWhere()` and it should do the same

nico-famedly: Fixed :)

It was mostly because I took that code block from elsewhere, where we await multiple account data changes and log for each of them.

krille-chan (approved):

PR #1564, opened 2023-09-19
feat: Implement member change type

krille-chan:
For easier handling and
comparing of m.room.member
state events to better
know what these event
actually changes.

Part of https://github.com/famedly/product-management/issues/496

### DIFF
--- a/lib/src/utils/event_localizations.dart
+++ b/lib/src/utils/event_localizations.dart
@@ -114,89 +114,59 @@ abstract class EventLocalizations {
       final userIsTarget = event.stateKey == event.room.client.userID;
       final userIsSender = event.senderId == event.room.client.userID;
 
-      // Fallback message if just nothing has changed:
-      var text = i18n.joinedTheChat(targetName);
-
-      // Has the membership changed?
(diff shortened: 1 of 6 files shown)


lib/src/utils/event_localizations.dart
```diff
@@ -114,89 +115,62 @@ abstract class EventLocalizations {
...
-        if (newAvatar != oldAvatar) {
-          text = i18n.changedTheProfileAvatar(targetName);
-        }
-        // Has the user displayname changed?
-        else if (newDisplayname != oldDisplayname && stateKey != null) {
-          text = i18n.changedTheDisplaynameTo(oldDisplayname, newDisplayname);
-        }
       }
-      return text;
+      return userIsTarget
```
mErrenst, line 171: Isn't that already in the default case above and would be done anyway?

mErrenst (dismissed):

lib/src/utils/event_localizations.dart
krille-chan: Fixed

mErrenst (approved):

PR #1566, opened 2023-09-20
refactor: Remove deprecated dart code metrics

krille-chan:


nico-famedly (approved):
Sad to see this go, but probably the right solution for now.

PR #1569, opened 2023-09-20
Updated package versions

ghost:
Updated all packages that had minor non-breaking versions to address dependency conflicts with packages used in the consuming app.  These were udpated:

webrtc_interface - minor (x.?.?)
olm - minor (x.?.?)
mime - minor (x.?.?)
matrix_api_lite - minor (x.?.?)
js - minor (x.?.?)
image - minor (x.?.?)
html - minor (x.?.?)
ffi - minor (x.?.?)
crypto - minor (x.?.?)
collection - minor (x.?.?)
blurhash_dart - minor (x.?.?)
async - minor (x.?.?)

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,28 +8,28 @@ environment:
   sdk: ">=3.0.0 <4.0.0"
 
 dependencies:
-  async: ^2.8.0
+  async: ^2.11.0
   base58check: ^2.0.0
-  blurhash_dart: ^1.1.0
-  canonical_json: ^1.1.0
-  collection: ^1.15.0
-  crypto: ^3.0.0
-  ffi: ^2.0.0
+  blurhash_dart: ^1.2.1
+  canonical_json: ^1.1.2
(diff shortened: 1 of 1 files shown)


nico-famedly:
This mostly looks good to me, however there a few issues, which would need to get resolved before we can merge this:

- The commit message should follow the conventional commit format. In this case just adding "chore: " as a prefix for your commit message would work.
- There seems to be a conflict when resolving the dependencies in the coverage job. Maybe we shouldn't update the minimum collection version yet, so that people using older dart or flutter versions can still resolve the dependencies without conflicts? Alternatively the flutter version in the workflow would need to be increased.
- It would be great if you could rebase this while you change the commit message, to resolve the current conflicts. That is my fault for not reacting earlier on this, however it would make merging this easier so I would appreciate it.

As always, sorry for my slow response times and thank you for picking this up!

krille-chan:
@rduffr ?

krille-chan:
CLosing this because of inactivity

PR #1571, opened 2023-09-22
chore: incrementally add left rooms to archive

TheOneWithTheBraid:
- store left rooms in archive during sync (as well as they are removed on join already)
- refactor room archive code
- fix typo

Internal reference: SMC-385

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -939,48 +939,58 @@ class Client extends MatrixApi {
     return _archivedRooms;
   }
 
-  Future<void> _storeArchivedRoom(String id, LeftRoomUpdate update) async {
-    final room = update;
-    final leftRoom = Room(
-      id: id,
-      membership: Membership.leave,
-      client: this,
-      roomAccountData:
-          room.accountData?.asMap().map((k, v) => MapEntry(v.type, v)) ??
+  /// [_storeArchivedRoom]
+  /// @leftRoom we can pass a room which was left so that we don't loose states
+  Future<void> _storeArchivedRoom(
+    String id,
+    LeftRoomUpdate update, {
+    Room? leftRoom,
+  }) async {
+    final roomUpdate = update;
+    final archivedRoom = leftRoom ??
+        Room(
+          id: id,
+          membership: Membership.leave,
+          client: this,
+          roomAccountData: roomUpdate.accountData
+                  ?.asMap()
+                  .map((k, v) => MapEntry(v.type, v)) ??
               <String, BasicRoomEvent>{},
-    );
-
+        );
+    // Set membership of room to leave, in the case we got a left room passed, otherwise
+    // the left room would have still membership join, which would be wrong for the setState later
+    archivedRoom.membership = Membership.leave;
     final timeline = Timeline(
-        room: leftRoom,
+        room: archivedRoom,
         chunk: TimelineChunk(
-            events: room.timeline?.events?.reversed
+            events: roomUpdate.timeline?.events?.reversed
                     .toList() // we display the event in the other sence
-                    .map((e) => Event.fromMatrixEvent(e, leftRoom))
+                    .map((e) => Event.fromMatrixEvent(e, archivedRoom))
                     .toList() ??
                 []));
 
-    leftRoom.prev_batch = room.timeline?.prevBatch;
-    room.state?.forEach((event) {
-      leftRoom.setState(Event.fromMatrixEvent(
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -927,58 +927,66 @@ class Client extends MatrixApi {
...
+        leftRoom,
+      ));
+    });
 
-        _archivedRooms.add(ArchivedRoom(room: leftRoom, timeline: timeline));
+    for (var i = 0; i < timeline.events.length; i++) {
+      // Try to decrypt encrypted events but don't update the database.
+      if (leftRoom.encrypted && leftRoom.client.encryptionEnabled) {
+        if (timeline.events[i].type == EventTypes.Encrypted) {
+          unawaited(
```
techno-disaster, line 975: Can you add a comment for why this is unawaited? (is not a very obvious change for td)

lib/src/client.dart
```diff
@@ -2161,9 +2169,14 @@ class Client extends MatrixApi {
       // stop stale group call checker for left room.
       room.stopStaleCallsChecker(room.id);
 
+      // in order to keep the archive in sync, add left room to archive
+      if (chatUpdate is LeftRoomUpdate) {
```
techno-disaster, line 2173: the `membership == Membership.leave` above should be enough

techno-disaster (changes requested):

lib/src/client.dart
nico-famedly: No, because then one would need a cast.

nico-famedly: Unless the type checker can see, that the only way for membership to be leave, is for it to be a LeftRoomUpdate? I don't think it usually sees that, but if it does, the if statement is unnecessary yes.

techno-disaster: in an ideal world it would, but doesn't look like it does rn. Thanks!

lib/src/client.dart
TheOneWithTheBraid: Hmmmm, I was unsure about that one. I didn't find it necessary to do this synchronously because of the number of changes (up to the `_handleRooms` method) but maybe yeah, might be better to await this. Correspondingly updated it.

techno-disaster (approved):

TheOneWithTheBraid:
It's now ready so far but maybe in regard of the comment I sent in [matrix] we might consider still waiting merging this up to the next weekly.

TheOneWithTheBraid:
Ready for review, @techno-disaster !

nico-famedly (approved):

nico-famedly:
Well, of course the archive room test fails now, since there will be archived rooms already in the test possibly :)

TheOneWithTheBraid:
> [CLA assistant check](https://cla-assistant.io/famedly/matrix-dart-sdk?pullRequest=1571) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/famedly/matrix-dart-sdk?pullRequest=1571) before we can accept your contribution.**1** out of **2** committers have signed the CLA.✅ TheOneWithTheBraid❌ Clemens-ToegelYou have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/famedly/matrix-dart-sdk?pullRequest=1571) it.

@Clemens-Toegel CLA required.

TheOneWithTheBraid:
@nico-famedly Could you start a pipeline ?

nico-famedly:
(Changes still fail the formatting check it seems)

lib/src/client.dart
```diff
@@ -2199,7 +2221,7 @@ class Client extends MatrixApi {
           requestHistoryOnLimitedTimeline) {
         Logs().v(
             'Limited timeline for ${rooms[roomIndex].id} request history now');
-        runInRoot(rooms[roomIndex].requestHistory);
+        await runInRoot(rooms[roomIndex].requestHistory);
```
nico-famedly, line 2224: Why is this getting awaited now? We probably want to do this in the background and not block the sync on a possibly slow history fetch?

nico-famedly, line 2224: Why is this getting awaited now? We probably want to do this in the background and not block the sync on a possibly slow history fetch?

nico-famedly (approved):

nico-famedly:
Finally. Never go on holiday, while a merge is still pending, I guess... Sorry about the delay!

PR #1584, opened 2023-10-17
fix: validate account_data values instead of checking them in syncUpdates

techno-disaster:
Fixes: https://github.com/famedly/product-management/issues/1401

fixes race where sometimes you miss the secrets in sync

edit from product: this also closes https://github.com/famedly/matrix-dart-sdk/pull/1584

### DIFF
--- a/lib/encryption/ssss.dart
+++ b/lib/encryption/ssss.dart
@@ -248,15 +248,16 @@ class SSSS {
     }()
         .firstWhere((keyId) => getKey(keyId) == null);
 
-    final accountDataType = EventTypes.secretStorageKey(keyId);
+    final accountDataTypeKeyId = EventTypes.secretStorageKey(keyId);
     // noooow we set the account data
-    final waitForAccountData = client.onSync.stream.firstWhere((syncUpdate) =>
-        syncUpdate.accountData != null &&
-        syncUpdate.accountData!
-            .any((accountData) => accountData.type == accountDataType));
+
     await client.setAccountData(
-        client.userID!, accountDataType, content.toJson());
-    await waitForAccountData;
+        client.userID!, accountDataTypeKeyId, content.toJson());
+
+    while (!client.accountData.containsKey(accountDataTypeKeyId)) {
+      Logs().v('Waiting accountData to have $accountDataTypeKeyId');
+      await client.oneShotSync();
+    }
 
     final key = open(keyId);
     await key.setPrivateKey(privateKey);
@@ -327,7 +328,7 @@ class SSSS {
     }
     final enc = encryptedContent.tryGetMap<String, Object?>(keyId);
     if (enc == null) {
-      throw Exception('Wrong / unknown key');
+      throw Exception('Wrong / unknown key: $type, $keyId');
     }
     final ciphertext = enc.tryGet<String>('ciphertext');
     final iv = enc.tryGet<String>('iv');
@@ -750,6 +751,14 @@ class OpenSSSS {
       throw Exception('SSSS not unlocked');
     }
     await ssss.store(type, secret, keyId, privateKey, add: add);
+    while (!ssss.client.accountData.containsKey(type) ||
+        !(ssss.client.accountData[type]!.content
+            .tryGetMap<String, Object?>('encrypted')!
+            .containsKey(keyId)) ||
+        await getStored(type) != secret) {
+      Logs().d('Wait for secret of $type to match in accountdata');
+      await ssss.client.oneShotSync();
+    }
   }
 
   Future<void> validateAndStripOtherKeys(String type, String secret) async {
--- a/lib/encryption/utils/bootstrap.dart
+++ b/lib/encryption/utils/bootstrap.dart
@@ -90,6 +90,8 @@ class Bootstrap {
 
   // cache the secret analyzing so that we don't drop stuff a different client sets during bootstrapping
   Map<String, Set<String>>? _secretsCache;
+
+  /// returns ssss from accountdata, eg: m.megolm_backup.v1, or your m.cross_signing stuff
   Map<String, Set<String>> analyzeSecrets() {
     final secretsCache = _secretsCache;
     if (secretsCache != null) {
@@ -292,12 +294,12 @@ class Bootstrap {
         }
         // alright, we re-encrypted all the secrets. We delete the dead weight only *after* we set our key to the default key
       }
-      final updatedAccountData = client.onSync.stream.firstWhere((syncUpdate) =>
-          syncUpdate.accountData != null &&
-          syncUpdate.accountData!.any((accountData) =>
-              accountData.type == EventTypes.SecretStorageDefaultKey));
       await encryption.ssss.setDefaultKeyId(newSsssKey!.keyId);
-      await updatedAccountData;
+      while (encryption.ssss.defaultKeyId != newSsssKey!.keyId) {
+        Logs().v(
+            'Waiting accountData to have the correct m.secret_storage.default_key');
+        await client.oneShotSync();
+      }
       if (oldSsssKeys != null) {
         for (final entry in secretMap!.entries) {
           Logs().v('Validate and stripe other keys ${entry.key}...');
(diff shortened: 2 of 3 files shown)


lib/encryption/utils/bootstrap.dart
```diff
@@ -581,14 +581,6 @@ class Bootstrap {
       );
       Logs().v('Store the secret...');
       await newSsssKey?.store(megolmKey, base64.encode(privKey));
```
nico-famedly, line 583: can you also update the other places?

techno-disaster: This was the only place with `store` where we had this (broken) logic, so don't need anything more afaik

krille-chan (approved):

lib/encryption/utils/bootstrap.dart
krille-chan: That's correct

nico-famedly: Then how is the logic in https://github.com/famedly/matrix-dart-sdk/blob/200c9a5ad23bdae4d3446fe19ed36f7262db4283/lib/encryption/utils/bootstrap.dart#L493 different?

lib/encryption/utils/bootstrap.dart
```diff
@@ -292,12 +292,15 @@ class Bootstrap {
...
       await encryption.ssss.setDefaultKeyId(newSsssKey!.keyId);
-      await updatedAccountData;
+      // how useful is this? wouldn't a old ssss be fine here anyway?
+      while (client.accountData
+          .where((_, BasicEvent event) =>
+              event.type == EventTypes.SecretStorageDefaultKey)
+          .isNotEmpty) {
+        Logs().v('Waiting accountData to have m.secret_storage.default_key');
+        await client.oneShotSync(timeout: Duration(seconds: 3));
+      }
```
nico-famedly, line 303: Why do you add the wait logic here, if you already added that to the store function? Also the store should maybe now happen in parallel like the wait did before?

lib/encryption/utils/bootstrap.dart
```diff
@@ -479,33 +482,20 @@ class Bootstrap {
...
-          client.onSync.stream
-              .firstWhere((syncUpdate) =>
-                  syncUpdate.accountData != null &&
-                  syncUpdate.accountData!
-                      .any((accountData) => accountData.type == entry.key))
-              .then((_) =>
-                  Logs().v('New Key with type ${entry.key} was created')),
-        );
         Logs().v('Store new SSSS key ${entry.key}...');
         await newSsssKey?.store(entry.key, entry.value);
```
nico-famedly, line 495: You now made the await sequential, why not still do it in parallel using Future.wait?

lib/encryption/utils/bootstrap.dart
techno-disaster: because it doesn't use the store function, atleast not for the defaultKeyId? `await encryption.ssss.setDefaultKeyId(newSsssKey!.keyId);`

lib/encryption/utils/bootstrap.dart
techno-disaster: ah I assumed doing it sequentially on `store` would cause all kinds of mishaps with db stuff. Previously the db was sequential but waiting for accountdata was parallel.

lib/encryption/ssss.dart
```diff
@@ -248,15 +248,16 @@ class SSSS {
...
-            .any((accountData) => accountData.type == accountDataType));
+
     await client.setAccountData(
-        client.userID!, accountDataType, content.toJson());
-    await waitForAccountData;
+        client.userID!, accountDataTypeKeyId, content.toJson());
+
+    while (!client.accountData.containsKey(accountDataTypeKeyId)) {
+      Logs().v('Waiting accountData to have $accountDataTypeKeyId');
+      await client.oneShotSync(timeout: Duration(seconds: 3));
```
nico-famedly, line 259: Does the timeout here even do anything?

lib/encryption/ssss.dart
```diff
@@ -750,6 +751,16 @@ class OpenSSSS {
...
     await ssss.store(type, secret, keyId, privateKey, add: add);
+
+    // ideally never happens unless new key?
+    while (!ssss.client.accountData.containsKey(type)) {
+      Logs().d('Wait for type $type');
+      await ssss.client.oneShotSync(timeout: Duration(seconds: 3));
+    }
+    while (await getStored(type) != secret) {
+      Logs().d('Wait for secret of $type to match');
+      await ssss.client.oneShotSync(timeout: Duration(seconds: 3));
```
nico-famedly, line 762: What would this second loop be for?

lib/encryption/utils/bootstrap.dart
nico-famedly: Imo you should still store all the secrets in parallel, so that you don't need to wait for 3 syncs sequentially.

lib/encryption/ssss.dart
td-famedly: you could already have the key but with the old secret, the first loop is for when the key was never there

lib/encryption/ssss.dart
td-famedly: Looks like I confused the sync timeout with min time to wait thanks

lib/encryption/ssss.dart
nico-famedly: So the first loop is redundant?

lib/encryption/utils/bootstrap.dart
```diff
@@ -479,33 +481,25 @@ class Bootstrap {
...
-                  Logs().v('New Key with type ${entry.key} was created')),
-        );
-        Logs().v('Store new SSSS key ${entry.key}...');
-        await newSsssKey?.store(entry.key, entry.value);
+      if (newSsssKey != null) {
+        final storeFutures = <Future<void>>[];
+        for (final entry in secretsToStore.entries) {
+          storeFutures.add(newSsssKey!.store(entry.key, entry.value));
+        }
+        Logs().v('Store new SSSS key entry...');
```
nico-famedly, line 497:
```suggestion
        Logs().v('Store new SSSS key entries...');
```

lib/encryption/utils/bootstrap.dart
```diff
@@ -479,33 +481,25 @@ class Bootstrap {
...
+        await Future.wait(storeFutures);
       }
+
       Logs().v(
           'Wait for MasterKey and ${secretsToStore.entries.length} keys to be created');
```
nico-famedly, line 502: This log entry is now wrong

lib/src/client.dart
```diff
@@ -1652,8 +1652,8 @@ class Client extends MatrixApi {
 
   /// Immediately start a sync and wait for completion.
   /// If there is an active sync already, wait for the active sync instead.
-  Future<void> oneShotSync() {
-    return _sync();
+  Future<void> oneShotSync({Duration? timeout}) {
+    return _sync(timeout: timeout);
```
nico-famedly, line 1656: This is unused now?

test/encryption/ssss_test.dart
```diff
@@ -107,9 +107,15 @@ void main() {
...
-      await handle.store('best animal', 'foxies');
+
+      // OpenSSSS store waits for accountdata to be updated before returning
+      // but we can't update that before the below endpoint is not hit.
+      await handle.ssss
+          .store('best animal', 'foxies', handle.keyId, handle.privateKey!);
       // alright, since we don't properly sync we will manually have to update
       // account_data for this test
+
+      //   await handle.store('best animal', 'foxies');
```
nico-famedly, line 118: Should this get removed now or what is this additional commented out code for? :3

lib/encryption/ssss.dart
td-famedly: Well `getStored` fails if there is no event with that key hence, wait for the key before checking the secret.

lib/src/client.dart
td-famedly: yes but I don't see why not to keep it

nico-famedly: We shouldn't keep dead code, it just adds friction later imo. If there is no usecase for this, it shouldn't be there. Especially since this is a different timeout than people expect, imo.

nico-famedly (approved):

lib/src/client.dart
td-famedly: hm oki

td-famedly:
brb something broke :p

PR #1585, opened 2023-10-19
refactor: Trigger upload keys on sync and not in background job and u…

krille-chan:
…pload them before logout

related to https://github.com/famedly/product-management/issues/1238

### DIFF
--- a/lib/encryption/key_manager.dart
+++ b/lib/encryption/key_manager.dart
@@ -747,15 +747,34 @@ class KeyManager {
     }
   }
 
-  bool _isUploadingKeys = false;
+  Future<void>? _uploadingFuture;
 
-  Future<void> backgroundTasks() async {
+  void startAutoUploadKeys() {
+    _uploadKeysOnSync = encryption.client.onSync.stream
+        .listen((_) => uploadInboundGroupSessions(skipIfInProgress: true));
+  }
+
+  /// This task should be performed after sync processing but should not block
+  /// the sync. To make sure that it never gets executed multiple times, it is
+  /// skipped when an upload task is already in progress. Set `skipIfInProgress`
+  /// to `false` to await the pending upload task instead.
+  Future<void> uploadInboundGroupSessions(
+      {bool skipIfInProgress = false}) async {
     final database = client.database;
     final userID = client.userID;
-    if (_isUploadingKeys || database == null || userID == null) {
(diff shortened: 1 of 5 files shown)


lib/encryption/key_manager.dart
```diff
@@ -749,13 +749,23 @@ class KeyManager {
...
+  /// the sync. To make sure that it never gets executed multiple times, it
+  /// is skipped if the [_isUploadingKeys] flag is set and therefore there is
+  /// already an ongoing upload task.
+  Future<void> uploadInboundGroupSessions() async {
     final database = client.database;
     final userID = client.userID;
     if (_isUploadingKeys || database == null || userID == null) {
       return;
     }
     _isUploadingKeys = true;
```
nico-famedly, line 767: Can you instead assign to a future local variable and await that here? Otherwise, if the upload is already in progress from sync processing, we will not be waiting for it to complete, I think, since we just return immediately.

nico-famedly (changes requested):

lib/encryption/key_manager.dart
krille-chan: Why should we wait for it to complete and not just skip it? It would be tried again at the next sync anyway.

Before it was:
```
Every 10 seconds:
  check if there are keys to upload and upload all of them
```

Now it is:
```
After each SyncUpdate (every 1-30 seconds):
  check if there are keys to upload and upload all of them
```

So in the worst case it has one key to upload and would wait 30 seconds to upload it while before it was wait 10 seconds. But it happens quite frequently that we have 10 syncs in a row in one second because of typing updates, status updates or whatever and then it would be very much overhead to also do 10 upload key checks. They can just be skipped IMO

krille-chan: I would instead more prefer to implement a "cooldown" so that this is not triggered if the last trigger is less than 10 seconds ago

nico-famedly: Because of the second half of your commit message, where you want to upload them before logout. If you don't wait there, there is a good chance that a bigger upload will just fail.

nico-famedly (approved):
Neat!

lib/encryption/key_manager.dart
krille-chan: Addressed

PR #1586, opened 2023-10-19
feat: Add methods to load all room keys from online key backup

krille-chan:
This makes it possible to load
and sync all room keys
right after the bootstrap if the
app wants to do this.

### DIFF
--- a/test/encryption/online_key_backup_test.dart
+++ b/test/encryption/online_key_backup_test.dart
@@ -67,9 +67,35 @@ void main() {
           .request(client.getRoomById(roomId)!, sessionId, senderKey);
       expect(
           client.encryption!.keyManager
-                  .getInboundGroupSession(roomId, sessionId) !=
-              null,
-          true);
+              .getInboundGroupSession(roomId, sessionId)
+              ?.sessionId,
+          sessionId);
+    });
+
+    test('Load all Room Keys', () async {
+      if (!olmEnabled) return;
+      final keyManager = client.encryption!.keyManager;
+      const roomId = '!getroomkeys726s6s6q:example.com';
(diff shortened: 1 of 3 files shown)


test/encryption/online_key_backup_test.dart
```diff
@@ -65,6 +65,30 @@ void main() {
...
+    test('Load all Room Keys', () async {
+      if (!olmEnabled) return;
+      const roomId = '!726s6s6q:example.com';
+      const sessionId = 'ciM/JWTPrmiWPPZNkRLDPQYf9AW/I46bxyLSr+Bx5oU';
+      await client.encryption!.keyManager.loadAllKeysFromRoom(roomId);
+      expect(
+          client.encryption!.keyManager
+              .getInboundGroupSession(roomId, sessionId)
+              ?.sessionId,
+          sessionId);
```
nico-famedly, line 84: Can you add the expect call before the load as well and expect the key to not be present there? Otherwise I am not quite sure, if those keys didn't get loaded by a previous test.

krille-chan: it did because we use the same key everywhere in our data. I would more count this as a need for a general improvement of our tests because the same key is used in other tests as well so a lot of these tests could be false positive :-(

nico-famedly: Can we maybe use different keys just for these tests? Afaik they call the moc anyway, so that should work if we change the room id or so (apart from the all sessions call, but maybe we can add a key there, that we don't return on the others)?

krille-chan: changing room id is a good idea. Creating a new mock key is just complicated. I will try it out

krille-chan: Works :)

nico-famedly (approved):

PR #1588, opened 2023-10-20
build: Bump version to 0.22.5

krille-chan:


techno-disaster (approved):

nico-famedly:
Did you test the markdown line break changes? Doesn't that break pre tags like code blocks?

nico-famedly:
I think b5b5cfee5efa640551f841c0be37aec118b50371 broke code blocks with more than one line and we need to fix that.

PR #1595, opened 2023-10-28
fix: canRequestHistory is true while requestHistory throws an excepti…

krille-chan:
…on when events is empty

### DIFF
--- a/lib/src/timeline.dart
+++ b/lib/src/timeline.dart
@@ -75,7 +75,6 @@ class Timeline {
   bool _collectHistoryUpdates = false;
 
   bool get canRequestHistory {
-    if (events.isEmpty) return true;
     return room.prev_batch != null && events.last.type != EventTypes.RoomCreate;
   }
 


lib/src/timeline.dart
```diff
@@ -75,7 +75,6 @@ class Timeline {
   bool _collectHistoryUpdates = false;
 
   bool get canRequestHistory {
-    if (events.isEmpty) return true;
     return room.prev_batch != null && events.last.type != EventTypes.RoomCreate;
```
nico-famedly, line 78: You can only access events.last, if events is not empty.

nico-famedly (changes requested):

lib/src/timeline.dart
nico-famedly: for the normal room history, events should never have to be empty, since we should always be able to request our own history for our join event at least. This also might cause issues, since possibly on the first history pagination a timeline has no prev_batch token and is empty, if you just want to request the history for an arbitrary room? Not sure.

krille-chan: okay true, somehow I have missread the whole method. Will close this then

PR #1597, opened 2023-11-02
stricter lints

nico-famedly:
I guess the only controversial change is the runInRoot stuff, but I don't see, why we would ever want to await those. The few cases, where we do, seem to fit better when we just await them directly?

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,17 +2,18 @@ include: package:lints/recommended.yaml
 
 linter:
   rules:
-    avoid_print: true
-    prefer_final_locals: true
-    prefer_final_in_for_each: true
-    sort_pub_dependencies: true
-    always_use_package_imports: true
     always_declare_return_types: true
+    always_use_package_imports: true
+    avoid_bool_literals_in_conditional_expressions: true
+    avoid_print: true
+    non_constant_identifier_names: false # seems to wrongly diagnose static const variables
+    prefer_final_in_for_each: true
+    prefer_final_locals: true
     prefer_single_quotes: true
     sort_child_properties_last: true
+    sort_pub_dependencies: true
     unawaited_futures: true
     unsafe_html: true
-    non_constant_identifier_names: false # seems to wrongly diagnose static const variables
 
 analyzer:
   errors:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -8,14 +8,19 @@ linter:
     avoid_print: true
     cancel_subscriptions: true
     discarded_futures: true
-    non_constant_identifier_names: false # seems to wrongly diagnose static const variables
+    no_adjacent_strings_in_list: true
     prefer_final_in_for_each: true
     prefer_final_locals: true
     prefer_single_quotes: true
     sort_child_properties_last: true
     sort_pub_dependencies: true
+    test_types_in_equals: true
+    throw_in_finally: true
     unawaited_futures: true
+    unnecessary_statements: true
     unsafe_html: true
+
+    non_constant_identifier_names: false # seems to wrongly diagnose static const variables
 
 analyzer:
   errors:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -6,6 +6,8 @@ linter:
     always_use_package_imports: true
     avoid_bool_literals_in_conditional_expressions: true
     avoid_print: true
+    cancel_subscriptions: true
+    discarded_futures: true
     non_constant_identifier_names: false # seems to wrongly diagnose static const variables
     prefer_final_in_for_each: true
     prefer_final_locals: true
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,23 +2,43 @@ include: package:lints/recommended.yaml
 
 linter:
   rules:
-    always_declare_return_types: true
-    always_use_package_imports: true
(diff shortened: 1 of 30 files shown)


techno-disaster (commented):
Mostly LGTM, can you explain why some of the functions don't run in root anymore? `updateSessionUsage` for example

analysis_options.yaml
```diff
@@ -2,19 +2,24 @@ include: package:lints/recommended.yaml
 
 linter:
   rules:
-    camel_case_types: true
+    always_declare_return_types: true
```
krille-chan, line 5: Do you have some infos how you have decided for this lint rules? Are they from a source or have you checked every lint rule by yourself if they make sense? I would prefer to have a reason for any point where we differ from the default

lib/encryption/encryption.dart
```diff
@@ -100,11 +100,12 @@ class Encryption {
 
   void handleDeviceOneTimeKeysCount(
       Map<String, int>? countJson, List<String>? unusedFallbackKeyTypes) {
-    runInRoot(() => olmManager.handleDeviceOneTimeKeysCount(
+    runInRoot(() async => olmManager.handleDeviceOneTimeKeysCount(
```
krille-chan, line 103: I don't get why this makes a difference 🤔 the method `handleDeviceOneTimeKeysCount` is void and not `Future<void>` anyway and not even async so we are not awaiting this anyway

krille-chan: Probably a lint rule enforces this?

analysis_options.yaml
nico-famedly: I checked each of those rules one by one depending on what value they provide. I don't think that needs to be repeated in the config file, since the reasoning is pretty much the same as in their description, however to repeat it here:

- avoid_bool_literals_in_conditional_expressions is about cases like `condition ? true : boolExpression`, which are just a more verbose way to say `condition || boolExpression`
- cancel_subscriptions: Is about not leaking scream subscriptions, which would cause performance degradations over time or memory leaks
- discarded_futures: is like unawaited futures, but in functions which have not been marked as async. We tend to make this mistake quite often, that we don't await something and don't document why. This forces a linter annotation at least for cases where we do want that.
- no_adjacent_strings_in_list: Is about cases like `['a' 'b', 'c']`. Someone forgot the comma there. If that was intentional, they can pack it in one string literal or use a + to concatenate.
- test_types_in_equals: is a bit niche, so we could probably drop it. It is just on how to implement operator== correctly.
- throw_in_finally: Is a warning that throwing in a finally block is usually unintentional and could lead to weird program execution. Generally you want the throw statement outside of the finally block.
- unnecessary_statements: Just warns on statements, that return a value, that is never assigned and that have no side effect. It doesn't warn on function calls, that just return a value, but if you for example add `1+1` and don't assign that to a variable, that was probably a mistake.

In general I only enabled linter flags, that seem to have no drawbacks, but feel free to point out any that do. We diverge here from the default by being stricter than the default lints. We do that, because we want to avoid more bugs and this has pretty much no effect on existing code apart from where it was possibly already buggy to begin with.

lib/encryption/encryption.dart
nico-famedly: Yes, that is the one drawback I found from the discarded_futures. It does warn you when using a Future in a closure, but the closure is not marked as async. However I do think the value it provides elsewhere is beneficial.

analysis_options.yaml
krille-chan: Are you sure that all of those rules are not in the list of default ones? Might sound reasonable to not add an argumentation for every single rule but a comment above them to explain why this was added and maybe explain one or two could help in the future to understand this

lib/encryption/encryption.dart
krille-chan: Okay sounds fair

analysis_options.yaml
nico-famedly: I am pretty sure they are not in the default set. In fact I removed a rule, that was in the default set, that we had explicitly enabled (`avoid_function_literals_in_foreach_calls`).

Can you give me an example of what that comment would look like? I think it is obvious, that linter rules are enabled to prevent bugs and each of those linter rules has a better explanation in the dart docs, that I would prefer not to duplicate. And so far we never explained, why we enable linter rules either and we do have quite a few of them. We only document, why we disable some. As such, I have no idea what the comment would say...

Basically what I always did is, if I didn't understand a linter rule, I went to https://dart.dev/tools/linter-rules/avoid_function_literals_in_foreach_calls to figure out what it does and why one would want to use it. That seems to be more beneficial to me, than duplicating that information in a worse way.

krille-chan: Hmm something like:

# Added more strict linter rules than in default for better code quality.
# We picked "avoid_bool_literals_in_conditional_expressions", "no_adjacent_strings_in_list" and
# "unnecessary_statements" for better readability and "cancel_subscriptions", "discarded_futures" to
# avoid performance leaks.

Or maybe even group the lints by topics like "Code readability", "Performance" and so on

krille-chan (dismissed):

analysis_options.yaml
krille-chan: However, you will do it right. Have approved it :)

PR #1599, opened 2023-11-03
stale call checker leak

nico-famedly:
We were accumulating stale call checkers with every room. Also when we removed a room, the stale call checker kept running in some instances. We can just run the checker after the sync without a timer though, which means we don't need to clean up the timers in the end and we don't need to call sync in every call checker.


nico-famedly:
(depends on #1597 )

techno-disaster (approved):

PR #1600, opened 2023-11-05
fix: catch correct exception type for connection problems

krille-chan:


nico-famedly (approved):
Do we not need the MatrixConnectionException then? Or is that also derived from IOException?

krille-chan:
> Do we not need the MatrixConnectionException then? Or is that also derived from IOException?

No, this is only used in other requests which I also have changed in a PR in the matrix api lite. Sync should never throw an MatrixConnectionException

nico-famedly:
So we also don't need to depend on the newer api_lite, since it was already an IO error for this request?

krille-chan:
> So we also don't need to depend on the newer api_lite, since it was already an IO error for this request?

Correct. It is for years already

nico-famedly (approved):

PR #1602, opened 2023-11-07
fix: Wait for secrets come down sync

krille-chan:
Somehow on some servers
we wait forever for the keys
while they are not detected
in the sync. I think this has
something to do with how
different server handle uri
encoding. However, this should
make it more stable.

Other things I tried: Put the key we are waiting for in Uri.encodeComponent() or wait on both encoded and not encoded. Found no better solution to this :-(


krille-chan:
Haven't seen that TD fixed it already at https://github.com/famedly/matrix-dart-sdk/pull/1584

PR #1606, opened 2023-11-13
fix: clear local database on logout even if server timesout

techno-disaster:


techno-disaster:
could use a test

nico-famedly (approved):
looks reasonable to me.

test/client_test.dart
```diff
@@ -335,6 +335,94 @@ void main() {
...
+    });
+
+    test('Logout but this time server is dead', () async {
+      final oldapi = FakeMatrixApi.currentApi?.api;
+      expect(
+          (FakeMatrixApi.currentApi?.api['PUT']?.keys.where((element) =>
+              element.startsWith('/client/v3/room_keys/keys?version')))?.length,
+          1);
+      // not a huge fan, open to ideas
+      FakeMatrixApi.currentApi?.api = {};
```
nico-famedly, line 403: An alternative could have been to instead timeout on the key upload call if it contains some "magic" key, but this works :)

PR #1609, opened 2023-11-15
fix: Do only convert linebreaks to br tags in p blocks

krille-chan:
This changes the behavior
of the markdown method to
only convert linebreaks inside
of p blocks. I found no better
solution yet for the problem
as otherwise also lists
will have linebreaks between
the list items. Unfortunately
the default linebreak syntax
seems not to fulfill our needs.


nico-famedly (dismissed):

nico-famedly:
Ah, this breaks some tests. I guess this needs fixes :)

nico-famedly (approved):

Issue #1579, opened 2023-10-04
Update dehydrated devices feature to work with latest synapse version


nikzen:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

The SDK implemented the dehydrated devices feature in the first place. However till our implementation some work has happened on the server side so it is not compatible anymore.

### Describe your ideal solution

Update the SDK to work with the dehydrated devices implementation of the latest synapse version.

nico-famedly:
Okay, so the changes needed are mostly to merge the device key upload with the device creation. Basically you create the device id locally now and then just upload keys and device in one request.

PR #1614, opened 2023-11-16
feat: Update dehydrated devices implementation to current MSC

nico-famedly:
BREAKING CHANGE: This replaces the old dehydrated devices implementation, since there is no way to query what is supported easily and supporting both would be complicated.

fixes https://github.com/famedly/matrix-dart-sdk/issues/1579

### DIFF
--- a/lib/encryption/olm_manager.dart
+++ b/lib/encryption/olm_manager.dart
@@ -57,10 +57,12 @@ class OlmManager {
   final Map<String, List<OlmSession>> _olmSessions = {};
 
   // NOTE(Nico): On initial login we pass null to create a new account
-  Future<void> init(
-      {String? olmAccount,
-      required String? deviceId,
-      String? pickleKey}) async {
+  Future<void> init({
+    String? olmAccount,
+    required String? deviceId,
+    String? pickleKey,
+    String? dehydratedDeviceAlgorithm,
+  }) async {
     ourDeviceId = deviceId;
     if (olmAccount == null) {
       try {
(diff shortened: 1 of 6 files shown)


nico-famedly:
testing instructions:

1. Have a server with MSC3814 enabled
2. Patch your app to set the "enableDehydratedDevices" flag to true on the client
3. Sign in a device on user A -> you should see it upload a dehydrated device during bootstrap
4. Send user A a message -> user should be able to read it
5. Sign out the device from user A
6. Send another message to user A -> this should create a new megolm key
7. Sign in a new device for user A -> during bootstrap you should see it querying the dehydrated device, receiving the messages and replacing the device. After that user A should be able to read the second message, even though he was signed out when it was sent.

lib/encryption/olm_manager.dart
```diff
@@ -57,21 +57,25 @@ class OlmManager {
...
     if (olmAccount == null) {
       try {
         await olm.init();
         _olmAccount = olm.Account();
         _olmAccount!.create();
         if (!await uploadKeys(
-            uploadDeviceKeys: true,
-            updateDatabase: false,
-            // dehydrated devices don't have a device id when created, so skip upload in that case.
-            skipAllUploads: deviceId == null)) {
```
krille-chan, line 74: We no longer need to skip all uploads?

krille-chan (approved):
Had a hard time finding a server to test this 😭 however from what I see in the code this should not affect anything outside of this feature so it should be safe to merge as an experimental feature change

lib/encryption/olm_manager.dart
nico-famedly: No, we don't need to, since we actually upload the device keys at the same time as the other keys now. This is one of the changes in the msc, that the device id is generated client side, so that you can upload everything immediately.

PR #1615, opened 2023-11-17
handle call when multiple devices are involved for one account

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/1512

### DIFF
--- a/lib/src/voip/voip.dart
+++ b/lib/src/voip/voip.dart
@@ -337,11 +337,17 @@ class VoIP {
     await delegate.stopRingtone();
     Logs().v('[VOIP] onCallHangup => ${content.toString()}');
     final String callId = content['call_id'];
+    final String partyId = content['party_id'];
     final call = calls[callId];
     if (call != null) {
       if (call.room.id != roomId) {
         Logs().w(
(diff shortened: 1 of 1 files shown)


lib/src/voip/voip.dart
```diff
@@ -366,6 +373,11 @@ class VoIP {
             'Ignoring call reject for room $roomId claiming to be for call in room ${call.room.id}');
         return;
       }
+      if (call.remotePartyId != partyId) {
```
techno-disaster, line 376: remotePartyId is only set when there you already get a answer, if there was no answer, this would break ending calls

techno-disaster (changes requested):

lib/src/voip/voip.dart
coder-with-a-bushido: added a null check before that

techno-disaster (approved):
LGTM

techno-disaster (changes requested):
Can you quickly change the issue description so that we don't completely close the linked issue?

td-famedly (approved):

PR #1618, opened 2023-11-21
feat: Add sendRaw command

krille-chan:
Makes debugging and hacking
easier

### DIFF
--- a/lib/src/utils/commands_extension.dart
+++ b/lib/src/utils/commands_extension.dart
@@ -17,6 +17,7 @@
  */
 
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:matrix/matrix.dart';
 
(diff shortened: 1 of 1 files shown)


lib/src/utils/commands_extension.dart
```diff
@@ -276,6 +277,14 @@ extension CommandsClientExtension on Client {
         txid: args.txid,
       );
     });
+    addCommand('sendRaw', (args) async {
```
nico-famedly, line 280: Wouldn't this cause a problem, if this command was available in TIM clients?

krille-chan: TIM-Requirement is only to not be able to send custom state events afaik

nico-famedly (approved):

Issue #1621, opened 2023-11-24
New dart:io import breaks SDK compatibility with dart2js JIT compiler


TheOneWithTheBraid:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [X] I already asked for help in the chat

### In which Project did the bug appear?

Other

### On which platform did the bug appear?

Firefox

### SDK Version

b1b2b2ffe2bf49d7a0a09554bf866bd9ea6b47b1

### Describe the problem caused by this bug

Commit [b1b2b2f](https://github.com/famedly/matrix-dart-sdk/commit/b1b2b2ffe2bf49d7a0a09554bf866bd9ea6b47b1#r133430560) introduced more fine handling of network exceptions in the sync loop. Unfortunately, this dropped in `dart:io` as direct import into `client.dart`. The previous approach was not that fine graded, but imho way better than just silently dropping dart2js JIT support - at least as long as you have customers using it 😜.

### Steps To Reproduce

This issue is absolutely useless for external issue reporters btw.

### Screenshots or Logs

```
[WARNING] build_web_compilers:entrypoint on web/main.dart: Skipping compiling secure_messaging_app|web/main.dart with ddc because some of its
transitive libraries have sdk dependencies that not supported on this platform:

matrix|lib/src/client.dart
```

krille-chan:
Should be fixed now

PR #1622, opened 2023-11-24
fix: dart:io import in client.dart

TheOneWithTheBraid:
Fixes: #1621

Apart from reverting [b1b2b2f](https://github.com/famedly/matrix-dart-sdk/commit/b1b2b2ffe2bf49d7a0a09554bf866bd9ea6b47b1), I also restored the improved error message @krille-chan initially introduced with that commit.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -19,7 +19,6 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:core';
-import 'dart:io';
 import 'dart:math';
 import 'dart:typed_data';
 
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -1753,8 +1752,8 @@ class Client extends MatrixApi {
         Logs().w('The user has been logged out!');
         await clear();
       }
-    } on IOException catch (e, s) {
-      Logs().w('Syncloop failed: Client has not connection to the server');
+    } on MatrixConnectionException catch (e, s) {
```
krille-chan, line 1755: This is the wrong exception here and never thrown by the sync loop

TheOneWithTheBraid: Hmmm dow you have any other proposal ? Just a generic catch statement ?

krille-chan: > Hmmm dow you have any other proposal

Nope. Needs some investigation which exception is thrown by http package but not depending on dart:io

TheOneWithTheBraid: Should work ! Sorry, was in a meeting the entire time.

TheOneWithTheBraid:
Closed by #1626 .

PR #1623, opened 2023-11-24
fix: key uploads only running once

nico-famedly:
There were several issues here. Key uploads in general failed, because the await caused a Future<dynamic> to be returned, which failed type checking.

But also we never unset our future, which was used for the online key backup uploads, which meant we would very quickly stop uploading any keys at all.

### DIFF
--- a/lib/encryption/encryption.dart
+++ b/lib/encryption/encryption.dart
@@ -232,9 +232,12 @@ class Encryption {
         // the entry should always exist. In the case it doesn't, the following
         // line *could* throw an error. As that is a future, though, and we call
         // it un-awaited here, nothing happens, which is exactly the result we want
-        // ignore: discarded_futures
-        client.database?.updateInboundGroupSessionIndexes(
-            json.encode(inboundGroupSession.indexes), roomId, sessionId);
+        client.database
+            // ignore: discarded_futures
+            ?.updateInboundGroupSessionIndexes(
+                json.encode(inboundGroupSession.indexes), roomId, sessionId)
+            // ignore: discarded_futures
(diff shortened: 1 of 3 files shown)


lib/encryption/encryption.dart
```diff
@@ -232,9 +232,12 @@ class Encryption {
...
         // it un-awaited here, nothing happens, which is exactly the result we want
-        // ignore: discarded_futures
-        client.database?.updateInboundGroupSessionIndexes(
-            json.encode(inboundGroupSession.indexes), roomId, sessionId);
+        client.database
+            // ignore: discarded_futures
+            ?.updateInboundGroupSessionIndexes(
+                json.encode(inboundGroupSession.indexes), roomId, sessionId)
+            // ignore: discarded_futures
+            .onError((e, _) => Logs().e('Ignoring error for updating indexes'));
```
krille-chan, line 240: What does this change? Before it should have displayed a stacktrace in the error zone but now it pipes the error to the logs and ignores the stacktrace? It is already an not awaited async so it would not let this method crash

krille-chan:
I have tested this with FluffyChat now and it works great. Thank you very much.

Before it has only uploaded the keys on app start and on logout. Note that this only counts for incoming keys. The sessions started from the device itself are always uploaded. So this fixes a case where some keys are maybe not uploaded when someone looses the device and does not logout correctly or does never restart the device. Especially for web sessions which might get logged out externally this improves the situation and makes the key backup much more stable :+1: :+1: :+1:

krille-chan:
One of the main reasons, that the future is never set to null seems to be my fault. We should write a unit test for exactly this case (upload keys is triggered more than only once)

krille-chan (approved):

Issue #1624, opened 2023-11-24
[MSC 3814] Dehydrated devices no longer automatically upload group sessions


TheOneWithTheBraid:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [X] I already asked for help in the chat

### On which platform did the bug appear?

Firefox

### SDK Version

be1f125

### Describe the problem caused by this bug

Since [be1f125](https://github.com/famedly/matrix-dart-sdk/commit/be1f125414c30fa5bd12e5e55b6b0449dcba7d29#r133421492), the SDK no longer automatically uploads group session keys in case the device is hydrated.

### Steps To Reproduce

1. Log in to an account
2. Dehydrate your device
3. Re login claiming that dehydrated device
4. See how your group sessions are no longer automatically uploaded
5. This can be checked by repeating steps 2 - 3 after sending some messages

krille-chan:
Has been fixed by https://github.com/famedly/matrix-dart-sdk/pull/1623

PR #1625, opened 2023-11-24
fix: add missing upload for inbound group sessions on dehydrated devices

TheOneWithTheBraid:
In case the auto upload is already running, execute it once.

Fixes: #1624


TheOneWithTheBraid:
Closed by #1623 .

PR #1627, opened 2023-11-27
chore: bump version to 0.23.0

nico-famedly:


CHANGELOG.md
```diff
@@ -1,4 +1,27 @@
+## [0.23.0] - 27th November 2023
+
+This release includes a small breaking change. Starting a verification request now always returns a future.
```
krille-chan, line 3: Can you say which method exactly?

krille-chan (dismissed):

CHANGELOG.md
nico-famedly: Well, all of the start verification methods now return a future, so both https://github.com/famedly/matrix-dart-sdk/blob/ce1727ec52adead6b6d9c2f42249a6585fff9551/lib/src/utils/device_keys_list.dart#L70 and https://github.com/famedly/matrix-dart-sdk/blob/ce1727ec52adead6b6d9c2f42249a6585fff9551/lib/encryption/utils/key_verification.dart#L328

But I guess the one one that changes is the one in the device_keys_list. Will do.

krille-chan (approved):

PR #1629, opened 2023-11-29
build: Release v0.23.1

krille-chan:


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: matrix
 description: Matrix Dart SDK
-version: 0.23.0
+version: 0.23.1
```
nico-famedly, line 3: This is a breaking change at least for dehydrated devices users (since it breaks compat with the old server parts). Can we bump this to 0.24.0?

krille-chan: Done

nico-famedly (approved):

PR #1631, opened 2023-12-02
update read receipts from mainThreadupdate read receipts from mainThread

simon-moquist:
Fixes: https://github.com/famedly/product-management/issues/1020

Update other users read receipts from both global and mainThread.

This is primarily intended to fix read receipts not always updating in FluffyChat (relates to https://github.com/krille-chan/fluffychat/issues/549). I do this by not only checking for read receipts in global but also mainThread as some clients only seem to update mainThread and not global.

This is my first pull request and I am not experienced with dart so hopefully any issues with this PR are fixable :)

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -329,6 +329,19 @@ class Event extends MatrixEvent {
             room.unsafeGetUserFromMemoryOrFallback(entry.key),
             entry.value.timestamp))
         .toList();
+    if (receipts.mainThread?.otherUsers.entries != null) {
+      receipts.mainThread!.otherUsers.entries
+          .where((entry) => entry.value.eventId == eventId)
+          .forEach((entry) {
+        if (receiptsList
+                .indexWhere((element) => element.user.stateKey == entry.key) ==
+            -1) {
+          receiptsList.add(Receipt(
+              room.unsafeGetUserFromMemoryOrFallback(entry.key),
+              entry.value.timestamp));
+        }
+      });
+    }
 
     final own = receipts.global.latestOwnReceipt;
     if (own != null && own.eventId == eventId) {


simon-moquist:
Thank you for the reply.

I figured that the global timeline and mainThread would contain receipts for the same types of events depending on whether the client sending them has implemented threaded receipts or not.
Given that, I reasoned that all receipts in the main thread could safely be merged to the global list but not necessarily the  other way round since an un-threaded receipt could relate to a threaded message. Is this perhaps an incorrect assumption?

Anecdotally, I seem to not get any read receipts displayed in FluffyChat from other users using Element but they are displayed when sent from FluffyChat or Element X.

lib/src/event.dart
```diff
@@ -329,6 +329,19 @@ class Event extends MatrixEvent {
             room.unsafeGetUserFromMemoryOrFallback(entry.key),
             entry.value.timestamp))
         .toList();
+    if (receipts.mainThread?.otherUsers.entries != null) {
+      receipts.mainThread!.otherUsers.entries
```
nico-famedly, line 333:
```suggestion
    final mainThread = receipts.mainThread!.otherUsers.entries
    if (mainThread != null) {
      mainThread
```

lib/src/event.dart
```diff
@@ -329,6 +329,19 @@ class Event extends MatrixEvent {
...
+          .where((entry) => entry.value.eventId == eventId)
+          .forEach((entry) {
+        if (receiptsList
+                .indexWhere((element) => element.user.stateKey == entry.key) ==
+            -1) {
```
nico-famedly, line 338:
```suggestion
        if (receiptsList
                .every((element) => element.user.stateKey != entry.key)) {
```

nico-famedly (commented):
If you read the global timeline, you have more messages to read than in the main timeline. So if you only ever read the main timeline, you would not have read the full global timeline. So in theory this is wrong, but in practice doing the "wrong" thing might be the better UX.

So under that aspect I would cautiously approve this change. Not sure yet though, if we then also should change this in other places.

I left a few code comments though!

td-famedly:
took over and added some tests here https://github.com/famedly/matrix-dart-sdk/pull/1857

PR #1632, opened 2023-12-04
chore: create_gh_release job

td-famedly:


.github/workflows/publish.yml
```diff
@@ -12,3 +12,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create release
+        run: |
+          releaseRegex="^v[0-9]+\.[0-9]+\.[0-9]+$"
+          releaseCandidateRegex="^v[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$"
+          if [[ ${{ github.ref_name }} =~ $releaseRegex ]]; then
+            gh release create ${{ github.ref_name }} --generate-notes -t ${{ github.ref_name }} --verify-tag
+          elif [[ ${{ github.ref_name }} =~ $releaseCandidateRegex ]]; then
+            gh release create ${{ github.ref_name }} --generate-notes --prerelease -t ${{ github.ref_name }} --verify-tag
```
nico-famedly, line 29: I still don't think those notes are very useful, can't we just extract them from the changelog file?

nico-famedly: The script we had in the past worked fine for almost a year until we migrated to github and usually when a dev adds something manually, it is specific notes about the migrations, which can be quite important.

td-famedly: The specific notes part makes sense. Guess I need to hunt for that script now :3

nico-famedly (approved):

PR #1638, opened 2023-12-08
fix: canSendDefaultMessage ignores overwritten events

krille-chan:
This refactors three methods
regarding powerlevels. It
makes them a little bit more
readable and removes
unnecessary checks. it
also changes the canSendDefaultMessage
method in a way that it also
checks the events map in the
powerlevels so that it returns
true if message events are
overwritten there. It also
improves the dart docs and
explains in more detail what it
does.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1788,14 +1788,18 @@ class Room {
   /// level of 100, and all other users have a power level of 0.
   int getPowerLevelByUserId(String userId) {
     final powerLevelMap = getState(EventTypes.RoomPowerLevels)?.content;
-    if (powerLevelMap == null) {
-      return getState(EventTypes.RoomCreate)?.senderId == userId ? 100 : 0;
-    }
-    return powerLevelMap
-            .tryGetMap<String, Object?>('users')
-            ?.tryGet<int>(userId) ??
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -1877,14 +1881,17 @@ class Room {
...
+  /// The default level required to send message events. This checks if the
+  /// user is capable of sending `m.room.message` events.
+  /// Please be aware that this also returns false
+  /// if the room is encrypted but the client is not able to use encryption.
+  /// If you do not want this check or want to check other events like
+  /// `m.sticker` use `canSendEvent('<event-type>')`.
+  bool get canSendDefaultMessages {
+    if (encrypted && !client.encryptionEnabled) return false;
+
+    return canSendEvent(encrypted ? EventTypes.Encrypted : EventTypes.Message);
```
td-famedly, line 1893: Potential minor version change? This changes the previous behavior of sending events_default to the above events. (although afaik those are never set in powerlevels and canSendEvent should default to events_default later anyway)

td-famedly (approved):

lib/src/room.dart
krille-chan: Will release a new major version as the next anyway :)

PR #1639, opened 2023-12-08
refactor: Make possible to wait for first sync and await first sync before create megolm session

krille-chan:
Closes https://github.com/famedly/matrix-dart-sdk/issues/361

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1586,9 +1586,9 @@ class Client extends MatrixApi {
       );
 
       /// Timeout of 0, so that we don't see a spinner for 30 seconds.
-      final syncFuture = _sync(timeout: Duration.zero);
+      firstSyncReceived = _sync(timeout: Duration.zero);
       if (waitForFirstSync) {
-        await syncFuture;
+        await firstSyncReceived;
       }
       return;
     } catch (e, s) {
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -1586,9 +1586,9 @@ class Client extends MatrixApi {
       );
 
       /// Timeout of 0, so that we don't see a spinner for 30 seconds.
-      final syncFuture = _sync(timeout: Duration.zero);
+      firstSyncReceived = _sync(timeout: Duration.zero);
       if (waitForFirstSync) {
```
td-famedly, line 1590: Are you sure this would work for the first sync? 

The docs say: The server need only populate this property for an incremental /sync (i.e., one where the since parameter was specified) https://spec.matrix.org/v1.9/client-server-api/#e2e-extensions-to-sync

We could just use /keys/query here

krille-chan: I'm not sure that I get this. It works the same way as before but now you can await it from outside as well

td-famedly: Hm not sure either, I think I confused getting the device list with /sync device_list updates

td-famedly (approved):

PR #1644, opened 2023-12-11
refactor: Remove duplicated copyMap method and fix type error

krille-chan:
dynamic.copy returned a type
error so I reverted the previous
change of the copyMap
method to an extension. Also
I found out that we have used
two copyMap methods in the
SDK which did exactly the same.
I deleted the old one and
changed the tests.


lib/src/utils/copy_map.dart
```diff
@@ -1,22 +1,20 @@
...
 
-  dynamic _castValue(dynamic value) {
-    if (value is Map) {
-      return value.copy;
-    }
-    if (value is List) {
-      return value.map(_castValue).toList();
-    }
-    return value;
+dynamic _castValue(dynamic value) {
```
nico-famedly, line 12: Some day I would like to return an Object? here and possibly not reassign to a Map of dynamics in the loop above, but for now there is no benefit, I guess.

test/matrix_api/map_copy_extension_test.dart
```diff
@@ -30,11 +30,11 @@ void main() {
           'list': [1, 2],
         },
       };
-      final copy = original.copy();
+      final copy = copyMap(original);
       original['child']['attr'] = 'raccoon';
-      expect(copy['child']['attr'], 'bunny');
+      expect((copy['child'] as Map)['attr'], 'bunny');
```
nico-famedly, line 35: Do we need this cast now or is this just an additional check? I can't quite tell what we are doing different, that makes this necessary now? Was it because the old map was `<String, dynamic>` and this is `<String, Object?>` now?

nico-famedly (approved):

test/matrix_api/map_copy_extension_test.dart
krille-chan: We need it as the new copymap method now returns a `Map<String, Object>` and not `Map<String, dynamic>` anymore so `copy['child']` is a `Object?` and we need to cast it before we can access 'attr'

nico-famedly: I see, thanks for explaining! :)

PR #1645, opened 2023-12-12
feat: cache getConfig request

coder-with-a-bushido:
This caches the `getConfig` request which is rate-limited.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -22,6 +22,7 @@ import 'dart:core';
 import 'dart:math';
 import 'dart:typed_data';
 
+import 'package:async/async.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:http/http.dart' as http;
 import 'package:mime/mime.dart';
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -1011,6 +1011,20 @@ class Client extends MatrixApi {
     _archivedRooms.add(ArchivedRoom(room: archivedRoom, timeline: timeline));
   }
 
+  bool _isServerConfigCached = false;
+  ServerConfig? _cachedServerConfig;
```
td-famedly, line 1015: Do we need the bool? 

Also caching indefinitely sounds wrong, probably add a timeout of say 15 min (random number)

coder-with-a-bushido: I'm caching till we call `init` again. I think that's when the homeserver changes. Do you think we should cache based on timeout?

td-famedly: getConfig values could change and having it reflect without restart the client would be nice :D

coder-with-a-bushido: I can see the [go matrix sdk's bridge part](https://github.com/mautrix/go/blob/685d5f71da580e63c0f524e103abe83452d54d11/bridge/bridge.go#L637) and [python sdk's bridge part](https://github.com/mautrix/python/blob/8d39c733d7ad2d1a11b70c255babfad6941faf3a/mautrix/bridge/matrix.py#L259) caching it only once in the `start` call that comes after `init`. 

The both have a default value for the "m.upload.size" too - `50 * 1024 * 1024`.

nico-famedly: You could use an AsyncCache with a cache duration of one hour probably? https://pub.dev/documentation/async/latest/async/AsyncCache-class.html

lib/src/client.dart
```diff
@@ -1535,6 +1543,7 @@ class Client extends MatrixApi {
         encryption = null;
       }
       await encryption?.init(olmAccount);
+      _serverConfigCache.invalidate();
```
nico-famedly, line 1546: Nitpick, but I would invalidate that next to the groupCallSessionId

nico-famedly (approved):

td-famedly (approved):

PR #1646, opened 2023-12-12
refactor: Add delete database method

krille-chan:
This adds a delete database
method used for migration to
correctly delete the whole
legacy database instead just
empty it.

### DIFF
--- a/lib/src/database/sqflite_box.dart
+++ b/lib/src/database/sqflite_box.dart
@@ -1,20 +1,22 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:sqflite_common/sqlite_api.dart';
+import 'package:sqflite_common/sqflite.dart';
 
 /// Key-Value store abstraction over Sqflite so that the sdk database can use
 /// a single interface for all platforms. API is inspired by Hive.
 class BoxCollection {
   final Database _db;
   final Set<String> boxNames;
+  final DatabaseFactory? _factory;
 
-  BoxCollection(this._db, this.boxNames);
+  BoxCollection(this._db, this.boxNames, this._factory);
 
   static Future<BoxCollection> open(
     String name,
     Set<String> boxNames, {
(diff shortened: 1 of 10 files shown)


test/database_api_test.dart
```diff
@@ -472,6 +472,9 @@ void main() {
     test('Close', () async {
       await database.close();
     });
+    test('Delete', () async {
+      await database.delete();
```
nico-famedly, line 476: Can you test that reopening does not contain the same data?

nico-famedly (commented):
I have some tiny requests for improvements, otherwise this seems to look fine?

mErrenst (approved):
LGTM

lib/src/database/sqflite_box.dart
```diff
@@ -110,6 +112,9 @@ class BoxCollection {
       );
 
   Future<void> close() => _db.close();
+
+  Future<void> delete() =>
+      (_factory ?? databaseFactory).deleteDatabase(_db.path);
```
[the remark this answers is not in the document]
krille-chan: I was not sure about this 🤔 and had a hard time finding information but since 2019 it looks like that sqflite is force closing the database always by itself before deleting it: https://github.com/tekartik/sqflite/issues/213

test/database_api_test.dart
krille-chan: Added

lib/src/database/sqflite_box.dart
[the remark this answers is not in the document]
nico-famedly: Oh, they know how people will misuse the API, I see!

test/database_api_test.dart
nico-famedly: (this probably doesn't work properly, because in memory dbs vanish on close, but good enough, thanks!)

nico-famedly (approved):

PR #1647, opened 2023-12-12
refactor: Make getDisplayname and calcAvatar of room async

krille-chan:
This makes it possible to
load users lazy
for rooms to calculate the
name and the avatar. The
changes explain the new
behavior in detail in the
dart docs. The old
avatar getter now only
returns the avatar from the
room state m.room.avatar.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -223,6 +223,29 @@ class Room {
     return pinned is Iterable ? pinned.map((e) => e.toString()).toList() : [];
   }
 
+  /// Returns the heroes as `User` objects.
+  /// This is very useful if you want to make sure that all users are loaded
+  /// from the database, that you need to correctly calculate the displayname
+  /// and the avatar of the room.
+  Future<List<User>> loadHeroUsers() async {
+    var heroes = summary.mHeroes;
+    if (heroes == null) {
+      final directChatMatrixID = this.directChatMatrixID;
+      if (directChatMatrixID != null) {
+        heroes = [directChatMatrixID];
+      }
+    }
+
+    if (heroes == null) return [];
+
+    return await Future.wait(heroes.map((hero) async =>
+        (await requestUser(
+          hero,
+          ignoreErrors: true,
+        )) ??
+        User(hero, room: this)));
+  }
+
   /// Returns a localized displayname for this server. If the room is a groupchat
   /// without a name, then it will return the localized version of 'Group with Alice' instead
   /// of just 'Alice' to make it different to a direct chat.
--- a/test/room_test.dart
+++ b/test/room_test.dart
@@ -81,6 +81,9 @@ void main() {
         stateKey: '',
       ));
 
+      final heroUsers = await room.loadHeroUsers();
+      expect(heroUsers.length, 3);
+
       expect(room.id, id);
       expect(room.membership, membership);
       expect(room.notificationCount, notificationCount);


lib/src/room.dart
```diff
@@ -281,32 +304,39 @@ class Room {
     return t is String ? t : '';
   }
 
-  /// The avatar of the room if set by a participant.
+  /// The avatar of the room if set by a participant. If you want to also
+  /// let the avatar be calculated by the room heroes for DM rooms, use
+  /// `Room.calcAvatar()` instead.
   Uri? get avatar {
```
nico-famedly, line 310: I think it is a bad idea to change the behaviour of an existing function and I also don't really like, that this now makes the obvious function name the usually wrong one to use. If I want the room avatar, I will call room.avatar, I don't want to know about the subtleties of calculated avatars vs the avatar state event....

lib/src/room.dart
```diff
@@ -223,32 +221,57 @@ class Room {
...
+        User(hero, room: this)));
+  }
+
+  /// Returns a localized displayname for this room. If the room is a groupchat
   /// without a name, then it will return the localized version of 'Group with Alice' instead
   /// of just 'Alice' to make it different to a direct chat.
   /// Empty chats will become the localized version of 'Empty Chat'.
   /// This method requires a localization class which implements [MatrixLocalizations]
-  String getLocalizedDisplayname([
+  Future<String> getLocalizedDisplayname([
```
nico-famedly, line 254: I would actually prefer to cache this (and member count, room avatar, encryption status as well as membership) in a room info object and only load that when loading the room, and load all events lazily when requested instead.

lib/src/room.dart
```diff
@@ -281,32 +304,39 @@ class Room {
...
-      final user = directChatMatrixID;
-      if (user != null) {
-        return unsafeGetUserFromMemoryOrFallback(user).avatarUrl;
+    return null;
+  }
+
+  /// Calculates the avatar. First checks if `m.room.avatar` is set and then
+  /// tries to calculate it from the room heroes. It will use the first
+  /// hero whos avatar is not null.
+  /// For invitation this displays the avatar of the invitor as fallback.
```
nico-famedly, line 321: Do you want to reference https://spec.matrix.org/v1.9/client-server-api/#calculating-the-display-name-for-a-room here and why we diverge from it?

nico-famedly (commented):
Hm, I have some thoughts for this. I do think it is something we should improve, but I would have done it differently! :D

lib/src/room.dart
krille-chan: Not sure if I understand this correctly. You mean that we have a `await room.getInfo()` where then the calculated displayname and avatar is inside?

lib/src/room.dart
krille-chan: We should document if we differ from the spec on purpose. We don't do this on purpose here as we have implemented this logic a long time ago. I have tried do not change the logic much here (I think I improved it a little bit by not using the first user from heroes but the first one with an avatar url, but thats not a big change IMO). Changing it so that it is according to the spec could be a different issue

lib/src/room.dart
nico-famedly: More like, when we load the room from the database, it already loads the cached room name and avatar, so you can return it immediately and will not need to load any member events unless you want to update the room name (i.e. on sync).

krille-chan: This was the behavior with the old database but that was super slow because it did not scale that good. If we have 1000 rooms, then this will do 1000*x requests (x==heroes) and block the app start. While we actually only need the first 10 or so rooms as long as the user does not scroll down the chat list. Loading the missing users on demand has a much better performance

lib/src/room.dart
krille-chan: What would you suggest?

lib/src/room.dart
krille-chan: Other suggestion: We leave the methods as they are but add a method `room.loadHeroUsers()` and `bool get heroUsersAreCached`. So in the UI we could react on this instead.

But that would make it possible to mess it up in the app because the sdk consumer would need to know about it...

nico-famedly: No, you wouldn't load the heroes at all, you would only load them on /sync if the room name needs to be recalculated (because one of the heroes member events changed or the room name was set/unset, etc).

Basically you would calculate the room name and avatar once and then store it in the toJson/fromJson calls: https://github.com/famedly/matrix-dart-sdk/blob/94313a09047c444f332b8ec85cb8374262f6f662/lib/src/room.dart#L95

The calculation happens whenever any of the state contributing to the room name changes, which usually happens mostly on initial sync. Then whenever we load the room from the database, we just load 2 additional string values instead of loading each member event, which shouldn't add much overhead, since we already load id, prev_batch, notification counts and summary.

krille-chan: I've thought about this a lot and I'm not 100% sure if this fits our needs. There are advantages and disadvantages. However this would be a huge change to the sdk and the database and it might make more sense to do this when we have reduced the amount of databases which are in use.

What I currently would like to fix is the current behavior that we already lazy load the hero users but we do not know when to update the UI, because calling `room.getLocalizedDisplayname()` actually requests the user but we do not know when this is done to update the UI

lib/src/room.dart
krille-chan: But Avatar is now a Future and a Future should not be returned from a getter IMO. A getter is for something which is directly present. A Future is something we want to fetch, so this needs an operation to be done which is better displayed by a method(). However we would deprecate this avatar getter and mention the new method

nico-famedly (approved):

PR #1648, opened 2023-12-12
feat: Reset unread count directly on set read marker

krille-chan:
Like in other apps this resets
the unread count when
using setReadMarker without
waiting for the server to
responde. Should lead to
a much better UX when
dealing with a slow connection
to the server but is configurable
by the sdk consumer.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1356,8 +1356,28 @@ class Room {
   /// read receipt's location.
   /// If you set `public` to false, only a private receipt will be sent. A private receipt is always sent if `mRead` is set. If no value is provided, the default from the `client` is used.
   /// You can leave out the `eventId`, which will not update the read marker but just send receipts, but there are few cases where that makes sense.
-  Future<void> setReadMarker(String? eventId,
-      {String? mRead, bool? public}) async {
+  Future<void> setReadMarker(
+    String? eventId, {
+    String? mRead,
+    bool? public,
+    bool resetUnreadCountLocally = false,
+  }) async {
+    if (resetUnreadCountLocally) {
+      await _handleFakeSync(
+        SyncUpdate(
+          nextBatch: '',
+          rooms: RoomsUpdate(
+            join: {
+              id: JoinedRoomUpdate(
+                  unreadNotifications: UnreadNotificationCounts(
+                notificationCount: 0,
+                highlightCount: 0,
+              ))
+            },
+          ),
+        ),
+      );
+    }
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -1356,8 +1356,28 @@ class Room {
   /// read receipt's location.
   /// If you set `public` to false, only a private receipt will be sent. A private receipt is always sent if `mRead` is set. If no value is provided, the default from the `client` is used.
   /// You can leave out the `eventId`, which will not update the read marker but just send receipts, but there are few cases where that makes sense.
-  Future<void> setReadMarker(String? eventId,
-      {String? mRead, bool? public}) async {
+  Future<void> setReadMarker(
```
nico-famedly, line 1359: Can we add some local echo revert for this? Otherwise if we have no network connection, we will mark all rooms as read, but the server won't know this and will push us updated read counts without taking into account the events we have thought we have read already until we open the room again to read the new events, I think.

mErrenst (approved):
If I understand the code correctly if we set the read marker with resetUnreadCountLocally as true this would add a fake SyncUpdate locally to set the read marker even when there are server problems that delay the sync, right?
So the read marker would still be updated for others like before but locally we might not have the problem of not disappearing unread messages.
I can't see any reason not to merge this since it won't break anything with the optional new settings.

lib/src/room.dart
krille-chan: hmm, I know no good solution for this. Any suggestions?

nico-famedly: No, there is a reason I never implemented this :D 

A 50% solution would be to set the read marker to the previous value if the setReadMarker request fails and the read marker hasn't been updated in the mean time. That won't work when people kill the app during a request, but should work for the common network errors.

The more complicated solution would be to store the pending states in the database and periodically retry while they are pending (they should succeed once we sync again, so the difference in read counts can probably be ignored for one sync). That sounds like quite a bit of work though.

krille-chan: The complicated solution could mean that we implement a new feature "ReadMarkerManager" which stores read marker requests in the database, retrys then on failure after every sync similar to the ToDeviceQueue, handles that pending read marker requests are cleared when they come down the sync from another device and so on. You are right, that sounds like a very complicated feature. We could consider this in the future but for now this sounds like too much effort (tests would also be needed) just for this tiny convenience. Therefore I will close this PR for now

PR #1649, opened 2023-12-13
fix: ignore calls with age older than lifetime

td-famedly:
Fixes: https://github.com/famedly/product-management/issues/1469

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2142,6 +2142,22 @@ class Client extends MatrixApi {
               return false;
             });
           }
+
+          final age = callEvent.unsigned?.tryGet<int>('age') ??
+              (DateTime.now().millisecondsSinceEpoch -
+                  callEvent.originServerTs.millisecondsSinceEpoch);
+
+          callEvents.removeWhere((element) {
+            if (callEvent.type == EventTypes.CallInvite &&
+                age >
+                    (callEvent.content.tryGet<int>('lifetime') ??
+                        CallTimeouts.callInviteLifetime.inMilliseconds)) {
+              Logs().v(
+                  'Ommiting invite event ${callEvent.eventId} as age was older than lifetime');
+              return true;
+            }
+            return false;
+          });
         }
       }
     }
--- a/lib/src/voip/call.dart
+++ b/lib/src/voip/call.dart
@@ -30,18 +30,18 @@ import 'package:matrix/src/utils/cached_stream_controller.dart';
 /// version 1
 const String voipProtoVersion = '1';
 
-class Timeouts {
+class CallTimeouts {
   /// The default life time for call events, in millisecond.
-  static const lifetimeMs = 10 * 1000;
+  static const defaultCallEventLifetime = Duration(seconds: 10);
 
   /// The length of time a call can be ringing for.
-  static const callTimeoutSec = 60;
+  static const callInviteLifetime = Duration(seconds: 60);
 
   /// The delay for ice gathering.
-  static const iceGatheringDelayMs = 200;
+  static const iceGatheringDelay = Duration(milliseconds: 200);
 
   /// Delay before createOffer.
-  static const delayBeforeOfferMs = 100;
+  static const delayBeforeOffer = Duration(milliseconds: 100);
 }
 
 extension RTCIceCandidateExt on RTCIceCandidate {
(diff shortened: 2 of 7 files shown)


lib/src/client.dart
```diff
@@ -2118,6 +2118,17 @@ class Client extends MatrixApi {
                     'Ommit "${event.type}" event for an already terminated call');
                 return true;
               }
+              final age = callEvent.unsigned?.tryGet<int>('age') ??
```
nico-famedly, line 2121: Why are we not using the origin_server_ts?

nico-famedly: I guess the clocks between us and the server could be wrong?

td-famedly: > I guess the clocks between us and the server could be wrong?

Only reason I could come up with and the docs mention the same

lib/src/voip/call.dart
```diff
@@ -32,16 +32,16 @@ const String voipProtoVersion = '1';
 
 class Timeouts {
   /// The default life time for call events, in millisecond.
-  static const lifetimeMs = 10 * 1000;
+  static const lifetimeMs = Duration(milliseconds: 10000);
 
   /// The length of time a call can be ringing for.
-  static const callTimeoutSec = 60;
+  static const callTimeoutMs = Duration(milliseconds: 60000);
```
nico-famedly, line 38:
```suggestion
  static const callTimeout = Duration(seconds: 60);
```

lib/src/voip/call.dart
```diff
@@ -32,16 +32,16 @@ const String voipProtoVersion = '1';
 
 class Timeouts {
   /// The default life time for call events, in millisecond.
-  static const lifetimeMs = 10 * 1000;
+  static const lifetimeMs = Duration(milliseconds: 10000);
```
nico-famedly, line 35:
```suggestion
  static const lifetime = Duration(seconds: 10);
```

lib/src/voip/call.dart
```diff
@@ -32,16 +32,16 @@ const String voipProtoVersion = '1';
...
 
   /// The length of time a call can be ringing for.
-  static const callTimeoutSec = 60;
+  static const callTimeoutMs = Duration(milliseconds: 60000);
 
   /// The delay for ice gathering.
-  static const iceGatheringDelayMs = 200;
+  static const iceGatheringDelayMs = Duration(milliseconds: 200);
```
nico-famedly, line 41:
```suggestion
  static const iceGatheringDelay = Duration(milliseconds: 200);
```

lib/src/voip/call.dart
```diff
@@ -32,16 +32,16 @@ const String voipProtoVersion = '1';
...
 
   /// Delay before createOffer.
-  static const delayBeforeOfferMs = 100;
+  static const delayBeforeOfferMs = Duration(milliseconds: 100);
```
nico-famedly, line 44:
```suggestion
  static const delayBeforeOffer = Duration(milliseconds: 100);
```

nico-famedly (approved):
Some nits and maybe you can add a test for it although it might be possible it is already covered, since you modified some tests?

lib/src/voip/voip.dart
```diff
@@ -263,16 +267,22 @@ class VoIP {
...
     // by terminate.
     currentCID = callId;
-
-    await newCall.initWithInvite(
-        callType, offer, sdpStreamMetadata, lifetime, confId != null);
+    try {
+      await newCall.initWithInvite(
+          callType, offer, sdpStreamMetadata, lifetime, confId != null);
+    } catch (e, s) {
+      Logs().e('[VOIP] initWithInvite failed', e, s);
```
nico-famedly, line 274: So the call init failed, why do we show a call page? Shouldn't we error out? :D

test/calls_test.dart
```diff
@@ -0,0 +1,137 @@
...
+                  'offer': {'type': 'offer', 'sdp': 'sdp'}
+                },
+                senderId: '@alice:testing.com',
+                eventId: 'newevent',
+                originServerTs: DateTime.utc(1969),
+              )
+            ]))
+          })));
+      await Future.delayed(Duration(seconds: 3));
+      expect(voip.currentCID, null);
```
nico-famedly, line 81: Is there no way to synchronize this?

td-famedly: added a comment as discussed

PR #1650, opened 2023-12-14
fix: Increase timeout for initial sync from 10 seconds to 2 minutes

krille-chan:
Tested in FluffyChat on iOS and worked good for me. Initial sync was much faster :-)

This fixes the response
timeout for the initial
sync and sets it to 2
minutes instead of 10
seconds. This should increase
the speed for the initial
sync especially for large
accounts. This change
also adds some
documentation in the
code about what
timeout does what.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1695,7 +1695,12 @@ class Client extends MatrixApi {
       }
       Object? syncError;
       await _checkSyncFilter();
+
+      // The timeout we send to the server for the sync loop. It says to the
+      // server that we want to receive an empty sync response after this
+      // amount of time if nothing happens.
       timeout ??= const Duration(seconds: 30);
+
       final syncRequest = sync(
         filter: syncFilterId,
         since: prevBatch,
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -1702,8 +1707,15 @@ class Client extends MatrixApi {
...
-          await syncRequest.timeout(timeout + const Duration(seconds: 10));
+
+      // The timeout for the response from the server. If we do not set a sync
+      // timeout (for initial sync) we give the server a longer time to
+      // responde.
+      final responseTimeout = timeout.inMilliseconds == 0
+          ? const Duration(minutes: 2)
+          : timeout + const Duration(seconds: 10);
+
+      final syncResp = await syncRequest.timeout(responseTimeout);
```
nico-famedly, line 1718: Do we even need a timeout for initial sync? It will fail eventually anyway, if there was a network error or such, right?

krille-chan: No, it is possible that it loads forever with the http package. Happens from time to time

lib/src/client.dart
```diff
@@ -1702,8 +1707,15 @@ class Client extends MatrixApi {
       });
       _currentSyncId = syncRequest.hashCode;
       onSyncStatus.add(SyncStatusUpdate(SyncStatus.waitingForResponse));
-      final syncResp =
-          await syncRequest.timeout(timeout + const Duration(seconds: 10));
+
+      // The timeout for the response from the server. If we do not set a sync
+      // timeout (for initial sync) we give the server a longer time to
+      // responde.
+      final responseTimeout = timeout.inMilliseconds == 0
```
nico-famedly, line 1714:
```suggestion
      final responseTimeout = timeout == Duration.zero
```

nico-famedly (approved):

lib/src/client.dart
krille-chan: ah thanks. I have searched for something like `Duration.isZero` and was confused that I didn't found this. Now I know why :D

Issue #1642, opened 2023-12-09
Some accounts cant load from new MatrixSdkDatabase


Airyzz:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

matrix-dart-sdk

### On which platform did the bug appear?

Android

### SDK Version

0.24.1

### Describe the problem caused by this bug

I was testing the new database, and for two of my testing accounts it's working flawlessly, but when attempting to use my main account, I get an error when loading from DB

Initial impression is that the account data is too large and is breaking db. So far I have only been able to reproduce this on Android, both on my physical device and on an emulator. Linux seems to be working fine.

Maybe there is an issue in my implementation, but I'm not sure.
For reference my implementation is [here](https://github.com/commetchat/commet/blob/3b5551a3fc4e3fc2401785300620b84cbbbf8a2f/commet/lib/client/matrix/matrix_client.dart#L244)

### Steps To Reproduce

1. Log in with large account (not 100% sure what exactly in the account data is causing the problem at this point)
2. Close app
3. Reopen app to load from database
4. Error occurs

### Screenshots or Logs

```
DatabaseException(Row too big to fit into CursorWindow requiredPos=46, totalRows=47) sql 'SELECT * FROM box_room_account_data' args []

#0      wrapDatabaseException (package:sqflite/src/exception_impl.dart:11)
<asynchronous suspension>
#1      SqfliteDatabaseMixin.txnRawQuery.<anonymous closure> (package:sqflite_common/src/database_mixin.dart:575)
<asynchronous suspension>
#2      BasicLock.synchronized (package:synchronized/src/basic_lock.dart:33)
<asynchronous suspension>
#3      SqfliteDatabaseMixin.txnSynchronized (package:sqflite_common/src/database_mixin.dart:506)
<asynchronous suspension>
#4      Box.getAllValues (package:matrix/src/database/sqflite_box.dart:145)
<asynchronous suspension>
#5      MatrixSdkDatabase.getRoomList.<anonymous closure> (package:matrix/src/database/matrix_sdk_database.dart:579)
<asynchronous suspension>
#6      Client.init (package:matrix/src/client.dart:1578)
<asynchronous suspension>
```

krille-chan:
[...]

PR #1652, opened 2023-12-18
fix: Workaround for too big sqflite tables

krille-chan:
Closes https://github.com/famedly/matrix-dart-sdk/issues/1642

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -177,6 +177,7 @@ class Client extends MatrixApi {
     this.shareKeysWithUnverifiedDevices = true,
     this.enableDehydratedDevices = false,
     this.receiptsPublicByDefault = true,
+    this.maxStoreReadReceiptsPerRoom = 100,
   })  : syncFilter = syncFilter ??
             Filter(
               room: RoomFilter(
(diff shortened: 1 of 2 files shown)


td-famedly (commented):
This looks very specific to read receipts. 

2 more general solution IMHO would be:

- Handle this on the client side https://stackoverflow.com/questions/51959944/sqliteblobtoobigexception-row-too-big-to-fit-into-cursorwindow-requiredpos-0-t

- or try to create a custom sqflite db as mentioned here: https://github.com/tekartik/sqflite/issues/779#issuecomment-1066623758

krille-chan:
> This looks very specific to read receipts.
> 
> 2 more general solution IMHO would be:
> 
> * Handle this on the client side https://stackoverflow.com/questions/51959944/sqliteblobtoobigexception-row-too-big-to-fit-into-cursorwindow-requiredpos-0-t
> * or try to create a custom sqflite db as mentioned here: [Window is full tekartik/sqflite#779 (comment)](https://github.com/tekartik/sqflite/issues/779#issuecomment-1066623758)

This can be done as well and then the limit can just be disabled on the client side. However this is only a workaround as there is currently no better solution for https://github.com/famedly/matrix-dart-sdk/issues/1642

nico-famedly:
I don't think just not storing receipts is a good solution. We could split them up instead of storing them as one large json, which would be better anyway, so that in the future we don't need to load the whole json when we just need a few receipts (like our own). Alternatively one of the approaches to increase the window size on android, since that seems to be the only platform with that problem.

krille-chan:
> I don't think just not storing receipts is a good solution. We could split them up instead of storing them as one large json, which would be better anyway, so that in the future we don't need to load the whole json when we just need a few receipts (like our own). Alternatively one of the approaches to increase the window size on android, since that seems to be the only platform with that problem.

Well I never said this is a good solution or any solution at all. It is a workaround as any other solution would need such a huge change in all databases that it would make much more sense to do this once we have reduced the amount of databases in the sdk. But we cannot reduce the amount of databases as long as the new database is blocked by this edge case. So what solution would you suggest which does not need another weeks of development?

krille-chan:
Closed because I found a better solution

Malesenka:
thx

PR #1655, opened 2023-12-19
chore: throw more threads at dart test

td-famedly:
note: this probably does not help us a lot atm given that the base github runner is a 2 core vm

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,15 +1,17 @@
 #!/bin/bash
+
+thread_count=$(getconf _NPROCESSORS_ONLN)
+
 if which flutter >/dev/null; then
     flutter pub global activate junitreport
-    flutter test --coverage
+    flutter test --concurrency=$thread_count --coverage
     TEST_CODE=$?
     
     # coverage
     flutter pub global activate remove_from_coverage
     flutter pub global run remove_from_coverage:remove_from_coverage -f coverage/lcov.info -r '\.g\.dart$'
 else
-    dart test --coverage=coverage
-    dart test --file-reporter='json:TEST-report.json'
+    dart test --concurrency=$thread_count --coverage=coverage
     TEST_CODE=$?
     
     # coverage -> broken see https://github.com/dart-lang/test/issues/1698


krille-chan:
So if `getconf _NPROCESSORS_ONLN` returns 2 in the github runner, but the default value of `--concurrency=<threads>` is 4, this will mean that we reduce the amount of threads instead of throwing **more** threads at dart test. I don't think that this will help much :-/

td-famedly:
> So if `getconf _NPROCESSORS_ONLN` returns 2 in the github runner, but the default value of `--concurrency=<threads>` is 4, this will mean that we reduce the amount of threads instead of throwing **more** threads at dart test. I don't think that this will help much :-/

Incorrect, the docs say. so right now it probably runs using 1 core? https://pub.dev/packages/test#test-concurrency

> Test suites run concurrently by default, using half of the host's CPU cores.

nico-famedly:
With this change:

grafik


Without:

grafik

So I would say it made it faster?

scripts/test.sh
```diff
@@ -1,15 +1,18 @@
...
     TEST_CODE=$?
     
     # coverage
     flutter pub global activate remove_from_coverage
     flutter pub global run remove_from_coverage:remove_from_coverage -f coverage/lcov.info -r '\.g\.dart$'
 else
-    dart test --coverage=coverage
-    dart test --file-reporter='json:TEST-report.json'
+    dart test --concurrency=$thread_count --coverage=coverage
+    dart test --concurrency=$thread_count --file-reporter='json:TEST-report.json'
```
nico-famedly, line 15: Do we run the tests twice here???

nico-famedly (approved):

krille-chan:
> Incorrect, the docs say. so right now it probably runs using 1 core? https://pub.dev/packages/test#test-concurrency

This is the output of `dart test --help`:
```sh
-j, --concurrency=<threads>           The number of concurrent test suites run.
                                      (defaults to "4")
```

But it makes sense. Because I have 8 cores, the default is only for my machine. Sorry, I thought this "default" was meant globally :D

scripts/test.sh
td-famedly: hm the tests do indeed run twice :D /2 the time again owo

nico-famedly (approved):

scripts/test.sh
```diff
@@ -1,15 +1,17 @@
...
+    flutter test --concurrency=$thread_count --coverage
     TEST_CODE=$?
     
     # coverage
     flutter pub global activate remove_from_coverage
     flutter pub global run remove_from_coverage:remove_from_coverage -f coverage/lcov.info -r '\.g\.dart$'
 else
-    dart test --coverage=coverage
-    dart test --file-reporter='json:TEST-report.json'
+    dart test --concurrency=$thread_count --coverage=coverage
```
nico-famedly, line 14: No file reporter anymore?

td-famedly: ye we don't use it

PR #1656, opened 2023-12-20
feat: lcov and tag olm tests

td-famedly:
Needs: https://github.com/famedly/frontend-ci-templates/pull/35
Fixes: https://github.com/famedly/product-management/issues/1933

ignore whitespaces, only concerning changes should be the tags, requesting keys before key verification manually and fake api changes


.gitlab-ci.yml
```diff
@@ -25,7 +25,7 @@ coverage:
     - sed -i 's/#\([^ ]\)/\1/g' pubspec.yaml
     - rm -r example
     - ./scripts/prepare.sh
-    - ./scripts/test.sh
+    - ./scripts/test.sh coverage
```
coder-with-a-bushido, line 28: wouldn't `which flutter >/dev/null` be `true` because of the flutter image?
If yes, then the `$1` doesn't get used?

td-famedly: correct, but why not /shrug :D

.github/workflows/app.yml
```diff
@@ -42,32 +42,58 @@ jobs:
...
+    strategy:
+      matrix:
+        app: [coverage, coverage-without-olm]
+    needs: [coverage, coverage_without_olm]
+    secrets: inherit
+    uses: famedly/frontend-ci-templates/.github/workflows/review-app.yml@td/artifactName
+    with:
+      projectname: "dart-sdk-${{ matrix.app }}"
+      pr: ${{ github.event.pull_request.number }}
+      environment: "review"
```
nico-famedly, line 95: Doesn't this mean the "view deployment" button leads to a random one?

Do you think it would be possible to merge the coverage instead like https://stackoverflow.com/questions/60544108/merge-lcov-results-in-one-report-and-keep-the-test-ids-testing-each-line ?

scripts/test.sh
```diff
@@ -2,33 +2,17 @@
...
-    dart pub global run remove_from_coverage:remove_from_coverage -f coverage/lcov.info -r '\.g\.dart$'
+    tagFlag="-t"
 fi
 
-# coverage html report
-genhtml -o coverage coverage/lcov.info || true
 
-# https://github.com/eriwen/lcov-to-cobertura-xml
-lcov_cobertura.py coverage/lcov.info || true
+dart test --concurrency=$thread_count --coverage=$1 $tagFlag 'olm'
```
nico-famedly, line 12: Wouldn't it make more sense in the NO_OLM case to just add the `-x olm` in that case?

scripts/test.sh
```diff
@@ -2,33 +2,17 @@
...
 
-exit $TEST_CODE
+# lets you do more stuff like reporton
+dart pub global activate coverage
+dart pub global run coverage:format_coverage --lcov -i $1 -o $1/lcov.info --report-on=lib/
+dart pub global activate remove_from_coverage
+dart pub global run remove_from_coverage:remove_from_coverage -f $1/lcov.info -r '\.g\.dart$'
```
nico-famedly, line 18: You should either check, that $1 is actually set to a valid directory or so, or give it some default value. I don't think we should write to /lcov.info or even worse, if you have a space in it allow injecting arbitrary parameters?

scripts/test.sh
```diff
@@ -2,33 +2,17 @@
...
 fi
 
-# coverage html report
-genhtml -o coverage coverage/lcov.info || true
 
-# https://github.com/eriwen/lcov-to-cobertura-xml
-lcov_cobertura.py coverage/lcov.info || true
+dart test --concurrency=$thread_count --coverage=$1 $tagFlag 'olm'
 
-exit $TEST_CODE
```
nico-famedly, line 34: Doesn't this break the error codes?

nico-famedly (changes requested):
The tests seem to look mostly okay, but the test script not so much. Haven't looked at the merging yet.

scripts/test.sh
td-famedly: Hm yes but then the coverage job would run both the non olm and the olm tests

td-famedly: hm oki, it shouldn't matter if they have olm or not but might as well test it I guess :D

nico-famedly (approved):

PR #1659, opened 2023-12-21
feat: Make client background service ready

krille-chan:
This makes it possible to init a client without starting the syncloop. Useful when using the same Client in a different isolate.


krille-chan:
Found a different approach to do this. See https://github.com/krille-chan/fluffychat/pull/733

PR #1661, opened 2023-12-22
feat: Make possible to fetch presence from database only

krille-chan:


lib/src/client.dart
```diff
@@ -2973,16 +2973,24 @@ class Client extends MatrixApi {
 
   /// The newest presence of this user if there is any. Fetches it from the
   /// database first and then from the server if necessary or returns offline.
-  Future<CachedPresence> fetchCurrentPresence(String userId) async {
+  Future<CachedPresence> fetchCurrentPresence(
+    String userId, {
+    bool fetchFromDatabase = true,
+    bool fetchFromServer = true,
```
coder-with-a-bushido, line 2979: Maybe I'm wrong but can we have just one option? Like `{ bool fetchOnlyFromCached = false }` and in that case even if what's in DB is null, just return `CachedPresence.neverSeen(userId)`.

krille-chan: Thanks, I would agree. A situation where we want to fetch only from cache makes sense but a situation where we do **not** want to fetch from cache but from the server sounds like we want to workaround a broken cache, which sounds like a bad pattern. I have changed it to only one option

coder-with-a-bushido (approved):

PR #1663, opened 2023-12-25
fix: typerror in removeEvent method from new database

krille-chan:


lib/src/database/matrix_sdk_database.dart
```diff
@@ -808,7 +808,8 @@ class MatrixSdkDatabase extends DatabaseApi {
     for (final key in keys) {
       final multiKey = TupleKey.fromString(key);
       if (multiKey.parts.first != roomId) continue;
-      final eventIds = await _timelineFragmentsBox.get(key) ?? [];
+      final eventIds =
+          List<String>.from(await _timelineFragmentsBox.get(key) ?? []);
```
coder-with-a-bushido, line 812: can we use `tryGetList<String>` instead?

krille-chan: no, that is for fetching a list from json. Here we just have an immutable list we get from the database which we need to transform to a mutable list. Otherwise it will throw this error: https://github.com/krille-chan/fluffychat/issues/735#issuecomment-1868953427

Everywhere else this is done with the `copyMap()` method but this box is the only box which returns a list and not a map. However as it is only a List<String> we do not need a deep copy and just use `List<String>.from` solves the problem

coder-with-a-bushido (approved):

PR #1670, opened 2023-12-30
refactor: Throw client init exception on client init fail

krille-chan:
This changes the behavior
when client init fails. It
no longer calls logout and does
only clear local data while
returning all available
information in a new
exception type so that the
SDK consumer can decide
to logout or try again to init
with these information. This
should make it much more rare
that users loose their sessions.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -33,6 +33,7 @@ import 'package:matrix/encryption.dart';
 import 'package:matrix/matrix.dart';
 import 'package:matrix/src/models/timeline_chunk.dart';
 import 'package:matrix/src/utils/cached_stream_controller.dart';
+import 'package:matrix/src/utils/client_init_exception.dart';
 import 'package:matrix/src/utils/compute_callback.dart';
 import 'package:matrix/src/utils/multilock.dart';
 import 'package:matrix/src/utils/run_benchmarked.dart';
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -1600,12 +1608,24 @@ class Client extends MatrixApi {
...
+        accessToken: accessToken,
+        userId: userID,
+        deviceId: deviceID,
+        deviceName: deviceName,
+        olmAccount: olmAccount,
+      );
+      await clear();
+      throw clientInitException;
+    } finally {
       _initLock = false;
```
nico-famedly, line 1628: The init lock will never be released in this case.

nico-famedly (commented):
I'm not quite sure on throwing the same exception for all those different cases. Some of them are clearly programmer errors and should possibly be an Error instead of an Exception, while others might be useful to handle. I guess I would like to see, how it gets used though, since possibly I am just not understanding the usecase.

lib/src/client.dart
krille-chan: But I don't understand why. I thought so because it was not in the finally block before but I don't understand the reason because `finally{}` is supposed to be executed at any condition.

Like this in dartpad works:

```dart
void main() {
  try {
    print('Test');
    throw 'Bumm';
  } catch(e) {
    rethrow; // Same with `throw 'Bumm 2';`
  } finally {
    print('Finally trotzdem ausgeführt'); // Appears in the output
  }
}
```

krille-chan:
> I'm not quite sure on throwing the same exception for all those different cases. Some of them are clearly programmer errors and should possibly be an Error instead of an Exception, while others might be useful to handle. I guess I would like to see, how it gets used though, since possibly I am just not understanding the usecase. But that is not really a blocking concern, I only didn't approve because of the initLock issue.

Do I understand this correctly that you would suggest to use a second Error class like `ClientInitPreconditionError`? Would sound like a good idea to me

lib/src/client.dart
nico-famedly: Hm, interesting. I guess that is my mistake then. It does NOT execute if there is a return, but it does execute when there is a throw... I have no idea why and no other language I know does that, so it always trips me up. I just don't think it is very intuitive to have control flow in a catch block AND a finally block. But this code is fine then, yes.

nico-famedly:
Yep, basically. Cases like db corruption should be handled by the developer, things like calling init or passing the wrong args twice should be fixed by the developer.

krille-chan:
> Yep, basically. Cases like db corruption should be handled by the developer, things like calling init or passing the wrong args twice should be fixed by the developer.

Alright :) How is this?

nico-famedly (approved):
Looks reasonable to me!

PR #1676, opened 2024-01-04
feat: add SQfLite encryption helper

TheOneWithTheBraid:
This patch introduces a helper class for SQfLite encryption related operations.

Most matrix clients will encrypt their sqlite database at rest. Since this is a quite fragmented task using the Flutter sqlite ecosystem, this helper aims to simplify some more complex operations.

It in particular helps with the following tasks :

- loading the correct shared objects / dynamic libraries for sqlcipher
- check whether a database is encrypted
- migrate an unencrypted SQLite database to SQLCipher
- apply the cipher to a database while opening it and ensure it loads

This code is not exactly matrix related, though presumably any matrix client will use parts of it.

Possible regressions :

- `package:sqlite3` became a direct dependency. As of now it already was a transitive dependency of the SDK itself.

### DIFF
--- a/lib/src/database/sqflite_encryption_helper/io.dart
+++ b/lib/src/database/sqflite_encryption_helper/io.dart
@@ -1,0 +1,220 @@
+import 'dart:ffi';
+import 'dart:io';
+import 'dart:math' show max;
+
+import 'package:sqflite_common/sqlite_api.dart';
+import 'package:sqlite3/open.dart';
+
+import 'package:matrix/matrix.dart';
+
+/// A helper utility for SQfLite related encryption operations
+///
+/// * helps loading the required dynamic libraries - even on cursed systems
+/// * migrates unencrypted SQLite databases to SQLCipher
+/// * applies the PRAGMA key to a database and ensure it is properly loading
+class SQfLiteEncryptionHelper {
+  /// the factory to use for all SQfLite operations
+  final DatabaseFactory factory;
+
+  /// the path of the database
+  final String path;
+
+  /// the (supposed) PRAGMA key of the database
+  final String cipher;
+
+  const SQfLiteEncryptionHelper({
+    required this.factory,
+    required this.path,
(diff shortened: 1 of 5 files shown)


lib/src/database/sqflite_encryption_helper/io.dart
```diff
@@ -0,0 +1,231 @@
...
+    if (!await file.exists()) {
+      return;
+    }
+
+    // no work to do in case the DB is already encrypted
+    if (!await _isPlainText(file)) {
+      return;
+    }
+
+    log(
```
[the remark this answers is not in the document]
TheOneWithTheBraid: Sure ...

lib/src/database/sqflite_encryption_helper/io.dart
```diff
@@ -0,0 +1,218 @@
...
+      // Make sure that we're actually using SQLCipher, since the pragma
+      // used to encrypt databases just fails silently with regular
+      // sqlite3
+      // (meaning that we'd accidentally use plaintext databases).
+      throw StateError(
+        'SQLCipher library is not available, '
+        'please check your dependencies!',
+      );
+    } else {
+      Logs().d('PRAGMA supported by bundled SQLite. Encryption supported.');
```
krille-chan, line 164: Just an idea, if we have a `cipherVersion` can we also print it here, like:

```dart
Logs().d('PRAGMA supported by bundled SQLite. Encryption supported with version: "$cipherVersion"');
```

TheOneWithTheBraid: Sounds good, applied.

TheOneWithTheBraid:
You should consider updating the Dart version in the CI. 3.0.6 is pretty old, but this is not related to this PR, downgrading the dependency constraint for `sqlite3` hence.

krille-chan (approved):

PR #1679, opened 2024-01-13
fix: type safety in Matrix Dart SDK database

TheOneWithTheBraid:
- removes unsafe casts
- builds list using List<T>.from constructor
- removes unwanted late statement

Got the following premium bug report sent via matrix : 

bug-report

### DIFF
--- a/lib/src/database/matrix_sdk_database.dart
+++ b/lib/src/database/matrix_sdk_database.dart
@@ -86,6 +86,7 @@ class MatrixSdkDatabase extends DatabaseApi {
   late Box<String> _seenDeviceIdsBox;
 
   late Box<String> _seenDeviceKeysBox;
+
   @override
   bool get supportsFileStoring => fileStoragePath != null;
   @override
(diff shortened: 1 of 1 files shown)


TheOneWithTheBraid:
Fixes: https://github.com/krille-chan/fluffychat/issues/794

lib/src/database/matrix_sdk_database.dart
```diff
@@ -433,7 +434,7 @@ class MatrixSdkDatabase extends DatabaseApi {
                 ? timelineEventIds.getRange(start, end).toList()
                 : []);
 
-        return await _getEventsByIds(eventIds.cast<String>(), room);
+        return await _getEventsByIds(eventIds, room);
```
krille-chan, line 437: This has already been fixed here: https://github.com/famedly/matrix-dart-sdk/commit/2afd7040a7fa7f8a65cab6196c0f56407334dc65

lib/src/database/matrix_sdk_database.dart
```diff
@@ -693,14 +694,10 @@ class MatrixSdkDatabase extends DatabaseApi {
...
-                  .toList()
-                  .cast<Map<String, dynamic>>(),
-              crossSigningEntries
-                  .where((c) => c != null)
-                  .toList()
-                  .cast<Map<String, dynamic>>(),
+              List<Map<String, dynamic>>.from(
+                  childEntries.whereType<Map<String, dynamic>>()),
+              List<Map<String, dynamic>>.from(
+                  crossSigningEntries.whereType<Map<String, dynamic>>()),
```
krille-chan, line 700: `List<T>.from()` is not better than `list.cast<T>()` as it crashes in the same way when there is an entry which is not `T`. The difference is that `List<T>.from()` create a non-deep copy. Do we need this here? Also this is unrelated to the bug described in the commit message

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1489,7 +1486,7 @@ class MatrixSdkDatabase extends DatabaseApi {
     try {
       await clear();
       await open();
-      final json = Map.from(jsonDecode(export)).cast<String, Map>();
+      final json = Map<String, Map>.from(jsonDecode(export));
```
krille-chan, line 1489: Same here. This does not prevent type errors or makes anything more type safe. It just creates a non-deep copy

PR #1683, opened 2024-01-15
chore: Add null checks to jsonPayload in event.dart

mErrenst:
Should fix https://famedly-ip.sentry.io/issues/4841856533


lib/src/event.dart
```diff
@@ -207,9 +207,9 @@ class Event extends MatrixEvent {
         stateKey: jsonPayload['state_key'],
         prevContent: prevContent,
         content: content,
-        type: jsonPayload['type'],
+        type: jsonPayload['type'] ?? '',
         eventId: jsonPayload['event_id'] ?? '',
-        senderId: jsonPayload['sender'],
+        senderId: jsonPayload['sender'] ?? '',
```
krille-chan, line 212: I don't agree with this and we should not allow this. If this happens then the json is incorrect and therefore we should throw an exception and then fix the reason for where this incorrect json was created and not mask the error here

mErrenst: Does this go for both changes? Then I would just close the PR without merging

krille-chan: > Does this go for both changes?

yes

mErrenst: Closed PR and changed Sentry issue back to unresolved

PR #1694, opened 2024-01-24
ci: Add tests for database on web

krille-chan:


td-famedly:
You probably also want to make `dart_web_compatible` and `database_web_tests` both required with this PR?

lib/src/database/indexeddb_box.dart
```diff
@@ -13,6 +13,8 @@ class BoxCollection with ZoneTransactionMixin {
 
   BoxCollection(this._db, this.boxNames, this.name);
 
+  static const bool isWeb = true;
```
td-famedly, line 16: Could the isWeb checks be replaced by the kIsWeb implementation? if they run on a chromedriver I assume kisWeb should be true. We already do this in some voip code.

https://api.flutter.dev/flutter/foundation/kIsWeb-constant.html

td-famedly: no but you can just do `bool.fromEnvironment('dart.library.js_util');`

krille-chan: oh really? Wasn't aware of this. That's great! Thank you so much :)

krille-chan: Have changed it :)

td-famedly (approved):

PR #1695, opened 2024-01-24
refactor: Use dart records for checkHomeserver method

krille-chan:
This makes use of dart
records to return three
different values for one
method. This makes the
special class no longer
necessary just for
returning data.


nico-famedly:
While I do think this is nice, is this small change worth breaking the API for? Maybe we should just bundle this in when we have other reasons to break the API?

nico-famedly (approved):

PR #1696, opened 2024-01-24
feat: famedly calls

td-famedly:
Fixes: https://github.com/famedly/matrix-dart-sdk/issues/347
Needed for: https://github.com/famedly/product-management/issues/1427
Needed for: https://github.com/famedly/product-management/issues/1606


td-famedly:
> Should this be reviewed or is this still a draft?

yes this can be reviewed now

lib/src/client.dart
```diff
@@ -1250,11 +1249,12 @@ class Client extends MatrixApi {
   /// onRoomEvent( "m.room.message", "!chat_id:server.com", "timeline", {sender: "@bob:server.com", body: "Hello world"} )
   final CachedStreamController<EventUpdate> onEvent = CachedStreamController();
 
-  /// The onToDeviceEvent is called when there comes a new to device event. It is
-  /// already decrypted if necessary.
```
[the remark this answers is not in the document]
td-famedly: added back

lib/src/voip/models/call_backend.dart
```diff
@@ -0,0 +1,75 @@
+abstract class CallBackend {
```
nico-famedly, line 1: Is this an extension point we want external users to be able to use? Or should this be sealed? (The factory method wouldn't work if external users extend this)

lib/src/voip/models/group_call_events.dart
```diff
@@ -0,0 +1,29 @@
...
+  final dynamic err;
+  GroupCallError(this.code, this.msg, this.err);
+
+  @override
+  String toString() {
+    return 'Group Call Error: [$code] $msg, err: ${err.toString()}';
+  }
+}
+
+class GroupCallState {
```
nico-famedly, line 15: This sounds like an enum.

lib/src/voip/models/group_call_events.dart
```diff
@@ -0,0 +1,29 @@
...
+  static String LocalCallFeedUninitialized = 'local_call_feed_uninitialized';
+  static String InitializingLocalCallFeed = 'initializing_local_call_feed';
+  static String LocalCallFeedInitialized = 'local_call_feed_initialized';
+  static String Entering = 'entering';
+  static String Entered = 'entered';
+  static String Ended = 'ended';
+}
+
+abstract class ICallHandlers {
```
nico-famedly, line 24: I don't think we use the I for interfaces atm? (nit)

lib/src/voip/models/participant.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:matrix/matrix.dart';
+
+class Participant {
```
nico-famedly, line 3: This should probably namespaced, like CallParticipant or so.

lib/src/voip/models/participant.dart
```diff
@@ -0,0 +1,48 @@
...
+      identical(this, other) ||
+      other is Participant &&
+          userId == other.userId &&
+          deviceId == other.deviceId;
+
+  @override
+  int get hashCode => userId.hashCode ^ deviceId.hashCode;
+
+  factory Participant.fromId(String id) {
+    final int lastIndex = id.lastIndexOf(':');
```
nico-famedly, line 27: Are you sure you can do that? Nothing prevents you from having a : in the deviceId afaik.

lib/src/voip/utils/famedly_call_extension.dart
```diff
@@ -0,0 +1,150 @@
...
+
+    memberships.forEach((key, value) {
+      for (final mem in value) {
+        if (!mem.isExpired) ids.add(mem.callId);
+      }
+    });
+    return ids.toList();
+  }
+
+  static const staleCallCheckerDuration = Duration(seconds: 30);
```
nico-famedly, line 81: Do we still need this?

lib/src/voip/utils/stream_helper.dart
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (track != null) {
+      await sender?.replaceTrack(track);
+    }
+  }
+}
+
+String roomAliasFromRoomName(String roomName) {
+  return roomName.trim().replaceAll('-', '').toLowerCase();
+}
```
nico-famedly, line 66: Why is this in voip?

lib/src/voip/utils/stream_helper.dart
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  for (var i = 0; i < list1.length; i += 1) {
+    if (list1[i] != list2[i]) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
nico-famedly, line 88: Is this a voip function? Should this really be public?

lib/matrix.dart
```diff
@@ -30,13 +30,21 @@ export 'src/database/sqflite_encryption_helper.dart';
...
+export 'src/voip/models/webrtc_delegate.dart';
+export 'src/voip/models/participant.dart';
+export 'src/voip/models/call_backend.dart';
+export 'src/voip/models/key_provider.dart';
+export 'src/voip/utils/conn_tester.dart';
+export 'src/voip/utils/voip_constants.dart';
+export 'src/voip/utils/rtc_candidate_extension.dart';
+export 'src/voip/utils/famedly_call_extension.dart';
+export 'src/voip/utils/types.dart';
+export 'src/voip/utils/wrapped_media_stream.dart';
```
nico-famedly, line 47: Utils generally implies, that something is used internally to help implementing something. Do we really want to export all the public names in those files?

lib/src/voip/utils/types.dart
```diff
@@ -0,0 +1,180 @@
+// ignore_for_file: constant_identifier_names
+
+class VoIPEventTypes {
```
nico-famedly, line 3: I guess this can't be an enum, since we need the string values quite often?

lib/src/voip/utils/types.dart
```diff
@@ -0,0 +1,180 @@
...
+  kConnected,
+
+  /// The call was received, but no processing has been done yet.
+  kRinging,
+
+  /// End of call
+  kEnded,
+}
+
+class CallErrorCode {
```
nico-famedly, line 45: this should be an enum.

lib/src/voip/utils/types.dart
```diff
@@ -0,0 +1,180 @@
...
+  final dynamic err;
+  CallError(this.code, this.msg, this.err);
+
+  @override
+  String toString() {
+    return '[$code] $msg, err: ${err.toString()}';
+  }
+}
+
+enum CallEvent {
```
nico-famedly, line 113: This is quite a confusing name.

lib/src/voip/utils/voip_constants.dart
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class CallConstants {
+  static const callEndedEventTypes = {
+    EventTypes.CallAnswer,
+    EventTypes.CallHangup,
+    EventTypes.CallReject,
+    EventTypes.CallReplaces,
+  };
+  static const ommitWhenCallEndedTypes = {
```
nico-famedly, line 48: Typo? Also, are any of these constants used anywhere? Couldn't find it at a glance.

lib/src/voip/utils/wrapped_media_stream.dart
```diff
@@ -0,0 +1,99 @@
...
+  MediaStream? stream;
+  final Participant participant;
+  final Room room;
+
+  /// Current stream type, usermedia or screen-sharing
+  String purpose;
+  bool audioMuted;
+  bool videoMuted;
+  final Client client;
+  final bool isWeb;
```
nico-famedly, line 18: We need this, because we can't depend on flutter?

lib/src/voip/voip.dart
```diff
@@ -6,199 +6,293 @@ import 'package:webrtc_interface/webrtc_interface.dart';
...
+    for (final callEvent in callEvents) {
+      await _callStreamByCallEvent(callEvent);
+    }
+  }
 
-      final confId = event.content['conf_id'];
-      final groupCall = groupCalls[confId];
-      if (groupCall == null) {
-        Logs().d('[VOIP] onToDeviceEvent: groupCall is null.');
+  Future<void> _callStreamByCallEvent(BasicEventWithSender event) async {
```
nico-famedly, line 143: I have no idea, what this function does based on its name. I would expect it to return a call stream, but it clearly does not :D

lib/src/voip/voip.dart
```diff
@@ -573,87 +647,118 @@ class VoIP {
...
+  /// Ideally only group calls would need setting both userId and deviceId to allow
+  /// having 2 devices from the same user in a group call
+  ///
+  /// For p2p call, you want to have all the devices of the specified `userId` ring
   ///
-  /// [type] The type of call to be made.
-  Future<CallSession> inviteToCall(String roomId, CallType type) async {
+  /// yes I forced you to call a particular user in a room, letting anyone answer
+  /// doesn't make sense
+  Future<CallSession> inviteToCall(
```
nico-famedly, line 662: Should we just require passing the room here?

lib/src/voip/voip.dart
```diff
@@ -539,16 +617,12 @@ class VoIP {
...
     }
 
     return CallType.kVoice;
   }
 
-  Future<bool> requestTurnServerCredentials() async {
-    return true;
-  }
-
   Future<List<Map<String, dynamic>>> getIceSevers() async {
```
nico-famedly, line 626: Typo, which took me a while when searching for this function :D

nico-famedly (commented):
Didn't review everything yet, but some initial comments

lib/src/client.dart
```diff
@@ -2080,9 +2046,17 @@ class Client extends MatrixApi {
         }
         await encryption?.handleToDeviceEvent(toDeviceEvent);
       }
+      if (toDeviceEvent.type.startsWith(
+          RegExp(r'm.call.|org.matrix.msc3401.call.|com.famedly.call.'))) {
```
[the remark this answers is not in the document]
td-famedly: msc3401 sends invite, candidates, etc over to-device events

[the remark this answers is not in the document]
nico-famedly: And it doesn't namespace them?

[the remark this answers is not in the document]
td-famedly: Well it does mention `m.call` here but that's the state event https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/group-voip/proposals/3401-group-voip.md#unstable-prefix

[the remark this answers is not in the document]
td-famedly: Although we already break it in some other places

lib/src/voip/voip.dart
td-famedly: well it's just `final room = client.getRoomById(roomId);` either in the client or in the sdk. idm either way

td-famedly (commented):
addressed comments round 1

lib/src/client.dart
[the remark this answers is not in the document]
nico-famedly: It only seems to mention that for the stable spec implementation, which does not exist yet. Until then we should only use and support the unstable prefix. That's why we have unstable prefixes and also assigning behaviour to a stable prefix before it is stable would make unstable prefixes useless.

lib/src/voip/call_session.dart
```diff
@@ -1984,43 +1692,43 @@ class CallSession {
...
+    if (isGroupCall && remoteUserDeviceId != null) {
       final toDeviceSeq = this.toDeviceSeq++;
+      final Map<String, Object> data = {
+        ...content,
+        'seq': toDeviceSeq,
+        'dest_session_id': remoteUserDeviceId!,
+        'sender_session_id': client.deviceID!,
+      };
 
       if (mustEncrypt) {
```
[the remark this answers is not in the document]
nico-famedly: Sending your IP unencrypted into a room is pretty bad, another option would be to not allow calls in unencrypted rooms, but that may be too heavy handed, I guess?

lib/src/voip/models/call_backend.dart
nico-famedly: It is an API, that we define. It is easier to prohibit extending this class until we have a use case for that instead of prohibiting it later. And if we don't want to expose it externally, we could even thing about making the class private. We expose quite a big API for calls and we really should make explicit, what should be called externally and what not :D

lib/src/voip/utils/types.dart
nico-famedly: You have several definitions, for what a "CallEvent" is. We should probably use different names for Events for calls and events for call state changes. Especially since we already have a "CallState" that maps very closely to this. How about naming this a "CallChange" instead? Or "CallStateChange"? I know that "CallEvent" isn't wrong, but it is very confusing when you already have Matrix events for calls.

lib/src/voip/utils/wrapped_media_stream.dart
nico-famedly: Yes, that would be better, since it would also be a constant allowing for better code generation.

lib/src/voip/voip.dart
nico-famedly: I would then suggest requiring to pass the room everywhere for consistency. Especially since we seem to be getting the room already just to get the id and then the room from the id again....

lib/src/room.dart
```diff
@@ -1844,22 +1844,16 @@ class Room {
...
   bool get groupCallsEnabled {
     final powerLevelMap = getState(EventTypes.RoomPowerLevels)?.content;
     if (powerLevelMap == null) return false;
-    return powerForChangingStateEvent(EventTypes.GroupCallMemberPrefix) <=
-            getDefaultPowerLevel(powerLevelMap) &&
-        powerForChangingStateEvent(EventTypes.GroupCallPrefix) <=
-            getDefaultPowerLevel(powerLevelMap);
+    return powerForChangingStateEvent(VoIPEventTypes.FamedlyCallMemberEvent) <=
+        getDefaultPowerLevel(powerLevelMap);
   }
```
nico-famedly, line 1857: Isn't this the same as canChangeStateEvent?

lib/src/voip/README.md
```diff
@@ -1,20 +1,98 @@
...
 `CallSession` objects are created by calling `inviteToCall` and `onCallInvite`.
 
-`GroupCall` objects are created by calling `createGroupCall`.
+`GroupCallSession` objects are created by calling `fetchOrCreateGroupCall`.
+
+## Group Calls
+
+All communication for group calls happens over to-device events except the `com.famedly.call.member` event.
+
+**To-device events must have a `party_id` or a `sender_session_id` set to the device id of the sender, this is partly attributed to making it easy to maintain all the kinds of calls and supporting calling between 2 devices of the same user.**
```
nico-famedly, line 26: Outdated?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  static const expireTsBumpDuration = Duration(seconds: 45);
+  static const activeSpeakerInterval = Duration(seconds: 5);
+
+  final Client client;
+  final VoIP voip;
+  final Room room;
+
+  /// is a list of backend to allow passing multiple backends in the future
+  /// we use the first backend everywhere as of now
+  final List<CallBackend> backends;
```
nico-famedly, line 49: You could always make it a list, once the code is ready for it and deprecate the single member then.

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  final String? application;
+  final String? scope;
```
nico-famedly, line 51: ? Oidc stuff? Could use a comment.

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  GroupCallState state = GroupCallState.localCallFeedUninitialized;
+  StreamSubscription<CallSession>? _callSubscription;
+
+  /// participant:volume
+  final Map<CallParticipant, double> audioLevelsMap = {};
+  CallParticipant? activeSpeaker;
+  WrappedMediaStream? localUserMediaStream;
+  WrappedMediaStream? localScreenshareStream;
+  String? localDesktopCapturerSourceId;
+  List<CallSession> callSessions = [];
```
nico-famedly, line 62: Wouldn't this be backend specific?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+
+  CallParticipant? get localParticipant => voip.localParticipant;
+
+  /// userId:deviceId
```
nico-famedly, line 66: ?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  List<CallParticipant> participants = [];
+  List<WrappedMediaStream> userMediaStreams = [];
+  List<WrappedMediaStream> screenshareStreams = [];
+  late String groupCallId;
```
nico-famedly, line 70: I don't think this needs to be late, but you might need a factory constructor then?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  Timer? memberLeaveEncKeyRotateDebounceTimer;
+
+  final CachedStreamController<GroupCallSession> onGroupCallFeedsChanged =
+      CachedStreamController();
+
+  final CachedStreamController<GroupCallState> onGroupCallState =
+      CachedStreamController();
+
+  final CachedStreamController<GroupCallEvent> onGroupCallEvent =
+      CachedStreamController();
```
nico-famedly, line 86: We got it, this is a group call... the prefix seems unnecessary?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+
+  final CachedStreamController<WrappedMediaStream> onStreamAdd =
+      CachedStreamController();
+
+  final CachedStreamController<WrappedMediaStream> onStreamRemoved =
+      CachedStreamController();
+
+  bool get isLivekitCall => backends.first is LiveKitBackend;
```
nico-famedly, line 94: If you did your "Backend" abstraction right, there should be no need for this.

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  Timer? resendMemberStateEventTimer;
+  Timer? memberLeaveEncKeyRotateDebounceTimer;
+
+  final CachedStreamController<GroupCallSession> onGroupCallFeedsChanged =
+      CachedStreamController();
+
+  final CachedStreamController<GroupCallState> onGroupCallState =
+      CachedStreamController();
+
+  final CachedStreamController<GroupCallEvent> onGroupCallEvent =
```
nico-famedly, line 85: How about you make this an actual event? As in have different classes, that have objects emitted, that describe the change like

```dart
class GroupCallMutedChanged {
  final bool muted;
}

class GroupCallStateChanged {
  final GroupCallState state;
}
```

etc?

lib/src/voip/group_call_session.dart
td-famedly: How do you mean? Atleast right now callSessions are only triggered by the mesh backend

lib/src/voip/group_call_session.dart
td-famedly: yes but also makes searching easier (and not specifying what files to search in :D)

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  WrappedMediaStream? localScreenshareStream;
+  String? localDesktopCapturerSourceId;
+  List<CallSession> callSessions = [];
+
+  CallParticipant? get localParticipant => voip.localParticipant;
+
+  /// userId:deviceId
+  List<CallParticipant> participants = [];
+  List<WrappedMediaStream> userMediaStreams = [];
+  List<WrappedMediaStream> screenshareStreams = [];
```
nico-famedly, line 69: Should these really be modifiable by external API consumers?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+        addScreenshareStream(localScreenshareStream!);
+
+        onGroupCallEvent.add(GroupCallEvent.localScreenshareStateChanged);
+        for (final call in callSessions) {
+          await call.addLocalStream(
+              await localScreenshareStream!.stream!.clone(),
+              localScreenshareStream!.purpose);
+        }
+
+        //await sendMemberStateEvent();
```
nico-famedly, line 431: Why is this commented out?

lib/src/voip/group_call_session.dart
```diff
@@ -0,0 +1,1361 @@
...
+  }
+
+  String get avatarName =>
+      getUser().calcDisplayname(mxidLocalPartFallback: false);
+
+  String? get displayName => getUser().displayName;
+
+  User getUser() {
+    return room.unsafeGetUserFromMemoryOrFallback(client.userID!);
+  }
```
nico-famedly, line 119: Should we really expose an unsafe function as safe? Is it safe in this context?

lib/src/voip/models/call_participant.dart
```diff
@@ -0,0 +1,57 @@
...
+  // factory CallParticipant.fromJson(Map<String, dynamic> json) =>
+  //     CallParticipant(
+  //       userId: json['userId'] as String,
+  //       deviceId: json['deviceId'] as String,
+  //     );
+
+  // Map<String, dynamic> toJson() => {
+  //       'userId': userId,
+  //       'deviceId': deviceId,
+  //     };
```
nico-famedly, line 56: dead code?

lib/src/voip/call_session.dart
```diff
@@ -609,12 +326,9 @@ class CallSession {
...
 
-    if (requestScreenshareFeed) {
-      await pc!.addTransceiver(
-          kind: RTCRtpMediaType.RTCRtpMediaTypeVideo,
-          init:
-              RTCRtpTransceiverInit(direction: TransceiverDirection.RecvOnly));
-    }
+    await pc!.addTransceiver(
+        kind: RTCRtpMediaType.RTCRtpMediaTypeVideo,
+        init: RTCRtpTransceiverInit(direction: TransceiverDirection.RecvOnly));
```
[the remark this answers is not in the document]
td-famedly: fixed

lib/src/voip/group_call_session.dart
td-famedly: ehh, sounds a bit overkill imo, usually clients can just setState and flutter should handle everything if keys were used properly, maybe a future task? (not exactly new to this PR :D)

lib/src/voip/group_call_session.dart
td-famedly: well it's cleaner but I am not a fan of passing GroupCallSession to everywhere that needs a backend agnostic function

lib/src/voip/group_call_session.dart
td-famedly: moved it just to mesh backend

lib/src/voip/group_call_session.dart
td-famedly: fixed

lib/src/voip/models/call_participant.dart
td-famedly: fixed

lib/src/voip/group_call_session.dart
td-famedly: fixed

lib/src/voip/group_call_session.dart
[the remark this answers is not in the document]
td-famedly: fixed

lib/src/voip/group_call_session.dart
td-famedly: fixed

lib/src/voip/group_call_session.dart
td-famedly: calls being scoped to user or room, added comment

lib/src/voip/README.md
td-famedly: edited

lib/src/room.dart
td-famedly: yes

lib/src/voip/voip.dart
td-famedly: done

lib/src/voip/utils/wrapped_media_stream.dart
td-famedly: done

lib/src/voip/models/call_backend.dart
td-famedly: hm sealed classes can't seem to be extended over different files of the same library? considering how big the files are now we should then keep it abstract

lib/src/voip/call_session.dart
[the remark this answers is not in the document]
td-famedly: Hm I would let clients handle this however they want (maybe show a warning message or so), as an sdk provider we should try and handle all cases

lib/src/client.dart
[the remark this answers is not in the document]
td-famedly: scoped to `com.famedly.call` now

lib/src/voip/utils/types.dart
td-famedly: makes sense, td likes CallStateChange changed

lib/src/room.dart
td-famedly: actually no, getDefaultPowerLevel checks if group calls are enabled for all other users with default power

lib/src/voip/group_call_session.dart
nico-famedly: Well, there is quite a bit of code, that listens to those events and then calls another function to check the proper state, which is error prone and confusing imo. But yeah, maybe that can be done later, although it would be another API break.

lib/matrix_api_lite/model/event_types.dart
```diff
@@ -96,4 +94,24 @@ abstract class EventTypes {
   // Spaces
   static const String spaceParent = 'm.space.parent';
   static const String spaceChild = 'm.space.child';
+
+  // MatrixRTC
+  static const String GroupCallMember = 'com.famedly.call.member';
```
nico-famedly, line 99: Why are these uppercae, but the others before it are all lower case?

nico-famedly (approved):
I still think the call change events should look different, but this looks good enough. Clearly lots of work was put into this! :)

lib/matrix_api_lite/model/event_types.dart
td-famedly: hm only spaceChild and spaceParent are lowerCase, do we want to fix that in this breaking release as well?

nico-famedly: Ah, well, I didn't uncollapse that far, I guess.

lib/src/voip/group_call_session.dart
td-famedly: moved to https://github.com/famedly/product-management/issues/1950 for now

PR #1697, opened 2024-01-25
fix: newly started call is terminated even before the first member event is sent

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/1711

### DIFF
--- a/lib/src/voip/voip_room_extension.dart
+++ b/lib/src/voip/voip_room_extension.dart
@@ -65,14 +65,17 @@ extension GroupCallUtils on Room {
     // whose state event we haven't recieved yet in sync.
     // (option 2 was local echo member state events, but reverting them if anything
     // fails sounds pain)
-
-    final expiredfr = groupCallMemberStateEvent.originServerTs
-            .add(staleCallCheckerDuration)
-            .millisecondsSinceEpoch <
-        DateTime.now().millisecondsSinceEpoch;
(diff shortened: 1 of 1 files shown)


td-famedly:
Stale call checks for the group calls are different than the normal calls. In group calls you check for the expires ts of each device/user and if all of them are expired terminate the call. Ideally both `.call` and `.call.member` events should be fired at the same time so the bug you mentioned shouldn't happen.

coder-with-a-bushido:
@td-famedly please check the logs attached in the issue. This is what I concluded from the logs; but maybe I'm wrong. I also tested out this fix and it worked for me.

td-famedly:
> @td-famedly please check the logs attached in the issue. This is what I concluded from the logs and from my debugging; but maybe I'm wrong. I also tested out this fix and it worked for me.

yes the fix works because it completely disables what staleCallChecker is supposed to do. 

`org.matrix.msc3401.call` events are not updated periodically, only member events are, staleCallChecker is supposed to check if there is a existing `org.matrix.msc3401.call` event, if yes find `org.matrix.msc3401.call.member` events for that call, then if all the member events are expires (looking at the expires_ts value) we terminate that call. `org.matrix.msc3401.call.member` values are updated periodically

The fix here would be delaying the staleCallChecker when starting the call, or just having some additional buffer in the expiry logic (which iirc should be already implemented)

PR #1705, opened 2024-02-14
Bump changelog to v0.25.9

mrmcoding88:


td-famedly (approved):
Hm the changes do remove public methods, but in hindsight, they should've been private. So the non breaking bump is probably fine

PR #1706, opened 2024-02-15
feat: add fromLocalStoreOnly to Event.downloadAndDecryptAttachment

Romain-Guillot:
Add optional `fromLocalStoreOnly` boolean parameter to `Event.downloadAndDecryptAttachment` to be able to retrieve the MatrixFile from the local store only without making an HTTP request.

It is sometimes desirable to display the image already stored in the database, without making an HTTP call if the image is not stored (and letting the application/user choose whether to download the image).


nico-famedly (approved):
Looks good to me, thank you!

PR #1708, opened 2024-02-16
fix: Use name of other participant on archived rooms

Pat992:
Previously archived rooms where called Empty chat (was <Your own username>).
Now the other username is selected in case of direct chats that are archived, via the room-member-states and client-userID

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -290,6 +290,19 @@ class Room {
           .calcDisplayname(i18n: i18n);
       if (sender != null) return sender;
     }
+    if (isArchived) {
+      // In case of a direct chat, we need to get the room-members via the RoomMember event-type
+      // then we have to make sure to get the other room-member that is not us, to add them to the display name.
+      final roomMemberEvents = states.entries
+          .firstWhereOrNull((element) => element.key == EventTypes.RoomMember);
+      final otherRoomMember = roomMemberEvents?.value.entries
+          .firstWhereOrNull((entry) => entry.value.senderId != client.userID);
+
+      if (otherRoomMember?.value.prevContent?.tryGet('displayname') != null) {
+        return i18n.wasDirectChatDisplayName(
+            otherRoomMember!.value.prevContent!.tryGet('displayname')!);
+      }
+    }
     if (membership == Membership.leave) {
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -290,6 +290,19 @@ class Room {
...
     }
+    if (isArchived) {
+      // In case of a direct chat, we need to get the room-members via the RoomMember event-type
+      // then we have to make sure to get the other room-member that is not us, to add them to the display name.
+      final roomMemberEvents = states.entries
+          .firstWhereOrNull((element) => element.key == EventTypes.RoomMember);
+      final otherRoomMember = roomMemberEvents?.value.entries
+          .firstWhereOrNull((entry) => entry.value.senderId != client.userID);
+
+      if (otherRoomMember?.value.prevContent?.tryGet('displayname') != null) {
```
td-famedly, line 301: why prevContent here?

lib/src/room.dart
```diff
@@ -290,6 +290,19 @@ class Room {
           .calcDisplayname(i18n: i18n);
       if (sender != null) return sender;
     }
+    if (isArchived) {
```
td-famedly, line 293: Could just go in the membership.leave block

lib/src/room.dart
```diff
@@ -290,6 +290,19 @@ class Room {
...
+      // In case of a direct chat, we need to get the room-members via the RoomMember event-type
+      // then we have to make sure to get the other room-member that is not us, to add them to the display name.
+      final roomMemberEvents = states.entries
```
td-famedly, line 296: hm idm this, but you could also just use directChatMatrixId to simplify this

lib/src/room.dart
```diff
@@ -291,6 +291,11 @@ class Room {
       if (sender != null) return sender;
     }
     if (membership == Membership.leave) {
+      if (directChatMatrixID != null) {
```
td-famedly, line 294: I would merge this with the lines just below this, looks like prev_sender is never sent in member events

td-famedly (changes requested):

krille-chan (approved):
LGTM 👍

PR #1712, opened 2024-02-22
fix: convert boxNames to List in clear function when creating transaction

ggurdin:
Currently, the boxNames object is a Set, which causes this error when calling it: "Unable to clear database - NotFoundError: Failed to execute 'transaction' on 'IDBDatabase': One of the specified object stores was not found."


krille-chan:
Thank you so much. According to our rules all contributions must be signed with one of our own keys. Can you give me right access so I can force push on your branch to sign your commit?

krille-chan (approved):

ggurdin:
Thank you! I've added you to my fork in case it comes up in the future.

PR #1714, opened 2024-02-23
chore: migrate to dart 3.3.0 and flutter 3.19.0

coder-with-a-bushido:


lib/src/models/receipts.dart
```diff
@@ -44,7 +44,7 @@ class Receipt {
   const Receipt(this.user, this.time);
 
   @override
-  bool operator ==(dynamic other) => (other is Receipt &&
+  bool operator ==(other) => (other is Receipt &&
```
nico-famedly, line 47: What did the analyzer warning actually say? Are you sure it didn't tell you to take an Object instead of dynamic here?

coder-with-a-bushido: the default when no type is specified here is `Object`. so, didn't change it.

lib/src/utils/crypto/native.dart
```diff
@@ -13,7 +13,7 @@ abstract class Hash {
   FutureOr<Uint8List> call(Uint8List data) {
     final outSize = EVP_MD_size(ptr);
     final mem = malloc.call<Uint8>(outSize + data.length);
-    final dataMem = mem.elementAt(outSize);
+    final dataMem = mem + outSize;
```
td-famedly, line 16: hm? 🤔

coder-with-a-bushido:
```
  /// Pointer arithmetic (takes element size into account).
  @Deprecated('Use operator + instead')
  Pointer<Uint8> elementAt(int index) =>
      Pointer.fromAddress(address + sizeOf<Uint8>() * index);

  /// A pointer to the [offset]th [Uint8] after this one.
  ///
  /// Returns a pointer to the [Uint8] whose address is
  /// [offset] times the size of `Uint8` after the address of this pointer.
  /// That is `(this + offset).address == this.address + offset * sizeOf<Uint8>()`.
  ///
  /// Also `(this + offset).value` is equivalent to `this[offset]`,
  /// and similarly for setting.
  @Since('3.3')
  Pointer<Uint8> operator +(int offset) =>
      Pointer.fromAddress(address + sizeOf<Uint8>() * offset);
```

lib/src/models/receipts.dart
coder-with-a-bushido: done

coder-with-a-bushido:
Closed in favour of https://github.com/famedly/matrix-dart-sdk/pull/1717. CI was not updated and was failing.

PR #1720, opened 2024-02-27
refactor: Allow calling init when in soft logout state

krille-chan:


lib/src/client.dart
```diff
@@ -43,7 +43,7 @@ import 'package:matrix/src/utils/try_get_push_rule.dart';
 
 typedef RoomSorter = int Function(Room a, Room b);
 
-enum LoginState { loggedIn, loggedOut }
+enum LoginState { loggedIn, loggedOut, softLoggedOut }
```
nico-famedly, line 46: I am pretty sure I wrote code somewhere, that assumes this only has 2 states, can we make it a minor bump and mention this as a breaking change?

lib/src/client.dart
```diff
@@ -1820,20 +1860,20 @@ class Client extends MatrixApi {
...
+
       // Call onSoftLogout 5 minutes before access token expires to prevent
       // failing network requests.
       final tokenExpiresAt = accessTokenExpiresAt;
       if (onSoftLogout != null &&
           tokenExpiresAt != null &&
-          tokenExpiresAt.difference(DateTime.now()) <= Duration(minutes: 5)) {
-        await onSoftLogout?.call(this);
+          tokenExpiresAt.difference(DateTime.now()) <= timeout * 2) {
+        await _handleSoftLogout();
```
nico-famedly, line 1874: If this function fails, no exception is thrown. This means we did call clear(), but try to continue with the sync. Should _handleSoftLogout throw some exception or otherwise return, that you are not signed in anymore?

lib/src/client.dart
krille-chan: Sounds good to me.

I searched for all "LoginState" in the app so far and I think I covered everything in my soft logout PR there. When working with an enum we often use `switch` which let the analyzer ring when there is a new not handeld case

lib/src/client.dart
krille-chan: I added a `rethrow`

lib/src/client.dart
krille-chan: https://github.com/famedly/matrix-dart-sdk/pull/1729

PR #1721, opened 2024-02-28
refactor: Store lastEvent in room object instead of room state

krille-chan:
closes https://github.com/famedly/matrix-dart-sdk/issues/360

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -216,8 +216,6 @@ class Client extends MatrixApi {
     importantStateEvents.addAll([
       EventTypes.RoomName,
       EventTypes.RoomAvatar,
-      EventTypes.Message,
-      EventTypes.Encrypted,
       EventTypes.Encryption,
       EventTypes.RoomCanonicalAlias,
       EventTypes.RoomTombstone,
@@ -2507,7 +2505,8 @@ class Client extends MatrixApi {
 
         // Is this event redacting the last event?
         if (event.type == EventTypes.Redaction &&
-            event.redacts == room.lastEvent?.eventId) {
+            (event.content.tryGet<String>('redacts') ?? event.redacts) ==
+                room.lastEvent?.eventId) {
           room.lastEvent?.setRedactionEvent(event);
           break;
         }
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -233,12 +233,8 @@ class Client extends MatrixApi {
       EventTypes.CallAnswer,
       EventTypes.CallReject,
       EventTypes.CallHangup,
-
-      /// hack because having them both in important events and roomPreivew
-      /// makes the statekey '' which means you can only have one event of that
-      /// type
-      // EventTypes.GroupCallPrefix,
-      // EventTypes.GroupCallMemberPrefix,
+      EventTypes.GroupCallPrefix,
+      EventTypes.GroupCallMemberPrefix,
     ]);
 
     // register all the default commands
@@ -2131,7 +2127,12 @@ class Client extends MatrixApi {
       final id = entry.key;
       final syncRoomUpdate = entry.value;
 
-      await database?.storeRoomUpdate(id, syncRoomUpdate, this);
+      // Is the timeline limited? Then all previous messages should be
+      // removed from the database!
+      if (syncRoomUpdate is JoinedRoomUpdate &&
+          syncRoomUpdate.timeline?.limited == true) {
+        await database?.deleteTimelineForRoom(id);
+      }
       final room = await _updateRoomsByRoomUpdate(id, syncRoomUpdate);
 
       final timelineUpdateType = direction != null
@@ -2202,6 +2203,7 @@ class Client extends MatrixApi {
           await _handleRoomEvents(room, state, EventUpdateType.inviteState);
         }
       }
+      await database?.storeRoomUpdate(id, syncRoomUpdate, room.lastEvent, this);
     }
   }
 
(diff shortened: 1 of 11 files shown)


lib/src/client.dart
```diff
@@ -2429,50 +2435,66 @@ class Client extends MatrixApi {
...
-                  importantStateEvents.contains(stateEvent.type);
-          if ((noMessageOrNoEdit || editingLastEvent || consecutiveEdit) &&
-              importantOrRoomLoaded) {
-            room.setState(stateEvent);
+        room.setState(Event.fromJson(eventUpdate.content, room));
+        break;
+      case EventUpdateType.state:
+        final event = Event.fromJson(eventUpdate.content, room);
+        if (event.stateKey != null) break;
+        // For invite states or post loaded rooms, we set everything in-memory.
```
nico-famedly, line 2444: This comment seems to be outdated.

lib/src/client.dart
```diff
@@ -2429,50 +2435,66 @@ class Client extends MatrixApi {
...
+            importantStateEvents.contains(event.type)) {
+          room.setState(event);
+        }
+        break;
+      case EventUpdateType.timeline:
+        final event = Event.fromJson(eventUpdate.content, room);
+        // For invite states or post loaded rooms, we set everything in-memory.
+        // For other rooms we only set "important" state events.
+        if (event.stateKey != null &&
+            (!room.partial || importantStateEvents.contains(event.type))) {
```
nico-famedly, line 2459: You seem to be duplicating a lot of code between the "state" and "timeline" cases. That makes it rather hard to spot bugs. For example this would not update state events we already store in memory.

lib/src/room.dart
```diff
@@ -153,50 +159,18 @@ class Room {
...
+    if ((stateKey == null || roomId == null) &&
+        !client.importantStateEvents.contains(state.type)) {
+      Logs().w(
+        'Tried to set a non state event with type "${state.type}" as state event for a room',
+      );
       return;
     }
 
-    (states[state.type] ??= {})[stateKey] = state;
+    (states[state.type] ??= {})[stateKey ?? ''] = state;
```
nico-famedly, line 173: How would the stateKey still be null here? Isn't the check for the importantStateEvents wrong here? Since we want to store all state events but only state events?

lib/src/client.dart
```diff
@@ -2153,6 +2158,7 @@ class Client extends MatrixApi {
           await _handleRoomEvents(room, state, EventUpdateType.inviteState);
         }
       }
+      await database?.storeRoomUpdate(id, syncRoomUpdate, room.lastEvent, this);
```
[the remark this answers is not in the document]
krille-chan: because the lastEvent of the room is calculated in the _handleRoomEvents method. If I want to keep this method where it was before, I would had to duplicate the logic to update the lastEvent. So... lazyness leaded to this :D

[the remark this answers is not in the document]
nico-famedly: I thought some of the _handleRoomEvents functions depended on the data already being persisted, but I guess that wasn't the case then! Thanks!

lib/src/client.dart
krille-chan: Fixed

lib/src/room.dart
krille-chan: It can be null if the incoming state is of type `m.room.message`. We still need to support it here as a fallback. The legacy rooms still use this to determine the last event. The new lastEvent slot is only set once a new message comes in.

Yes, I could have written a database migration, but I had no motivation to do this for three databases at once. Once we have reduced them, we could refactor this but actually it does not hurt to have this here. The complexity is not that huge and we still use this logic to determine the last event if we have no other last event for a room

lib/src/client.dart
krille-chan: Fixed

lib/src/room.dart
nico-famedly: But this is the setState call, couldn't we at least stop writing to it?

krille-chan: no, this is called on client init when loading the room from the database. This way we load the lastEvent from the previous sdk versions to memory

lib/src/client.dart
```diff
@@ -2466,50 +2472,58 @@ class Client extends MatrixApi {
...
+            room.lastEvent = event;
           }
+          break;
         }
+
+        // Ignore reactions for last event.
+        if (event.relationshipType == RelationshipTypes.reaction) break;
+
+        // Event is a valid new lastEvent:
+        room.lastEvent = event;
```
nico-famedly, line 2525: So, this leaves me a bit confused:

- You set the last event to event, if it is null (makes sense)
- You set the last event if the last event is not a preview event (makes sense)
- Then you skip everything, that isn't a room preview event (this seems to early to me)
- If the event is a redaction, it redacts the last event (good, but you already skipped redactions, since they are not in the preview events)
- If it is an edit, update last event (good)
- You skip reactions (good, but we never should have any normal reactions there, since their event type was not in the roomPreviewEvents)
- Then you set the last event

I guess most of that is fine, the only thing probably an issue is that redactions would need to be handled before the previewEvents check and that we might have a weird preview in that case?

lib/src/room.dart
nico-famedly: We should just set the last event then instead of writing it to the room state? And possibly only do that, when the last event is null? I guess that would have complications, if you have multiple events with a null state key, since this function doesn't know about that. But since we load events from the db into the room states, and that might include message events, we continue writing message events into the database and the room state on sync, which just seems confusing and not beneficial.

I guess the cleanest solution would be a database migration, but if you don't want to do that, at least document why this is still writing non-state events to the room state and what should be done about it in the future and maybe make the logic optional, if we already have a last event or so.

lib/src/client.dart
```diff
@@ -2190,6 +2195,7 @@ class Client extends MatrixApi {
           await _handleRoomEvents(room, state, EventUpdateType.inviteState);
         }
       }
+      await database?.storeRoomUpdate(id, syncRoomUpdate, room.lastEvent, this);
     }
   }
```
nico-famedly, line 2201: There isn't really a place to put it, but can you update the comment in the roomPreviewLastEvents.addAll call in this file and possibly resolve it, now that we don't set the state key to empty for those events anymore?

lib/src/client.dart
krille-chan: Thanks. You are right. I have fixed it by move the check if type is a room preview event type below the check for the redaction and removed the check for reaction as this is handled by the room preview event type anyway.

lib/src/client.dart
krille-chan: there is no comment in the roomPreviewLastEvents.addAll call

nico-famedly: I meant this one: https://github.com/famedly/matrix-dart-sdk/blob/cf4dac79693a55883591d5e968c69473d0b7f6ea/lib/src/client.dart#L237

lib/src/room.dart
krille-chan: We have multiple candidates for the lastEvent in the room state and the logic to pick the correct one is in the `Room.lastEvent` getter. We would need to completely rewrite this logic which would blow the PR even more up, if we want to set the lastEvent at this place, while we know that the legacy way is still working fine. Instead of a database migration we can just wait a little bit as the lastEvent picked by this logic is used to be stored in the database after the next sync for this room so this is somehow self healing already.

I can note this down there of course

lib/src/client.dart
krille-chan: Talked with @td-famedly about it and commented it in and removed the hack comment

lib/src/room.dart
krille-chan: Added a comment

lib/src/database/hive_collections_database.dart
```diff
@@ -1130,12 +1130,7 @@ class HiveCollectionsDatabase extends DatabaseApi {
             ? ''
             : eventUpdate.content['state_key'];
```
nico-famedly, line 1131: We probably shouldn't do this anymore, if we have a migration now.

nico-famedly (approved):

PR #1723, opened 2024-02-29
feat: hyperlink phone numbers in messages

coder-with-a-bushido:
Completes part of https://github.com/famedly/product-management/issues/1445

### DIFF
--- a/lib/src/utils/markdown.dart
+++ b/lib/src/utils/markdown.dart
@@ -204,6 +204,29 @@ class MentionSyntax extends InlineSyntax {
   }
 }
 
+class PhoneNumberSyntax extends InlineSyntax {
+  PhoneNumberSyntax()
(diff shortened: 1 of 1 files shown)


krille-chan (changes requested):
Are you sure this is the correct place for it? We are explicitly not doing link-detection in the sdk but we are doing it in the app for a reason that we don't want to send them as hyperlinks at this stage. This now does link detection for phone numbers but not for web-urls which would be inconsistent :-(

In the app the entry point would be here: https://github.com/famedly/app/blob/main/lib/widgets/html_message.dart#L61-L92

coder-with-a-bushido:
@krille-chan Thanks for pointing it out!

I'm closing this PR in favour of https://github.com/famedly/app/pull/5649

PR #1730, opened 2024-03-07
fix: add missing supported versions

krille-chan:


lib/src/client.dart
```diff
@@ -1235,7 +1235,18 @@ class Client extends MatrixApi {
...
+  static const Set<String> supportedVersions = {
+    'v1.1',
+    'v1.2',
+    'v1.3',
+    'v1.4',
+    'v1.5',
+    'v1.6',
+    'v1.7',
+    'v1.8',
+    'v1.9',
```
nico-famedly, line 1247: I think this is not quite true, because we don't support the new pushrules, which causes some different results when evaluating whether a message notifies you. Did you actually go through https://spec.matrix.org/v1.9/changelog/ ?

nico-famedly: Similarly we may need to verify we don't rely on the creator field anywhere.

nico-famedly: And if we implement the redaction rules somewhere, we may need to update those as well.

nico-famedly: I think we may also not support the new mac method yet (for verification).

nico-famedly: Without all of that, I think we can only claim 1.4 support.

krille-chan:
Will start from scratch here :D

PR #1741, opened 2024-03-21
chore: emit handleCallEnded on ice fail

td-famedly:


lib/src/voip/call.dart
```diff
@@ -1404,7 +1404,7 @@ class CallSession {
           await updateMuteStatus();
           missedCall = false;
         } else if (state == RTCIceConnectionState.RTCIceConnectionStateFailed) {
-          await hangup(CallErrorCode.IceFailed, false);
+          await hangup(CallErrorCode.IceFailed);
```
nico-famedly, line 1407: This (and group calls) seem to have been the only place where the parameter was set to false. Can we remove it now? Do you know why that was done originally?

td-famedly: I went through all of them and toggled them to true where I don't remember why they are false. 

IIRC shouldEmit was only supposed to handle `fireCallEvent(CallEvent.kHangup);` at first but has changed to do much more now. 

There's 2 other places where it's still set to false and I added a reasoning there.

nico-famedly (approved):

PR #1742, opened 2024-03-24
fix: Make database deleteable without the need to init the boxcollection

krille-chan:
Usually we want to delete a corrupted database. To do this
we need to init the BoxCollection first
which is not possible with a
corrupted database.

To fix this, this PR makes the
delete method a static function so that we can delete a
database at any time. This
makes it no longer necessary
to store the database factory
in the boxcollection so I
have removed it.


nico-famedly (approved):
It looks a bit awkward, but since this is a mostly internal API anyway, that should be fine?

PR #1753, opened 2024-04-04
fix: dart:io import in matrix_sdk_database

TheOneWithTheBraid:
- removes `Directory` field in high-level `MatrixSdkDatabase`
- migrates `Directory fileStoragePath` to `Uri fileStorageLocation`
- makes file operations in `MatrixSdkDatabase` conditional on `dart.libraries.js_util`
- implements a tiny stub of the file operations used in `MatrixSdkDatabase`

It seems like the Flutter tool can compile despite these imports. Sadly the Dart standalone dart2js compiler doesn't reach there. While refactorying the code, I decided it's likely cleaner to have a `Uri` as storage location provider than using some fake directory or String as relacement.

The advantage of a `Uri` at this place is the explicit `Uri.directory` constructor available to ensure type and encoding safe directory locations supporting both Windows and *nix.

Additionally, admitted, that's an edge-case, one could even easily extend the use of a `Uri` based descriptor to support future storage location accesses (e.g. IPFS or custom schemes for e.g. local web browser based file system APIs). Using a `Uri`, one would only need to override the three methods making use of the `fileStorageLocation` property to handle different Uri schemes too.

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -47,6 +47,11 @@ jobs:
           rm -r example
           ./scripts/prepare.sh
           ./scripts/test.sh
+      - name: Ensure SDK compiles on web
+        run: |
+          pushd web_test
+          dart pub get
+          dart run webdev build
 
   coverage_without_olm:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


krille-chan:
Is now merged in a different PR

PR #1754, opened 2024-04-04
refactor: Move file storage to mixin to not import dart:io

krille-chan:


lib/src/database/matrix_sdk_database.dart
```diff
@@ -153,10 +148,14 @@ class MatrixSdkDatabase extends DatabaseApi {
     this.database,
     this.idbFactory,
     this.sqfliteFactory,
-    this.maxFileSize = 0,
-    this.fileStoragePath,
-    this.deleteFilesAfterDuration,
-  });
+    int maxFileSize = 0,
+    Uri? fileStoragePath,
```
TheOneWithTheBraid, line 152: I'd propose to include the migration I added in my initial PR - this keeps the change non-breaking.

lib/src/database/database_file_storage_stub.dart
```diff
@@ -0,0 +1,21 @@
+import 'dart:typed_data';
+
+mixin DatabaseFileStorage {
+  bool get supportsFileStoring => fileStoragePath != null;
```
TheOneWithTheBraid, line 4: Using this stup is must always be `false`.

TheOneWithTheBraid (commented):
Two tiny proposals.

TheOneWithTheBraid:
Maybe it'd be useful to rebase this onto my PR ? This one lacks the tests I wrote to ensure the SDK will always build on web in future.

lib/src/database/database_file_storage_stub.dart
krille-chan: Would be better, yes. Will change it

lib/src/database/matrix_sdk_database.dart
krille-chan: Sounds good to me. One moment...

nico-famedly (changes requested):
Can you make sure that @TheOneWithTheBraid gets properly credited for reporting this issue, providing the tests and the initial implementation for a fix?

krille-chan:
> Can you make sure that @TheOneWithTheBraid gets properly credited for reporting this issue, providing the tests and the initial implementation for a fix?

How would you suggest?

The initial implementation is totally different from the new solution:

@TheOneWithTheBraid solution has created a new API for the `File` class which works on web and on io and has imported this to use it in the database class.

The new solution instead moves the code for File caching outside of the database class and makes it conditonally imported so that it works on io and web.

So there is nothing to credit her. I just copied some lines of comment to deprecate a member **after** she told me to do this.
The tests have been cherry-picked **after** she asked me to add them. The commit is still her's and signed off by her.

Credits for reporting the issue is hard if there was never opened one afaik.

TheOneWithTheBraid:
I mean, do however you want ; but you must admit you copied the whole explanation why to use a Uri, you copied the concept of using a Uri, you copied the migration. Sure nothing of this is complicated work. Nothing of this is hard to implement. Still you can be happy I signed a CLA, otherwise that'd be an obvious GPL attribution violation.

Since I signed a CLA I can't say anything against that, but this behavior is obviously not very open source of you.

The only thing you changed was moving the file operations into two other files - the rest is identically copied from my PR, while only the commit about the test is attributed. A simple rebase would have done the trick, it's not that difficult.

But I honestly don't care, I wrote this patch at work so handle it how ever you like.

krille-chan:
> I mean, do however you want ; but you must admit you copied the whole explanation why to use a Uri, you copied the concept of using a Uri, you copied the migration. Sure nothing of this is complicated work. Nothing of this is hard to implement. Still you can be happy I signed a CLA, otherwise that'd be an obvious GPL attribution violation. Since I signed a CLA I can't say anything against that, but this behavior is obviously not very open source of you. The only thing you changed was moving the file operations into two other files - the rest is identically copied from my PR, while only the commit about the test is attributed. A simple rebase would have done the trick, it's not that difficult. But I honestly don't care, I wrote this patch at work so handle it how ever you like.

I've edited the commit message. Can you say if you are happy with it? Doing it this way it was just easier as @nico-famedly explained before

TheOneWithTheBraid:
As mentioned before, I don't care, it's for work. It's only a question of respect.

The git-way of attributing this would be either rebasing or including my sign-off of the original commit - this way the history properly lists the external contribution. But I really don't care about it.

krille-chan:
> As mentioned before, I don't care, it's for work. It's only a question of respect.
> 
> The git-way of attributing this would be either rebasing or including my sign-off of the original commit - this way the history properly lists the external contribution. But I really don't care about it.

if you prefer it this way, I can do it. Just did it this way because this was faster (and doesn't need me to resign multiple previous commits) and I wasn't aware that this is important for you. Will take some time then as I will be unavailable over the next days

PR #1755, opened 2024-04-05
feat: Implement unpublished MSC custom refresh token lifetime

krille-chan:
Custom client implementation of
https://github.com/famedly/synapse/pull/10

### DIFF
--- a/lib/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart
+++ b/lib/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart
@@ -1,0 +1,56 @@
+import 'dart:convert';
+
+import 'package:http/http.dart';
+
+import 'package:matrix/matrix.dart';
+
+extension MscUnpublishedCustomRefreshTokenLifetime on MatrixApi {
+  static const String customFieldKey = 'com.famedly.refresh_token_lifetime_ms';
+
+  /// Refresh an access token. Clients should use the returned access token
+  /// when making subsequent API calls, and store the returned refresh token
+  /// (if given) in order to refresh the new access token when necessary.
+  ///
+  /// After an access token has been refreshed, a server can choose to
+  /// invalidate the old access token immediately, or can choose not to, for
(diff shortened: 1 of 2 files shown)


lib/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart
```diff
@@ -0,0 +1,55 @@
+import 'dart:convert';
+
+import 'package:http/http.dart';
+import 'package:matrix/matrix.dart';
+
+extension MscUnpublishedCustomRefreshTokenLifetime on MatrixApi {
+  static const String customFieldKey = 'com.famedly.refresh_token_lifetime_ms';
```
td-famedly, line 7: nit, maybe rename the variable to refreshTokenLifetime?

td-famedly (approved):

lib/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart
krille-chan: Hmm, does this really have a benefit? That's already in the name if you take the full name. As it is a static variable the full name is:
`MscUnpublishedCustomRefreshTokenLifetime.customFieldKey`

td-famedly: hm actually why is even a static variable then? It is only used in one place inside the extension itself. Can we just use the value instead of having it in a variable?

krille-chan: Because this way you can fetch it if you need it. For example if you want to listen on changes you can filter the sync by this account data type

nico-famedly (approved):

nico-famedly:
Hm, something seems to trigger 2 refreshes in parallel, which then causes 2 different refresh tokens to be returned, which then causes me to get signed out:

grafik

This might need some locking OR we need to track down, why this causes parallel refreshes.

PR #1757, opened 2024-04-06
feat: add sendAudioEvent and isVoiceMessage method to msc extensions

coder-with-a-bushido:
As td suggested here -> https://github.com/famedly/app/pull/5749#discussion_r1546030031


lib/src/room.dart
```diff
@@ -653,6 +653,37 @@ class Room {
         threadLastEventId: threadLastEventId);
   }
 
+  /// Sends an audio file with appropriate info to this room. Returns the event
+  /// ID generated by the server for this event.
+  Future<String?> sendAudioEvent(
```
td-famedly, line 658: Should go into msc_extensions/ iirc until stable

td-famedly (dismissed):

lib/msc_extensions/extension_audio_events/audio_events.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:matrix/matrix.dart';
+
+extension AudioEventExtension on Event {
+  /// returns if the event is a voice message
+  bool get isVoiceMessage =>
+      content['org.matrix.msc2516.voice'] != null ||
+      content['org.matrix.msc3245.voice'] != null;
```
krille-chan, line 7: Can we maybe instead check if this is a `Map<String,Object?>`? If someone puts in an `int` in this or something else, I think I would expect that this getter returns false. What do you think?

lib/src/room.dart
td-famedly: makes sense

lib/msc_extensions/extension_audio_events/audio_events.dart
coder-with-a-bushido: sure, sounds good to me.

krille-chan (approved):

PR #1758, opened 2024-04-07
fix: Do not use invitor avatar_url for room

krille-chan:
When user A invites user B to
a group chat, the avatar getter
should not return the
avatar of the inviting user A.
This is not intuitive. A client might
decide to do so but IMO it makes
no sense to do this by default.
If the
room is a DM invite, user A's
avatar is already used in the
code block before.


td-famedly (approved):
lgtm

PR #1760, opened 2024-04-08
fix: userMediaConstraints

td-famedly:


nico-famedly (approved):
Looks good to me. The class isn't exported, so not a public API, although removing the UserMediaOptions is technicalls a breaking change.

PR #1764, opened 2024-04-12
Don't fail sync if a presence event contains an empty presence field

morguldir:
Observed by using FluffyChat and conduit/conduwuit as the server

Example:
```
@nomad:daedric.net: {type: m.presence, content: {currently_active: false, last_active_ago: 1712913753736, presence: , status_msg...}
```


krille-chan:
Looks good to me. Thank you so much :) to be able to merge this, we would need to sign the commit with one of our company-internal keys. Can we get write access to your branch?

morguldir:
Allow edits by maintainers was checked already, but invited you as collaborators just in case

krille-chan (approved):

PR #1765, opened 2024-04-15
fix: canChangePowerLevel should return true for own user

krille-chan:


lib/src/user.dart
```diff
@@ -172,7 +172,8 @@ class User extends Event {
   /// Whether the client is allowed to change the power level of this user.
   /// Please be aware that you can only set the power level to at least your own!
   bool get canChangePowerLevel =>
-      room.canChangePowerLevel && powerLevel < room.ownPowerLevel;
+      (room.canChangePowerLevel && powerLevel < room.ownPowerLevel) ||
```
td-famedly, line 175: I have no idea what this function is supposed to do, how is it different from room.canChangePowerLevel, a rename would be nice.

Also `powerLevel < room.ownPowerLevel` does not make sense

lib/src/user.dart
```diff
@@ -172,7 +172,8 @@ class User extends Event {
...
+      (id == room.client.userID && powerLevel > 0);
```
td-famedly, line 176: do you actually need `powerLevel > 0`?

lib/src/user.dart
```diff
@@ -172,7 +172,8 @@ class User extends Event {
   /// Whether the client is allowed to change the power level of this user.
   /// Please be aware that you can only set the power level to at least your own!
   bool get canChangePowerLevel =>
```
td-famedly, line 174: td took way too long to understand what this was supposed to do, can we change this to `canChangeUserPowerLevel` or something better?

lib/src/user.dart
krille-chan: Hmm in the case this are you self, you can change your own powerlevel only in one direction. So I assumed this makes no sense to return true if the powerlevel is 0

td-famedly: yes but you can always create lower power levels, the spec allows you `[-(2**53)+1, (2**53)-1]`

krille-chan: okay I checked it out. The correct calculation should be that you:
1. Have the permission to change power levels in general
2. Your power-level is higher than the power level of the target user OR target user is yourself!

lib/src/user.dart
krille-chan: Okay changed :)

PR #1766, opened 2024-04-16
chore: downgrade sqflite_common_ffi to support dart sdk v3.2.0

coder-with-a-bushido:
Right now, only dart sdk versions >= 3.3.0 are supported. This PR downgrades `sqflite_common_ffi` to `2.3.2+1` in order to support dart sdk v3.2.x.


td-famedly (approved):
probably fine

PR #1768, opened 2024-04-16
refactor: Use prevBatch from server for pagination in event search

krille-chan:
When using the searchEvent()
method it was impossible to
paginate to the next
chunk when one chunk had
0 returned events. This
fixes it by returning the
prevBatch and makes it
possible to insert the prevBatch
again. sinceEventId is still
working but now deprecated.

### DIFF
--- a/lib/src/timeline.dart
+++ b/lib/src/timeline.dart
@@ -546,17 +546,39 @@ class Timeline {
     }
   }
 
+  @Deprecated('Use [startSearch] instead.')
+  Stream<List<Event>> searchEvent({
+    String? searchTerm,
+    int requestHistoryCount = 100,
+    int maxHistoryRequests = 10,
+    String? sinceEventId,
+    int? limit,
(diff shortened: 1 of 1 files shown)


lib/src/timeline.dart
```diff
@@ -552,11 +552,14 @@ class Timeline {
   /// events can already be displayed.
   /// Override the [searchFunc] if you need another search. This will then
   /// ignore [searchTerm].
-  Stream<List<Event>> searchEvent({
+  /// Returns the List of Events and the next prevBatch at the end of the
+  /// search.
+  Stream<(List<Event>, String?)> searchEvent({
```
td-famedly, line 557: is this worth a breaking change? maybe deprecate this method and create a new one?

krille-chan: Good idea. Changed :)

td-famedly (approved):
LGTM

PR #1769, opened 2024-04-17
fix: Make room.setPower more type safe and avoid change powerlevel in RAM before sending request to server

krille-chan:
This fixes the bug that the
actual dart Map in the state
has been manipulated because
we have not worked with a
copy of the map. Also this
crashes if the powerlevelmap
would had a wrong type in
users.


td-famedly (approved):
LGTM

PR #1770, opened 2024-04-17
fix: Fetch invite state after restart app

krille-chan:
This makes sure that the
users for an invitation are
correctly loaded so that we
can display the avatar, the
room displayname and
wether the room is a direct
chat or not.

### Before:

After app restart invitation looked like this:

Bildschirmfoto vom 2024-04-17 10-41-38

### After:

After app restart with this fix:

Bildschirmfoto vom 2024-04-17 10-41-20

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -213,6 +213,12 @@ class Room {
   /// from the database, that you need to correctly calculate the displayname
   /// and the avatar of the room.
   Future<List<User>> loadHeroUsers() async {
+    // For invite rooms request own user and invitor.
+    if (membership == Membership.invite) {
+      final ownUser = await requestUser(client.userID!, requestProfile: false);
+      if (ownUser != null) await requestUser(ownUser.senderId);
+    }
+
     var heroes = summary.mHeroes;
     if (heroes == null) {
       final directChatMatrixID = this.directChatMatrixID;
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -213,6 +213,12 @@ class Room {
   /// from the database, that you need to correctly calculate the displayname
   /// and the avatar of the room.
   Future<List<User>> loadHeroUsers() async {
+    // For invite rooms request own user and invitor.
+    if (membership == Membership.invite) {
+      final ownUser = await requestUser(client.userID!, requestProfile: false);
+      if (ownUser != null) await requestUser(ownUser.senderId);
```
td-famedly, line 219: this looks like something requestUser already handles internally with requestProfile true?

lib/src/room.dart
```diff
@@ -265,10 +271,12 @@ class Room {
       return isDirectChat ? result : i18n.groupWith(result);
     }
     if (membership == Membership.invite) {
-      final sender = getState(EventTypes.RoomMember, client.userID!)
-          ?.senderFromMemoryOrFallback
-          .calcDisplayname(i18n: i18n);
-      if (sender != null) return sender;
+      final ownMember = unsafeGetUserFromMemoryOrFallback(client.userID!);
```
td-famedly, line 274: just to be clear this just refactors the code right? Do we also want to loadHeroUsers in getLocalizedDisplayname? or is that expected to be done by clients?

lib/src/room.dart
krille-chan: I don't see where 🤔 this assumes that the own m.room.member state event has a different senderId than the stateKey and the senderId is the inviting user. We want to fetch this user in any case for invites as this user is likely be necessary for calculating room name and avatar

lib/src/room.dart
krille-chan: This is still expected to be done in clients. getLocalizedDisplayname has always returned the name sync and not async so we cannot load hero users here

lib/src/room.dart
td-famedly: ah yes I have misread this then

lib/src/room.dart
td-famedly: Can you add a comment so that sdk consumers know about this then? Thank you :D

krille-chan: Added

td-famedly (approved):
LGTM

PR #1777, opened 2024-04-23
chore: release v0.28.0

td-famedly:


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: matrix
 description: Matrix Dart SDK
-version: 0.27.0
+version: 0.28.0
```
krille-chan, line 3: Do we have a breaking change here? In my PR I made sure that we only deprecate stuff, but not removing apis

td-famedly: yep we merged the huge calls change

PR #1779, opened 2024-04-26
chore: update voip readme

td-famedly:


lib/src/voip/README.md
```diff
@@ -36,10 +34,11 @@ Sends the `com.famedly.call.member` event to signal an active membership. The fo
             "backend": {
                 "type": "mesh"
             },
-            "call_id": "!qoQQTYnzXOHSdEgqQp:im.staging.famedly.de",
+            "call_id": "",
```
nico-famedly, line 37: Why is that an empty string now?

nico-famedly (approved):

lib/src/voip/README.md
td-famedly: was done to be compatible with element but now it's a thing. Is mostly sugar as it wasn't used anywhere for m.room scoped calls.

m.room scoped calls don't have a callid essentially, callid is only supposed to be used for m.user scoped calls.

PR #1786, opened 2024-05-03
refactor: Use strippedstatevent as base for room state and user class

krille-chan:


lib/src/client.dart
```diff
@@ -1312,7 +1312,8 @@ class Client extends MatrixApi {
 
   final CachedStreamController<Event> onGroupMember = CachedStreamController();
 
-  final CachedStreamController<Event> onRoomState = CachedStreamController();
+  final CachedStreamController<StrippedStateEvent> onRoomState =
+      CachedStreamController();
```
nico-famedly, line 1316: A stripped state event won't know about its room, which then makes this stream very confusing, since you can't know, what room a state event belongs to, doesn't it? But maybe that is used nowhere, since then you can in most cases just use the per room stream?

krille-chan: Ah true. We could move this streamcontroller to the Room class itself to address this

krille-chan: Have changed it now to that it also returns the room ID by using the new dart type feature

nico-famedly (approved):

PR #1788, opened 2024-05-06
refactor: delete not sent events without eventupdate stream workaround

krille-chan:
Fixes https://github.com/famedly/product-management/issues/1984

This fixes several problems. First
sending a fake event through the
onEventUpdate stream was not a
good design and lead to problems
if you expect a timeline event but
then are unable to build the
event from json. This now uses
a new stream in the Client which
is listened to in the timeline to
delete an event which should be
much more reliable.
It also now throws an exception
if deleting the event fails
instead of returning true or false.
A deprecation note is added.
It also removes the behavior of
silently delete a file event where
the file is not stored. This is
confusing for the user and the
event should stay in the timeline
until the user activly decides to
try to send it again or delete the
message.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1312,6 +1312,9 @@ class Client extends MatrixApi {
 
   final CachedStreamController<Event> onGroupMember = CachedStreamController();
 
+  final CachedStreamController<String> onCancelSendEvent =
+      CachedStreamController();
+
   /// When a state in a room has been updated this will return the room ID
   /// and the state event.
   final CachedStreamController<({String roomId, StrippedStateEvent state})>
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -322,27 +322,26 @@ class Event extends MatrixEvent {
     return receiptsList;
   }
 
-  /// Removes this event if the status is [sending], [error] or [removed].
-  /// This event will just be removed from the database and the timelines.
-  /// Returns [false] if not removed.
+  @Deprecated('Use [cancelSend()] instead.')
   Future<bool> remove() async {
-    final room = this.room;
-
-    if (!status.isSent) {
-      await room.client.database?.removeEvent(eventId, room.id);
-
-      room.client.onEvent.add(EventUpdate(
-        roomID: room.id,
-        type: EventUpdateType.timeline,
-        content: {
-          'event_id': eventId,
-          'status': EventStatus.removed.intValue,
-          'content': {'body': 'Removed...'}
-        },
-      ));
+    try {
+      await cancelSend();
       return true;
+    } catch (_) {
+      return false;
     }
-    return false;
+  }
+
+  /// Removes an unsent or yet-to-send event from the database and timeline.
+  /// These are events marked with the status `SENDING` or `ERROR`.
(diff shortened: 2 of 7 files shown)


lib/src/event.dart
```diff
@@ -322,27 +322,28 @@ class Event extends MatrixEvent {
...
     }
-    return false;
+  }
+
+  /// Removes this event from the database and a timeline.
+  /// This event will just be removed from the database and the timelines.
+  /// This is only possible for events which are not sent yet, so have a
+  /// status of `SENDING` or `ERROR`. Throws an exception if used for an already
+  /// sent event!
+  Future<void> delete() async {
```
coder-with-a-bushido, line 340: do you think naming this something like `cancelSend` makes it more obvious?

lib/src/event.dart
```diff
@@ -322,27 +322,28 @@ class Event extends MatrixEvent {
...
+      return false;
     }
-    return false;
+  }
+
+  /// Removes this event from the database and a timeline.
+  /// This event will just be removed from the database and the timelines.
+  /// This is only possible for events which are not sent yet, so have a
+  /// status of `SENDING` or `ERROR`. Throws an exception if used for an already
+  /// sent event!
```
coder-with-a-bushido, line 339:
```suggestion
  /// Removes an unsent or yet-to-send event from the database and timeline.
  /// These are events marked with the status `SENDING` or `ERROR`. 
  /// Throws an exception if used for an already sent event!
```

lib/src/event.dart
krille-chan: Good idea 🤔

krille-chan: Changed

lib/src/event.dart
krille-chan: Done

lib/src/client.dart
```diff
@@ -1312,6 +1312,9 @@ class Client extends MatrixApi {
 
   final CachedStreamController<Event> onGroupMember = CachedStreamController();
 
+  final CachedStreamController<String> onDeleteLocalEvent =
```
coder-with-a-bushido, line 1315: now that the method is called `cancelSend`, we should also use the same terminology everywhere. This could be called `onCancelEventSend` or something?

lib/src/timeline.dart
```diff
@@ -44,6 +44,7 @@ class Timeline {
   StreamSubscription<EventUpdate>? sub;
   StreamSubscription<SyncUpdate>? roomSub;
   StreamSubscription<String>? sessionIdReceivedSub;
+  StreamSubscription<String>? deleteEventSub;
```
coder-with-a-bushido, line 47: `cancelEventSendSub` ?

lib/src/timeline.dart
```diff
@@ -291,6 +293,16 @@ class Timeline {
     }
   }
 
+  void _deleteEvent(String eventId) {
```
coder-with-a-bushido, line 296: `_cleanUpCancelledEvent` ?

coder-with-a-bushido (approved):
LGTM except those tiny naming suggestions :)

PR #1791, opened 2024-05-09
refactor: Cache direct chat matrix ID

krille-chan:
The direct chat account data
content is a map from matrixId
to roomId(s). If we want to find
the roomId for a given userId,
then this is not a problem. But
the other way around, if we
want to check if a given roomId
is a direct chat or not, needs to
process all entries which can be
hundreds. Doing this very
often in an app costs a lot of
performance. This adds a simple
cache in the room object and
validates the cache every time
to make this more efficient.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -356,15 +356,27 @@ class Room {
     });
   }
 
+  String? _cachedDirectChatMatrixId;
+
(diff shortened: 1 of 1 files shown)


td-famedly (approved):
LGTM

PR #1796, opened 2024-05-12
fix: group call voipIds not considering backend, application and scope

td-famedly:
Fixes: https://github.com/famedly/product-management/issues/1998


td-famedly:
missing currentGroupCID setter impl

td-famedly:
welp this evolved into a breaking change

td-famedly:
maybe redo again

PR #1798, opened 2024-05-13
feat: lazy load global profile from db

nico-famedly:
When using the getProfileFromUserId with getFromRooms == true, it would always fail to load any profile from the database. This is because by default rooms are partial/not postLoaded, which means no state is in memory. However we usually also don't want to load all members in all rooms just to have a usable offline behaviour.

So this changes the function to first check the non-partial rooms. If none of them have the profile, it then starts looking for the specific user profile in the partial rooms. This should only load one member event, which hopefully is faster than loading all members in all rooms.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -953,21 +953,43 @@ class Client extends MatrixApi {
     }
 
     if (getFromRooms) {
-      final room = rooms.firstWhereOrNull((Room room) =>
-          room.getParticipants().indexWhere((User user) => user.id == userId) !=
-          -1);
-      if (room != null) {
-        final user =
-            room.getParticipants().firstWhere((User user) => user.id == userId);
+      final rooms = this.rooms;
+
+      User? foundUser;
+      // check in memory first
+      for (final room in rooms) {
+        if (!room.partial) {
+          foundUser = await room.requestUser(userId,
+              ignoreErrors: true, requestProfile: false, requestState: false);
+          if (foundUser != null) break;
+        }
+      }
+
+      // If no hit, check the database
+      if (foundUser == null) {
+        for (final room in rooms) {
+          if (room.partial) {
+            foundUser = await room.requestUser(userId,
+                ignoreErrors: true, requestProfile: false, requestState: false);
+            if (foundUser != null) break;
+          }
+        }
+      }
+
+      if (foundUser != null) {
         final profileFromRooms = Profile(
           userId: userId,
-          displayName: user.displayName,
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -953,21 +953,43 @@ class Client extends MatrixApi {
...
-      if (room != null) {
-        final user =
-            room.getParticipants().firstWhere((User user) => user.id == userId);
+      final rooms = this.rooms;
+
+      User? foundUser;
+      // check in memory first
+      for (final room in rooms) {
+        if (!room.partial) {
+          foundUser = await room.requestUser(userId,
```
krille-chan, line 962: this does request the user from the database and even from the server but we only want to check the cache, right? So we can do this synchroniously by:

```dart
foundUser = room.getState(EventTypes.RoomMember, userId)?.asUser;
```

lib/src/client.dart
```diff
@@ -953,21 +953,43 @@ class Client extends MatrixApi {
...
+              ignoreErrors: true, requestProfile: false, requestState: false);
+          if (foundUser != null) break;
+        }
+      }
+
+      // If no hit, check the database
+      if (foundUser == null) {
+        for (final room in rooms) {
+          if (room.partial) {
+            foundUser = await room.requestUser(userId,
```
krille-chan, line 972: This would try to request the user from server for each room which would result into 1000 failed requests if we have 1000 rooms where this user is not participating

lib/src/room.dart
```diff
@@ -1688,11 +1688,14 @@ class Room {
...
   /// lazy loading. If the user can't be found this method tries to fetch
-  /// the displayname and avatar from the profile if [requestProfile] is true.
+  /// the displayname and avatar from the server if [requestState] is true.
+  /// If that fails, it falls back to requesting the global profile if
+  /// [requestProfile] is true.
   Future<User?> requestUser(
     String mxID, {
     bool ignoreErrors = false,
     bool requestProfile = true,
+    bool requestState = true,
```
td-famedly, line 1698: nitpick but `requestState` for requesting a profile from server sounds like a really confusing name for that variable. Can you make this a bit more verbose?

lib/src/client.dart
nico-famedly: No, it would not, because both requestProfile and requestState are set to false. None of these requests should be doing any server requests.

nico-famedly:
I'm going to close this since updating this now will be a pain and the code below has sufficiently changed, that a new PR would make more sense.

PR #1800, opened 2024-05-13
fix: ice restart mechanism

td-famedly:


lib/src/voip/call_session.dart
```diff
@@ -1153,8 +1145,15 @@ class CallSession {
           // fix any state/race issues we had with sdp packets and cloned streams
           await updateMuteStatus();
           _missedCall = false;
-        } else if (state == RTCIceConnectionState.RTCIceConnectionStateFailed) {
-          await hangup(reason: CallErrorCode.iceFailed);
+        } else if ({
+          RTCIceConnectionState.RTCIceConnectionStateFailed,
+          RTCIceConnectionState.RTCIceConnectionStateDisconnected
+        }.contains(state)) {
+          if (iceRestartedCount < 3) {
```
nico-famedly, line 1152: where is this incremented?

lib/src/voip/call_session.dart
```diff
@@ -1153,8 +1145,16 @@ class CallSession {
...
+            await restartIce();
+            iceRestartedCount++;
```
nico-famedly, line 1154: Shouldn't this be decremented at some point? Otherwise a call will drop if your connection changes 3 times during the call, right?

nico-famedly (approved):
Seems reasonable

PR #1801, opened 2024-05-14
fix: ensureNotSoftLoggedOut must be called before network reqeust in syncFilter check

krille-chan:
The ensureNotSoftLoggedOut()
method must be called before
we do any network request.
With having the sync filter
check before, we send a
failing POST request with the
possible wrong access token
which then would let the app
logout itself.


coder-with-a-bushido (approved):
LGTM :)

PR #1803, opened 2024-05-15
refactor: Also load lastEvent sender in loadHeroUsers()

krille-chan:
This makes it more easy
to load all users necessary
to display a nice ListTile in a
chat list.
This also now checks if there
are actually users to load and
returns null instead of a
future then which would safe
one build action in a flutter
FutureBuilder which should
also safe performance.


krille-chan:
Found a different solution for this already

PR #1813, opened 2024-05-19
refactor: Deprecate Streams in favor of client.onSync

krille-chan:
This deprecates all streams which can
be created by using client.onSync
anyway for a more predictable state
handling in consuming apps.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -202,6 +202,7 @@ class Room {
 
   /// If something changes, this callback will be triggered. Will return the
   /// room id.
+  @Deprecated('Use `client.onSync` instead and filter for this room ID')
   final CachedStreamController<String> onUpdate = CachedStreamController();
 
   /// If there is a new session key received, this will be triggered with
@@ -1702,6 +1703,7 @@ class Room {
     final dbuser = await client.database?.getUser(mxID, this);
     if (dbuser != null) {
       setState(dbuser);
+      // ignore: deprecated_member_use_from_same_package
       onUpdate.add(id);
       return dbuser;
     }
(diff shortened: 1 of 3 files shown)


lib/src/room.dart
```diff
@@ -200,6 +200,7 @@ class Room {
 
   /// If something changes, this callback will be triggered. Will return the
   /// room id.
+  @Deprecated('Use `client.onSync` instead and filter for this room ID')
   final CachedStreamController<String> onUpdate = CachedStreamController();
```
nico-famedly, line 204: That is not quite the same, right? Since it triggers also when a member is loaded or when history is fetched?

krille-chan: in this case you can listen on `client.onRoomState(roomId, EventTypes.RoomMember)`. For history fetch this is not called. But also this is called every time when something happens to a room while the value in it is always the same roomId so I would count this stream as not well thought out and would like to get rid of it

nico-famedly: If you call requestUser, that calls room.onUpdate. I don't think the client.onRoomState gets called in that case? While I do agree, that this doesn't seem well designed, I do think there is currently no alternative to it available?

krille-chan: It does. In requestUser when `Room.onUpdate` is triggered, at the same time `Room.setState` is always called as well which then triggers `Room.client.onRoomState`

nico-famedly: Ah, yes, you're right. I guess this only leaves pagination, which might be fine.

nico-famedly (approved):
I would like to see this tested in the app first to see how much pain this causes, but otherwise this seems fine.

Issue #1587, opened 2023-10-20
Better storage handling for megolm sessions needing upload


krille-chan:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

In a background job we check for every sync which sessions we need to upload. This results in that we fetch all sessions from the database every few seconds. This is very inefficient and we should store them in a different box instead.

### Describe your ideal solution

Store a reference to the megolm sessions which need upload in a different box.

krille-chan:
Blocked until we decided for a database

PR #1814, opened 2024-05-20
refactor: Store not uploaded group sessions in its own database queue

krille-chan:
This should highly increase the
performance as right now for every sync
ALL inbound group sessions are loaded
from the database and checked if
they need an upload. So if a user
has 10k stored sessions locally, this
would probably let the whole application
lag a lot. This stores the sessions which
need upload in a different table now,
similar how we do this with the to
device queue.

Closes https://github.com/famedly/matrix-dart-sdk/issues/1587


td-famedly (approved):
owo nice one! LGTM :D

Issue #1812, opened 2024-05-18
Client initialization failed


NiallBunting:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### On which platform did the bug appear?

Android

### SDK Version

0.29.2

### Describe the problem caused by this bug

I'm trying to migrate over from `HiveCollectionsDatabase` to the new MatrixSdkDatabase. I don't know if there is documentation for how to migrate correctly but on the flutter doc page the example seems to not include the latest changes.

Anyway I'm running into `ClientInitPreconditionError` exceptions: `!!!CRITICAL!!! Client initialization failed - LateInitializationError: Field '_clientBox@1543351407' has not been initialized.`.

```
I/flutter (11549): [Matrix] !!!CRITICAL!!! Client initialization failed - LateInitializationError: Field '_clientBox@1543351407' has not been initialized.
I/flutter (11549): #0      MatrixSdkDatabase._clientBox (package:matrix/src/database/matrix_sdk_database.dart)
I/flutter (11549): #1      MatrixSdkDatabase.getClient.<anonymous closure> (package:matrix/src/database/matrix_sdk_database.dart:381:28)
I/flutter (11549): #2      runBenchmarked (package:matrix/src/utils/run_benchmarked.dart:33:28)
I/flutter (11549): #3      MatrixSdkDatabase.getClient (package:matrix/src/database/matrix_sdk_database.dart:379:7)
I/flutter (11549): #4      Client.init (package:matrix/src/client.dart:1586:44)
I/flutter (11549): <asynchronous suspension>
I/flutter (11549): #5      MatrixProviderModel.login (package:myproj/models/matrixprovider.dart:200:7)
I/flutter (11549): <asynchronous suspension>
I/flutter (11549): #6      main.<anonymous closure> (package:myproj/main.dart:210:7)
I/flutter (11549): <asynchronous suspension>
I/flutter (11549): 
```

### Steps To Reproduce

My init looks like this:

```
    _matrixClient = Client('MyClient',
        databaseBuilder: (_) async {
          final dir =
              await getApplicationSupportDirectory(); // Recommend path_provider package

          Database db = await openDatabase(
            dir.toString() + '/database.sql',
            password: "somepassword",
          );

          return MatrixSdkDatabase('matrixdatabase', database: db);
      },
      legacyDatabaseBuilder: (_) async {
         final dir =
[...]

PR #1816, opened 2024-05-21
docs: Update example with matrixsdkdatabase

krille-chan:
Closes https://github.com/famedly/matrix-dart-sdk/issues/1812

### DIFF
--- a/example/main.dart
+++ b/example/main.dart
@@ -2,6 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:matrix/matrix.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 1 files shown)


example/main.dart
```diff
@@ -2,6 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:matrix/matrix.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
+import 'package:sqflite/sqflite.dart';
```
nico-famedly, line 5: I think we should import this as sqlite, so that the openDatabase becomes `sqlite.openDatabase`. Otherwise users might confuse it and assume it is a function from our SDK.

krille-chan: Done :)

nico-famedly (approved):

PR #1824, opened 2024-05-22
preShareKey improvements

td-famedly:
fix: only fire backend.onNewParticipant and backend.onLeftParticipant when it is not the local participant, this fixes the issue where onNewParticipant would get triggered when it detects a new call even though you were not in the call, as of now there is no code in those functions which needs to be triggered before you have joined the call so this should be fine

chore: also improve participants join leave tracking logic

### DIFF
--- a/lib/src/voip/group_call_session.dart
+++ b/lib/src/voip/group_call_session.dart
@@ -48,7 +48,7 @@ class GroupCallSession {
   CallParticipant? get localParticipant => voip.localParticipant;
 
   List<CallParticipant> get participants => List.unmodifiable(_participants);
-  final List<CallParticipant> _participants = [];
+  final Set<CallParticipant> _participants = {};
 
   String groupCallId;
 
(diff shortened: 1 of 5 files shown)


lib/src/voip/group_call_session.dart
```diff
@@ -246,16 +246,27 @@ class GroupCallSession {
 
     if (anyJoined.isNotEmpty || anyLeft.isNotEmpty) {
```
nico-famedly, line 247: Is this not redundant?

td-famedly: need that for the `onGroupCallEvent.add(GroupCallStateChange.participantsChanged);` below

nico-famedly: Ah, no, it is not, because the `onGroupCallEvent.add(GroupCallStateChange.participantsChanged);` is still inside of this, but maybe it shouldn't be triggered, if the change was only local...

nico-famedly (approved):

Issue #1830, opened 2024-05-28
users somehow get an old version of powerlevels state


td-famedly:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [X] I already asked for help in the chat

### In which Project did the bug appear?

Famedly App

### On which platform did the bug appear?

Chrome

### SDK Version

v0.29.8

### Describe the problem caused by this bug

1 user has been randomly getting an older version of the powerlevels event. This automatically gets fixed after a clear cache or a refresh. This also happens after a postLoad. 

### Steps To Reproduce

no clue

### Screenshots or Logs

![image](https://github.com/famedly/matrix-dart-sdk/assets/152161658/f42dcd6b-b5cf-4f47-acdd-58ebbfcdcb21)

was an admin and had famedly calls enabled but got a old pl state

nikzen:
We also saw this behavior:

### Initial State

<img width="270" alt="image" src="https://github.com/famedly/matrix-dart-sdk/assets/33399346/ef9f5965-ddd9-4b6b-b18d-5e2c4ca7cfea">

### Delete Cache & Refresh Page 

<img width="361" alt="image" src="https://github.com/famedly/matrix-dart-sdk/assets/33399346/12175db2-c4c6-4d21-94a9-a26a0ca3db06">

After deleting cache & refresh page it worked

nico-famedly:
This should be fixed in 0.29.10, please reopen, if you can still reproduce it with that.

PR #1832, opened 2024-05-28
postLoad before setting famedly calls

td-famedly:
Fixes: https://github.com/famedly/matrix-dart-sdk/issues/1830

Found a missing postLoad from a maybe code path, this does not fix the admin indicator missing bug, you probably want to set powerlevels as a required event for that


krille-chan (changes requested):
hmm, no that shouldn't be the cause. If you are in a room, you have already post loaded it. If you set powerlevels as important state event, you do not need to postload it.

I would not set powerlevel as important state event as this then would load it for all rooms which is bad for the performance on app start.

PR #1834, opened 2024-05-29
chore: set ref for reusable workflow files

coder-with-a-bushido:
Needed for https://github.com/famedly/product-management/issues/1624

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -95,7 +95,7 @@ jobs:
     if: github.event_name == 'pull_request'
     needs: [merge_converage]
     secrets: inherit
-    uses: famedly/frontend-ci-templates/.github/workflows/review-app.yml@main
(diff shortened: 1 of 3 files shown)


.github/workflows/app.yml
```diff
@@ -95,7 +95,7 @@ jobs:
     if: github.event_name == 'pull_request'
     needs: [merge_converage]
     secrets: inherit
-    uses: famedly/frontend-ci-templates/.github/workflows/review-app.yml@main
+    uses: famedly/frontend-ci-templates/.github/workflows/review-app.yml@aa793b1fe31af74beb3f2a8e450a0da7dccde65c
```
td-famedly, line 98: Tags would probably be more readable :D

nico-famedly:
So, I don't really like this, because it feels like we would just forget to update those refs and doing that across a lot of repos is a pain. Could we instead use like branches or so for breaking changes?

PR #1836, opened 2024-05-29
feat: add participant changed streams

td-famedly:
Part of: https://github.com/famedly/product-management/issues/1950

Needed for: https://github.com/famedly/product-management/issues/2068


lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:matrix/matrix.dart';
+
+mixin MatrixRTCCallEvent {}
```
nico-famedly, line 3: why does this use a mixin?

nico-famedly (approved):

PR #1841, opened 2024-06-04
feat: retry call encryption key request logic

td-famedly:


lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:async';
+
+/// Retries the `retryFunction` after a set `timeInterval` until `dispose` is called
+class RetryEventModel {
```
nico-famedly, line 4: Does that need to be public? Also it has a very enterprisy name.

lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,26 @@
...
+  /// Time interval in milliseconds
+  Duration timeInterval;
```
nico-famedly, line 6: how is that in milliseconds?

lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,26 @@
...
+  RetryEventModel({
+    required this.timeInterval,
+    required this.retryFunction,
+  }) {
+    startTimer();
+  }
+
+  void startTimer() {
+    _timer = Timer.periodic(timeInterval, retryFunction);
+  }
```
nico-famedly, line 20: this probably shouldn't be public, considering that it doesn't cancel the currently running timer.

lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:async';
+
+/// Retries the `retryFunction` after a set `timeInterval` until `dispose` is called
+class RetryEventModel {
+  /// Time interval in milliseconds
+  Duration timeInterval;
+  void Function(Timer timer) retryFunction;
+
+  Timer? _timer;
```
nico-famedly, line 9: this seems like it could be final

lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:async';
+
+/// Retries the `retryFunction` after a set `timeInterval` until `dispose` is called
+class RetryEventModel {
+  /// Time interval in milliseconds
+  Duration timeInterval;
+  void Function(Timer timer) retryFunction;
```
nico-famedly, line 7: this variable seems unnecessary?

lib/src/models/retry_event_model.dart
td-famedly: Isn't exported though? (also I would say is a useful class, just not something the matrix package should provide I guess)

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -348,9 +360,13 @@ class LiveKitBackend extends CallBackend {
     for (final key in keyContent.keys) {
       final encryptionKey = key.key;
       final encryptionKeyIndex = key.index;
+      final p =
+          CallParticipant(groupCall.voip, userId: userId, deviceId: deviceId);
+      requestEncrytionKeyPending[p]?.dispose();
+      requestEncrytionKeyPending.remove(p);
```
nico-famedly, line 366:
```suggestion
      requestEncrytionKeyPending.remove(p)?.dispose();
```

lib/src/models/retry_event_model.dart
td-famedly: _RetryEventModel would be private to the file not the library, not exporting is what you want here

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -319,6 +322,15 @@ class LiveKitBackend extends CallBackend {
...
     );
+
+    for (final rp in remoteParticipants) {
+      requestEncrytionKeyPending.addAll({
+        rp: RetryEventModel(
+          timeInterval: Duration(seconds: 2),
+          retryFunction: (_) => requestEncrytionKey(groupCall, [rp]),
+        )
+      });
+    }
```
nico-famedly, line 333:
```suggestion
    
    requestEncrytionKeyPending.addAll({
      for (final rp in remoteParticipants)
        rp: RetryEventModel(
          timeInterval: Duration(seconds: 2),
          retryFunction: (_) => requestEncrytionKey(groupCall, [rp]),
        )
    });
```

Although I am not so sure about the dispose semantics in that case and you might want to go for:

```dart
for (final rp in remoteParticipants) {
    requestEncryptionKeyPending[rp] ??= RetryEventModel(
          timeInterval: Duration(seconds: 2),
          retryFunction: (_) => requestEncrytionKey(groupCall, [rp]),
        );
}
```

?

lib/src/models/retry_event_model.dart
nico-famedly: I guess, although you could also just put it in the file you use it in for now. But yeah, I guess that is fine.

lib/src/models/retry_event_model.dart
td-famedly: ye but timer.periodic callback is of time Function(timer), hence the useless vairable

lib/src/voip/backend/livekit_backend.dart
td-famedly: Looks basically the same to me, do you have any reasons to pick this over the existing one?

td-famedly: skipped the ??= because we don't want to throttle it in the sdk by chance

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -348,9 +359,12 @@ class LiveKitBackend extends CallBackend {
     for (final key in keyContent.keys) {
       final encryptionKey = key.key;
       final encryptionKeyIndex = key.index;
+      final p =
+          CallParticipant(groupCall.voip, userId: userId, deviceId: deviceId);
+      requestEncrytionKeyPending.remove(p)?.dispose();
```
nico-famedly, line 364: this doesn't seem like it would need to be constructed in every loop iteration?

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -336,21 +347,26 @@ class LiveKitBackend extends CallBackend {
...
       Logs().w(
           '[VOIP E2EE] Received m.call.encryption_keys where keys is empty: callId=$callId');
       return;
     } else {
       Logs().i(
-          '[VOIP E2EE]: onCallEncryption, got keys from $userId:$deviceId ${keyContent.toJson()}');
+          '[VOIP E2EE]: onCallEncryption, got keys from ${p.id} ${keyContent.toJson()}');
     }
 
+    requestEncrytionKeyPending.remove(p)?.dispose();
```
nico-famedly, line 362: How can you be sure, that you are not disposing of the wrong request here? I.e. you have:

1. undecryptable frame
2. request keys
3. stream change
4. request keys
5. answer for 2.
6. cancel request for 4.

Does that not matter, since the _setEncryptionKey will fire another undecryptable event?

td-famedly: moved to https://github.com/famedly/product-management/issues/2078

nico-famedly (approved):

Issue #1772, opened 2024-04-22
Room name/avatar replaced with older value after historical events emitted


PhantomRay:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Matrix dart SDK

### On which platform did the bug appear?

Android, iOS

### SDK Version

`0.27`

### Describe the problem caused by this bug

Room info replaced after `timeline. timeline.requestHistory()`. From this point onwards, the room's state is changed in memory and local database. 

### Steps To Reproduce

1. Sync rooms' data as usual
2. use timeline.requestHistory for a particular room
3. when the older events contain old room state like display name and avatar, it will override room's current value in memory and local database

```[tasklist]
### Tasks
```

krille-chan:
Hello @PhantomRay can you try this again with 0.27.0? We have refactored the logic there how to handle incoming events

PhantomRay:
Thanks a lot @krille-chan. I cannot test it directly in my app. But can certain use your demo app. 

We rely on modified matrix api lite which supports jwt. 

Hopefully you guys can support it very soon. Only a few lines of code.

PhantomRay:
Hi @krille-chan could you please treat this bug as high priority?
Because once it happens, re-launch the app will show incorrect room name and avatar.

krille-chan:
@PhantomRay I've tried to reproduce the bug in this way:

1. Create a new room in FluffyChat and invite another test account
2. Close the app
3. From the other test account change the room name and send more than 10 messages
4. Restart FluffyChat and load history because we received limited timeline
5. Room Name was still correct 🤔

Do you have better way how to reproduce this?

PhantomRay:
I think my scenario might be different to yours, because you must be editing data using client API. 
In my case, Matrix is integrated with our own system. We rely on admin API to push display name and avatar changes to matrix. 

We use `PUT /_synapse/admin/v2/users/<user_id>` and change only `displayname` and `avatar_url`.

But either way, this shouldn't happen.

PhantomRay:
Btw, I'm currently testing direct chat (1 to1) only.

nico-famedly:
This should be fixed in 0.29.10, please reopen, if you can still reproduce it with that.

PhantomRay:
Thanks a lot!

PhantomRay:
[...]

PR #1846, opened 2024-06-10
fix: state updates being also applied when fetching history

nico-famedly:
This could sometimes lead to the room name or the names of members in the room being incorrect.

### DIFF
--- a/lib/src/database/hive_database.dart
+++ b/lib/src/database/hive_database.dart
@@ -1063,7 +1063,11 @@ class FamedlySdkHiveDatabase extends DatabaseApi with ZoneTransactionMixin {
 
     final stateKey = eventUpdate.content['state_key'];
     // Store a common state event
-    if (stateKey != null) {
+    if (stateKey != null &&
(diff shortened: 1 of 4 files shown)


lib/src/database/hive_database.dart
```diff
@@ -1063,7 +1063,11 @@ class FamedlySdkHiveDatabase extends DatabaseApi with ZoneTransactionMixin {
 
     final stateKey = eventUpdate.content['state_key'];
     // Store a common state event
-    if (stateKey != null) {
+    if (stateKey != null &&
+        // Don't store events as state updates when paginating backwards.
```
td-famedly, line 1067: Could use {types}.contains() here

td-famedly (approved):
LGTM except nitpick

PhantomRay:
As mentioned in #1772, it's not fixed:

> In 0.29.10, it's still not fixed. Just to re-iterate, both of the following are overridden by older events:
> * room name/avatar
> * user name/avatar

PR #1848, opened 2024-06-11
fix: Do not return invalid state events from database

PhantomRay:
Same fix from https://github.com/famedly/matrix-dart-sdk/pull/1835 has to applied to other two places.


td-famedly (approved):
LGTM

nico-famedly:
can you amend your commit to match the conventional commit format (like the title of this PR for example)?

PhantomRay:
commit message fixed. @nico-famedly

nico-famedly:
Thanks!

PR #1852, opened 2024-06-12
fix: update last event properly on cancel send

td-famedly:
Fixes: https://github.com/famedly/product-management/issues/1684

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -225,6 +225,7 @@ class Event extends MatrixEvent {
     if (originalSource != null) {
       data['original_source'] = originalSource?.toJson();
     }
+    data['status'] = status.intValue;
     return data;
   }
 
@@ -349,12 +350,42 @@ class Event extends MatrixEvent {
   /// Removes an unsent or yet-to-send event from the database and timeline.
   /// These are events marked with the status `SENDING` or `ERROR`.
   /// Throws an exception if used for an already sent event!
+  ///
   Future<void> cancelSend() async {
     if (status.isSent) {
       throw Exception('Can only delete events which are not sent yet!');
     }
 
     await room.client.database?.removeEvent(eventId, room.id);
+
+    if (room.lastEvent != null && room.lastEvent!.eventId == eventId) {
+      final redactedBecause = Event.fromMatrixEvent(
+        MatrixEvent(
+          type: EventTypes.Redaction,
+          content: {'redacts': eventId},
+          redacts: eventId,
+          senderId: senderId,
+          eventId: '${eventId}_cancel_send',
+          originServerTs: DateTime.now(),
+        ),
+        room,
+      );
+
+      await room.client.handleSync(
+        SyncUpdate(
+          nextBatch: '',
+          rooms: RoomsUpdate(
+            join: {
+              room.id: JoinedRoomUpdate(
+                timeline: TimelineUpdate(
+                  events: [redactedBecause],
+                ),
+              )
+            },
+          ),
+        ),
+      );
+    }
     room.client.onCancelSendEvent.add(eventId);
   }
 
@@ -687,12 +718,14 @@ class Event extends MatrixEvent {
       await fetchSenderUser();
     }
 
-    return calcLocalizedBodyFallback(i18n,
-        withSenderNamePrefix: withSenderNamePrefix,
-        hideReply: hideReply,
-        hideEdit: hideEdit,
-        plaintextBody: plaintextBody,
-        removeMarkdown: removeMarkdown);
+    return calcLocalizedBodyFallback(
+      i18n,
+      withSenderNamePrefix: withSenderNamePrefix,
+      hideReply: hideReply,
+      hideEdit: hideEdit,
+      plaintextBody: plaintextBody,
+      removeMarkdown: removeMarkdown,
+    );
   }
 
   @Deprecated('Use calcLocalizedBody or calcLocalizedBodyFallback')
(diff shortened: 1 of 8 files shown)


lib/src/client.dart
```diff
@@ -2426,6 +2426,11 @@ class Client extends MatrixApi {
       case EventUpdateType.timeline:
         final event = Event.fromJson(eventUpdate.content, room);
 
+        final setLastEvent =
+            (event.unsigned?[messageSendingStatusKey] as int? ??
```
krille-chan, line 2430: why not just using `event.status`?

krille-chan (changes requested):
Not sure. Currently in FluffyChat I display the current sending event in the chat list with a loading indicator which makes it clear that in this room I currently try to send a message. This would be lost after merging this.

td-famedly:
> Not sure. Currently in FluffyChat I display the current sending event in the chat list with a loading indicator which makes it clear that in this room I currently try to send a message. This would be lost after merging this.

otoh I have checked telegram and whatsapp, both seem to only set the last even in chats list once sending was successful. My initial changes just reverted the lastEvent to the next possible lastEvent in the room timeline, but the team suggested just not to set the last event until sent which is much simpler.

lib/src/client.dart
```diff
@@ -2426,6 +2426,11 @@ class Client extends MatrixApi {
       case EventUpdateType.timeline:
         final event = Event.fromJson(eventUpdate.content, room);
 
+        final setLastEvent =
```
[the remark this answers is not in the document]
td-famedly: aha even better, thanks :D

krille-chan:
> otoh I have checked telegram and whatsapp, both seem to only set the last even in chats list once sending was successful.

I can reproduce on WhatsApp (iOS) that the message is already set as last event even when it is not sent yet. Sent a message in airplane mode and can see it in the chat list. Deleting it changes the last event to "You have deleted this message".

td-famedly:
> > otoh I have checked telegram and whatsapp, both seem to only set the last even in chats list once sending was successful.
> 
> I can reproduce on WhatsApp (iOS) that the message is already set as last event even when it is not sent yet. Sent a message in airplane mode and can see it in the chat list. Deleting it changes the last event to "You have deleted this message".

I tried to send a huge file, closed and opened the app which caused sending to fail. Then deleted the errored file send, the last event then was the previous event, not "You have deleted this message".

Telegram on the other hand does this for every message, possibly due to lack of chat list last event in sending state preview feature.

lib/src/event.dart
```diff
@@ -356,6 +356,20 @@ class Event extends MatrixEvent {
 
     await room.client.database?.removeEvent(eventId, room.id);
     room.client.onCancelSendEvent.add(eventId);
+
+    if (room.lastEvent?.eventId == eventId) {
+      Logs().v('cancel send called on lastEvent, setting to null');
+      room.lastEvent = null;
```
krille-chan, line 362: Again this would move the room to the bottom of a chat list of potential 1000 rooms, leading to that the room magically disappears

krille-chan: Why not handling it like WhatsApp? There the last event then is displayed as redacted. Just redact the event without doing a network call

td-famedly: updated

td-famedly:
I think this is a better way to handle this, showing the deleted message as lastEvent would be confusing as it's a cancel and not a redact


https://github.com/user-attachments/assets/a689a2fe-b480-4fb5-9bda-c8b371739339

lib/src/database/hive_collections_database.dart
```diff
@@ -1613,6 +1613,78 @@ class HiveCollectionsDatabase extends DatabaseApi {
...
+      room.lastEvent!.setRedactionEvent(redactedBecause);
+      // and then store it in the db
+      await updateRoomBox();
+
+      // check if room was postLoaded
+      if (!room.partial) {
+        final newLastEvent =
+            (await room.getTimeline()).events.skip(1).firstWhereOrNull(
+                  (event) =>
+                      room.client.roomPreviewLastEvents.contains(event.type),
```
krille-chan, line 1671: You can't just load the last message like this because you don't know if in the timeline even is a potential last event candidate. There could be 100 member/call state events in between. That's why I said we should do it like whatsapp

td-famedly: > No the room is already postLoaded, so you probably have a visible event already. And incase the last 30 messages (the default output of getTImeline) does not have a proper last event, it then auotmatically fallsback to what whatsapp does and shows you "redacted"

As we decided this was too confusing I have removed it and kept it at https://github.com/famedly/matrix-dart-sdk/tree/td/intellgientfallbackstash for future viewing.

lib/src/event.dart
```diff
@@ -349,13 +349,38 @@ class Event extends MatrixEvent {
...
+    MatrixLocalizations i18n = const MatrixDefaultLocalizations(),
+  }) async {
     if (status.isSent) {
       throw Exception('Can only delete events which are not sent yet!');
     }
 
     await room.client.database?.removeEvent(eventId, room.id);
+
+    if (room.lastEvent?.eventId == eventId) {
+      room.lastEvent = Event(
```
krille-chan, line 363: that's not how I suggested it actually. I suggested to **redact** the lastEvent with a fakeEvent, not replace it. Then when it comes to the logic to generate the localizedBody of a redacted event, check the sending status and if it is not synced, then return `l10n. cancelledSend`

td-famedly: I don't see how that would affect the outcome 🤔 . `setRedactedEvent` just sets the redacted_because unsigned property and then we return a custom `Redacted` text from the .body getter. That is much more code to navigate rather than just having the .cancelSend() method set the 'Cancelled sending message' body.

krille-chan (approved):

PR #1853, opened 2024-06-13
chore: add localization for group call events

coder-with-a-bushido:
Needed for https://github.com/famedly/product-management/issues/1922

### DIFF
--- a/lib/src/utils/event_localizations.dart
+++ b/lib/src/utils/event_localizations.dart
@@ -240,6 +240,17 @@ abstract class EventLocalizations {
         event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n)),
     EventTypes.CallCandidates: (event, i18n, body) => i18n.sentCallInformations(
         event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n)),
+    EventTypes.GroupCallMember: (event, i18n, body) {
+      final memberships = event.room.getCallMembershipsFromEvent(event);
+      if (memberships
+          .takeWhile((membership) => !membership.isExpired)
+          .isEmpty) {
+        // if all the memberships have expired (regardless of call_id)
+        // it means that no call is currently active
+        return i18n.groupCallEnded;
+      }
+      return i18n.groupCallStarted;
+    },
     EventTypes.Encrypted: (event, i18n, body) =>
         _localizedBodyNormalMessage(event, i18n, body),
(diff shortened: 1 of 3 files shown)


lib/src/utils/event_localizations.dart
```diff
@@ -240,6 +240,11 @@ abstract class EventLocalizations {
         event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n)),
     EventTypes.CallCandidates: (event, i18n, body) => i18n.sentCallInformations(
         event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n)),
+    EventTypes.GroupCallMember: (event, i18n, body) {
+      final memberships = event.content.tryGetList('memberships');
```
td-famedly, line 244: also needs an expired check. I would just use `getCallMembershipsFromEvent` and do a .isExpired

lib/src/utils/event_localizations.dart
```diff
@@ -240,6 +240,11 @@ abstract class EventLocalizations {
...
+      if (memberships!.isEmpty) return i18n.groupCallEnded;
```
td-famedly, line 245: Hm that is actually incorrect, you could have 1 group call device indicate the end of 1 group call but then another device could be in a different call in the same room.

td-famedly: Deep checking between membership events could be a bit too advanced for our usecase though. For starters can we just do a callId + deviceId check? we can ignore the application and scope for now because they are mostly unused.

coder-with-a-bushido: I think we need the callId + deviceId check only in timeline. Here, we can just check the memberships and conclude
that some call is active if there's at least one membership that hasn't expired yet (regardless of callId). Else, we can conclude some call has ended.

coder-with-a-bushido: @td-famedly Pls let me know if this sounds good here

lib/src/utils/event_localizations.dart
coder-with-a-bushido: done

lib/src/utils/event_localizations.dart
td-famedly: Yes, but do we know where these localizations will be used? AFAIK we also fallback to these on the timeline. Moreover on every member event you will now get a group call started message while it was just a update to the existing member event.

PR #1854, opened 2024-06-14
fix: Request user causing state update loops for apps

nico-famedly:
This completely redoes the requestUser function.

It now doesn't stop requesting a user forever just because of a network error. But redundant calls are still deduplicated, however their parameters are taken into account now.

It also now only calls onUpdate and onRoomState when there is actually new state. This way apps can just listen to updates to rerender instead of having to implement the deduplication themselves.

It also now doesn't treat member events without membership as "join" anymore. And lookup errors don't cause an empty user to get inserted into the room state.

In general this should still request the profile from the server, if the displayname is unset in a leave event, but it should also allow the app to actually settle in the tests.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -18,7 +18,9 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:math';
 
+import 'package:async/async.dart';
 import 'package:collection/collection.dart';
 import 'package:html_unescape/html_unescape.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/room.dart
```diff
@@ -1648,102 +1650,171 @@ class Room {
...
+    if (requestState && foundUser == null) {
+      foundUser = await _requestSingleParticipantViaState(mxID,
+          ignoreErrors: ignoreErrors);
+    }
+
+    // If the user isn't found or they have left and no displayname set anymore, request their profile from the server
+    if (requestProfile) {
+      if (foundUser
+          case null ||
+              User(
```
td-famedly, line 1734: Is it possible to have a foundUser with membership leave/ban but displayName not null here?

lib/src/room.dart
```diff
@@ -1648,102 +1650,171 @@ class Room {
...
   }
 
-  final Set<String> _requestingMatrixIds = {};
-
-  /// Requests a missing [User] for this room. Important for clients using
-  /// lazy loading. If the user can't be found this method tries to fetch
-  /// the displayname and avatar from the profile if [requestProfile] is true.
-  Future<User?> requestUser(
+  // Internal helper to implement requestUser
+  Future<User?> _requestSingleParticipantViaState(
```
td-famedly, line 1654: _requestParticipantStateEventFromServer ?

lib/src/room.dart
```diff
@@ -1648,102 +1650,171 @@ class Room {
...
+        id,
+        EventTypes.RoomMember,
+        mxID,
+      );
 
-    // Is user already in cache?
-    var foundUser = getState(EventTypes.RoomMember, mxID)?.asUser(this);
+      // valid member events require a valid membership key
+      final membership = resp.tryGet<String>('membership', TryGet.required);
+      assert(membership != null);
```
td-famedly, line 1670: Why assert here?

lib/src/room.dart
nico-famedly: Possibly. Synapse unsets the displayname on leaves, but I don't think it has to be that way and nothing in the spec suggests it would.

lib/src/room.dart
nico-famedly: All of our "request" functions imply server requests, so that seems redundant? It also is an internal helper function, so the name doesn't seem that important.

lib/src/room.dart
nico-famedly: Because our tests actually had that field unset. Such events should never exist, but we might generate them internally, so this just makes sure such an event shows up as an error in tests and during development. Basically the assertion checks, that our asumption is correct. That's also what the comment explains.

lib/src/room.dart
td-famedly: Hm displayName isn't checked in the == override anyway, can we just remove it then?

nico-famedly: What == override? The original requirement was to load the profile, if we have no displayname (or avatar) set on a leave event, so that's what this does. If we always load the profile, then we would also fetch it, if we already fetched the displayname previously.

td-famedly: Makes sense, but the User object == override does not take the displayname and avatarUrl into account. So are you sure this still works as expected?

td-famedly (approved):
LGTM

PR #1857, opened 2024-06-17
fix: add main thread receipts to event.receipts getter

td-famedly:
superseeds: https://github.com/famedly/matrix-dart-sdk/pull/1631/files
fixes: https://github.com/famedly/product-management/issues/1020

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -312,12 +312,26 @@ class Event extends MatrixEvent {
             entry.value.timestamp))
         .toList();
 
-    final own = receipts.global.latestOwnReceipt;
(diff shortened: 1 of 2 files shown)


rmoradi88 (approved):
LGTM 🚀

td-famedly:
> Update: We are still only loading read receipts from the global timeline when using event.receipts getter. But I see receipts from Element. Will do some further investigation

need to test

td-famedly:
Seems to work https://github.com/famedly/product-management/issues/1020#issuecomment-2245507747, therefore will merge

PR #1863, opened 2024-06-25
refactor: Let searchEvents return a future instead of a stream

krille-chan:


td-famedly:
@krille-chan this is already done now correct? please reopen if not

PR #1864, opened 2024-06-25
refactor: Remove Matrix Connection Exception

krille-chan:
This makes no sense anymore.
In the past we tried to use this
Exception type to have a
unified exception for all
network errors. However, this
is only in use for the custom
request method and for every
other api call it is not. So to
have the same error messages
every time, we should just remove
it. This only produces confusion.


coder-with-a-bushido (approved):
I think we should consider this a breaking change and bump SDK version accordingly

PR #1867, opened 2024-06-26
feat: Automatically cancel typing indicators after 30 seconds

krille-chan:
fixes https://github.com/famedly/product-management/issues/2126

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -74,6 +74,8 @@ class Room {
 
   final _sendingQueue = <Completer>[];
 
+  Timer? _clearTypingIndicatorTimer;
+
   Map<String, dynamic> toJson() => {
         'id': id,
         'membership': membership.toString().split('.').last,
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -393,6 +395,16 @@ class Room {
     return lastEvent;
   }
 
+  void setEphemeral(BasicRoomEvent ephemeral) {
+    ephemerals[ephemeral.type] = ephemeral;
```
coder-with-a-bushido, line 399: Do you think this could remove a typing indicator incorrectly in a room when there are multiple people typing at the same time? Should we scope the timer to the sender of the ephemeral as well?

rmoradi88:
LGTM but afaik it's a new issue. Right? I'm wondering what changes cause this.

lib/src/room.dart
coder-with-a-bushido: Guess that won't happen because the timer would be reset with the latest ephemeral update.

coder-with-a-bushido: But that would mean:

A starts typing at 00s
B starts typing at 29s
The timer is reset now, so the ephemeral wouldn't clear A's typing status and both A & B would be shown typing.

Maybe we should scope to the sender id?

krille-chan: Yes that would be better in theory. However, it would be much more complex for a rare edge case 🤔 we would need to change the way how we handle ephemerals completely in the Room class

krille-chan: So we would lose the "We just store the ephemeral as it is" approach and go to a "handle typing indicators special". I don't wanna sound lazy but I would personally prefer a less complex solution with such a rare edge case. What do you think?

rmoradi88 (approved):
LGTM 🚀 
That would be very nice if you address all comments :)

lib/src/room.dart
coder-with-a-bushido: Hmm, sounds good to me. But do you mind writing an issue for this edge case somewhere so we can tackle it in the future with a different approach maybe?

I can also do that if you want. Please feel free to resolve this conversation :)

krille-chan: Would be great if you can do it <3

PR #1868, opened 2024-06-28
fix: Do not update lastEvent with state events

krille-chan:


rmoradi88 (approved):
LGTM

td-famedly (changes requested):
Can you please link the issue this is trying to fix or some more info on why we are doing this?

rmoradi88:
> Can you please link the issue this is trying to fix or some more info on why we are doing this?

The info is provided now. Will dismiss the review since it's blocking release.

PR #1870, opened 2024-07-01
fix: lastEvent after edit and redact

td-famedly:
Fixes: https://github.com/famedly/product-management/issues/2155


krille-chan (approved):
Thank you so much !! :)

Issue #1865, opened 2024-06-26
Cache well-known result in Client-Box


krille-chan:
### Preflight Checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

Multiple consumer apps are caching the well-known result with it's own logic. We could provide some functionality of this here.

### Describe your ideal solution

We should cache the result of a well-known request in the client-box and make it available in the Client object. We should also provide a method to update the cached value. We should not decide when to update the cache, this should be done by the app by calling this method.

PR #1875, opened 2024-07-11
feat: cache .well-known data

TheOneWithTheBraid:
- BREAKING: remove `DiscoveryInformation` from the `checkHomeserver` response tuple
- BREAKING: create `DatabaseApi.storeWellKnown` method
- BREAKING: create `DatabaseApi.getWellKnown` method
- add new getter `Client.wellKnown` containing cached `DiscoveryInformation`
- override `Client.homeserver` to invalidate `Client.wellKnown` in case the domain changed
- override `Client.getWellknown` to cache the resolved `DiscoveryInformation`
- add tests for well-known cache

Fixes: https://github.com/famedly/matrix-dart-sdk/issues/1865

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -25,7 +25,6 @@ import 'dart:typed_data';
 import 'package:async/async.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:http/http.dart' as http;
-import 'package:http/http.dart';
 import 'package:mime/mime.dart';
 import 'package:olm/olm.dart' as olm;
 import 'package:random_string/random_string.dart';
@@ -123,6 +122,24 @@ class Client extends MatrixApi {
 
   /// The timeout until a typing indicator gets removed automatically.
   final Duration typingIndicatorTimeout;
+
+  DiscoveryInformation? _wellKnown;
+
+  /// the cached .well-known file updated using [getWellknown]
+  DiscoveryInformation? get wellKnown => _wellKnown;
+
+  /// The homeserver this client is communicating with.
+  ///
+  /// In case the [homeserver]'s host differs from the previous value, the
+  /// [wellKnown] cache will be invalidated.
+  @override
+  set homeserver(Uri? homeserver) {
+    if (homeserver?.host != this.homeserver?.host) {
+      _wellKnown = null;
+      unawaited(database?.storeWellKnown(null));
+    }
+    super.homeserver = homeserver;
+  }
 
   Future<MatrixImageFileResizedResponse?> Function(
       MatrixImageFileResizeArguments)? customImageResizer;
@@ -529,6 +546,27 @@ class Client extends MatrixApi {
       homeserver = null;
       rethrow;
     }
+  }
+
+  /// Gets discovery information about the domain. The file may include
+  /// additional keys, which MUST follow the Java package naming convention,
+  /// e.g. `com.example.myapp.property`. This ensures property names are
+  /// suitably namespaced for each application and reduces the risk of
+  /// clashes.
+  ///
+  /// Note that this endpoint is not necessarily handled by the homeserver,
+  /// but by another webserver, to be used for discovering the homeserver URL.
+  ///
+  /// The result of this call is stored in [wellKnown] for later use at runtime.
+  @override
+  Future<DiscoveryInformation> getWellknown() async {
+    final wellKnown = await super.getWellknown();
+
+    // do not reset the well known here, so super call
+    super.homeserver = wellKnown.mHomeserver.baseUrl.stripTrailingSlash();
+    _wellKnown = wellKnown;
+    await database?.storeWellKnown(wellKnown);
+    return wellKnown;
   }
 
   /// Checks to see if a username is available, and valid, for the server.
@@ -1196,7 +1234,7 @@ class Client extends MatrixApi {
       path = '_matrix/media/v3/config';
     }
     final requestUri = Uri(path: path);
-    final request = Request('GET', baseUri!.resolveUri(requestUri));
+    final request = http.Request('GET', baseUri!.resolveUri(requestUri));
     request.headers['authorization'] = 'Bearer ${bearerToken!}';
     final response = await httpClient.send(request);
     final responseBody = await response.stream.toBytes();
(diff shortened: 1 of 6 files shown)


TheOneWithTheBraid:
Ping @krille-chan !

lib/src/client.dart
```diff
@@ -495,11 +512,9 @@ class Client extends MatrixApi {
       homeserver = homeserverUrl.stripTrailingSlash();
 
       // Look up well known
-      DiscoveryInformation? wellKnown;
       if (checkWellKnown) {
         try {
-          wellKnown = await getWellknown();
-          homeserver = wellKnown.mHomeserver.baseUrl.stripTrailingSlash();
```
krille-chan, line 502: This will change the behavior of `client.checkHomeserver(domain)` as this will now no longer connect to the baseUrl in the well-known file. As a client-developer, I just want to pass the input of the homeserver-textfield to `client.checkHomeserver(input)` and let the sdk handle this

lib/src/client.dart
```diff
@@ -121,6 +121,24 @@ class Client extends MatrixApi {
   /// The timeout until a typing indicator gets removed automatically.
   final Duration typingIndicatorTimeout;
 
+  DiscoveryInformation? _wellKnown;
```
krille-chan, line 124: Do we need to cache in the database AND in memory? I would leave the memory caching up to the client developer tbh. We already have waaay to many states to handle

TheOneWithTheBraid: I see but caching it at runtime it way more convenient since it can be accessed without async API. But however you like, I can remove it.

lib/src/client.dart
TheOneWithTheBraid: Ah, right, forgot to re-add it - I wanted to move it to the `getWellKnown` method but apparently forgot about it. Thanks.

TheOneWithTheBraid: Fixed it back.

lib/src/client.dart
```diff
@@ -121,6 +121,24 @@ class Client extends MatrixApi {
...
+
+  /// the cached .well-known file updated using [getWellknown]
```
td-famedly, line 126:
```suggestion
  /// the cached .well-known file updated using [getWellknown]
  /// to refresh call getWellknown() again
```

We could also refresh the cache everytime someone calls .wellKnown but do it in a non blocking way.

lib/src/client.dart
```diff
@@ -521,13 +536,40 @@ class Client extends MatrixApi {
...
+  /// Note that this endpoint is not necessarily handled by the homeserver,
+  /// but by another webserver, to be used for discovering the homeserver URL.
+  ///
+  /// In case is provided, [hostname] will be used in place of [homeserver] for
+  /// the HTTPS lookup. This is useful since the matrix homeserver does not
+  /// necessarily also host a copy of the `.well-known`.
+  ///
+  /// The result of this call is stored in [wellKnown] for later use at runtime.
+  @override
+  Future<DiscoveryInformation> getWellknown([String? hostname]) async {
```
krille-chan, line 561: This doesn't add any value for us now. I would prefer that this adds a cache-lifetime and returns the cached version if the lifetime has not expired, or fetches again from the server and stores it then in the db

lib/src/client.dart
krille-chan: yes IMO this can better be done by the app. We already have so many in-memory caches which leads to huge performance problems, that in the long term I would prefer to get rid of those and don't add more of them

@td-famedly @nico-famedly feel free to add your opinion here as well

lib/src/client.dart
```diff
@@ -521,13 +536,40 @@ class Client extends MatrixApi {
...
+    final wellKnown = await getDiscoveryInformationsByUserId(
+      hostname ?? baseUri!.host,
+    );
+
+    // do not reset the well known here, so super call
+    super.homeserver = wellKnown.mHomeserver.baseUrl.stripTrailingSlash();
```
nico-famedly, line 567: Changing the homeserver here might be a bit dangerous?  We already seem to set the homeserver in the checkHomeserver function, so for one this seems redundant, but if some developer actually passes a custom hostname to this function, it might break the client completely in an unexpected manner.

In general I would give this function only one purpose, i.e. "refreshWellknown" instead of "getWellknown". Then you could extract the hostname from the userid and request that and update the database. I don't think SDK users should be able to publicly call this with a different hostname.

lib/src/client.dart
```diff
@@ -521,13 +536,40 @@ class Client extends MatrixApi {
...
+  ///
+  /// In case is provided, [hostname] will be used in place of [homeserver] for
+  /// the HTTPS lookup. This is useful since the matrix homeserver does not
+  /// necessarily also host a copy of the `.well-known`.
+  ///
+  /// The result of this call is stored in [wellKnown] for later use at runtime.
+  @override
+  Future<DiscoveryInformation> getWellknown([String? hostname]) async {
+    final wellKnown = await getDiscoveryInformationsByUserId(
+      hostname ?? baseUri!.host,
```
nico-famedly, line 563: This could be a problem. Imagine you have a homeserver example.com delegating to matrix.example.com. On the first call this would request example.com/.well-known/matrix/client, on the second call it would request <b>matrix.</b>example.com/.well-known/matrix/client. Those may contain different values or the latter might not exist at all. I think we should always request the one from the userids servername unless we are pre-login.

lib/src/client.dart
```diff
@@ -121,6 +120,24 @@ class Client extends MatrixApi {
   /// The timeout until a typing indicator gets removed automatically.
   final Duration typingIndicatorTimeout;
 
+  DiscoveryInformation? _wellKnown;
```
nico-famedly, line 123: While I do think caching this is a good idea, where do we clear this? We probably should clear this on logout or in init even if the well-known request fails. And probably have a test for that.

nico-famedly (commented):
Sorry for the long review, but most stuff is just nitpicks/questions.

The bigger question is how do we want to cache it and do we want to allow SDK users to pass custom hostnames.

I think regarding the caching we have several options:
- Only store in the db <- bad UX imo, since you can't use it for simple config flags then
- Manual refresh of the cache
- Automatic refresh of the cache periodically (i.e. every 24h) with a manual "refreshNow"
- Automatic refresh on every "get" <- imo not that great, since it would have a hidden cost associated with the get, but might be okay when coupled with a 24h delay timer.

TheOneWithTheBraid:
Poke @nico-famedly !

nico-famedly (approved):

lib/src/client.dart
nico-famedly: I'm going to merge this as is for now, we can give this function a cache time at a later point. Users usually don't need to call this anyway, since they can access the well-known synchronously using the property.

PR #1877, opened 2024-07-12
feat: debounce ratcheting voip keys

td-famedly:
TODO: test


td-famedly:
We do not use ratcheting, maybe redo later

PR #1879, opened 2024-07-16
refactor: Better avatar and names for invites

krille-chan:
Group chats with no user
avatar should return null for
the room.avatar getter and not
a random hero avatar. This could
otherwise lead to confusion as
it looks like this is a DM which
is not the case.

For the name we should also
not just display the name of
the invitor like in a DM even for
group chats, but some more
additional information. I found
a String for
invites quite useful here as
this would name rooms without
a m.room.name like this:
"Invited by $senderName"
which should be short enough.
The alternative
"You have been invited by $senderName" could be too
long IMO.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -263,11 +263,11 @@ class Room {
     if (membership == Membership.invite) {
       final ownMember = unsafeGetUserFromMemoryOrFallback(client.userID!);
 
-      unsafeGetUserFromMemoryOrFallback(ownMember.senderId)
-          .calcDisplayname(i18n: i18n);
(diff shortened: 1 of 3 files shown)


lib/src/room.dart
```diff
@@ -291,25 +291,19 @@ class Room {
   /// that you have the room members, call and await `Room.loadHeroUsers()`
```
td-famedly, line 291: is this comment still valid?

krille-chan: Yes it is. The method also makes sure that the directChatMatrixId user is loaded

PR #1882, opened 2024-07-19
fix: cached profile returns null on app startup

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/2215

But I think this might also force `postLoad` on multiple rooms during client startup.


coder-with-a-bushido:
Closing in favour of https://github.com/famedly/app/pull/6009

PR #1883, opened 2024-07-22
refactor: Cache profiles in database and refactor API

krille-chan:


lib/src/client.dart
```diff
@@ -911,14 +911,15 @@ class Client extends MatrixApi {
     return id;
   }
 
-  @Deprecated('Use fetchOwnProfile() instead')
+  @Deprecated('Use getUserProfile() instead')
```
coder-with-a-bushido, line 914:
```suggestion
  @Deprecated('User `getUserProfile(userID)` instead')
```

lib/src/client.dart
```diff
@@ -955,13 +957,64 @@ class Client extends MatrixApi {
...
+      return cachedProfile;
+    }
+
+    late final Future<ProfileInformation> request;
+    try {
+      request = _userProfileRequests[userId] ??=
+          super.getUserProfile(userId).timeout(timeout);
+    } catch (e) {
+      Logs().d('Unable to fetch profile from server', e);
+      if (cachedProfile == null) rethrow;
```
coder-with-a-bushido, line 987: else, we could add a log that we're returning an outdated profile

lib/src/client.dart
```diff
@@ -955,13 +957,64 @@ class Client extends MatrixApi {
...
+      return cachedProfile;
+    }
+
+    final profile = await request;
```
coder-with-a-bushido, line 991: if I'm not wrong, the exception will occur here only and not in the `request` assignment.

lib/src/client.dart
krille-chan: Hmm, the consumer could fetch this infor mation easily when using `CachedProfileInformation.outdated == true`

lib/src/client.dart
krille-chan: You are right I think 🤔 thank you so much. Fixing it

lib/src/client.dart
coder-with-a-bushido: @krille-chan  we might also have to set guidelines for where to log and where not to, in the SDK. I'm convinced with your reply but I could still get confused on which places need a log and which do not most of the time

lib/src/client.dart
krille-chan: Fixed :)

lib/src/utils/cached_profile_information.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:matrix/matrix_api_lite.dart';
+
+class CachedProfileInformation extends ProfileInformation {
```
coder-with-a-bushido, line 3: can we also add `userID` to this so it's easy to replace with `Profile` during migration?

lib/src/database/matrix_sdk_database.dart
```diff
@@ -103,6 +103,8 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
 
   late Box<String> _seenDeviceKeysBox;
 
+  late Box<Map> _userProfilesBox;
```
coder-with-a-bushido, line 106: do we want to add this to `importDump()` and `exportDump()`?

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1618,4 +1627,31 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
+  }
+
+  @override
+  Future<CachedProfileInformation?> getUserProfile(String userId) =>
+      _userProfilesBox.get(userId).then((json) => json == null
+          ? null
+          : CachedProfileInformation.fromJson(copyMap(json)));
+
+  @override
+  Future<void> storeUserProfile(String userId, ProfileInformation profile) =>
```
coder-with-a-bushido, line 1652: shouldn't the `profile` here be `CachedProfileInformation` too?

lib/src/client.dart
```diff
@@ -955,13 +957,63 @@ class Client extends MatrixApi {
...
+  /// [userId] The user whose profile information to get.
+  @override
+  Future<CachedProfileInformation> getUserProfile(
+    String userId, {
+    Duration timeout = const Duration(seconds: 30),
+    Duration maxCacheAge = const Duration(days: 1),
+  }) async {
+    final cachedProfile = await database?.getUserProfile(userId);
+    if (cachedProfile != null &&
+        (cachedProfile.outdated != true ||
```
coder-with-a-bushido, line 976: just a nitpick
```suggestion
        (!cachedProfile.outdated ||
```

lib/src/utils/cached_profile_information.dart
coder-with-a-bushido, line 3: The name `CachedProfileInformation` can be misleading here

coder-with-a-bushido: I will reply here if I have a better name suggestion but do you think extending `ProfileInformation` for the existing `User` class is a good idea? Could be a mixin too

lib/src/client.dart
```diff
@@ -955,13 +957,63 @@ class Client extends MatrixApi {
...
+
   /// Get the combined profile information for this user.
   /// If [getFromRooms] is true then the profile will first be searched from the
   /// room memberships. This is unstable if the given user makes use of different displaynames
   /// and avatars per room, which is common for some bots and bridges.
   /// If [cache] is true then
   /// the profile get cached for this session. Please note that then the profile may
   /// become outdated if the user changes the displayname or avatar in this session.
+  @Deprecated('User `getUserProfile(userID)` instead')
   Future<Profile> getProfileFromUserId(String userId,
```
coder-with-a-bushido, line 1017: can we also change the usage of `getProfileFromUserId` in `Room.requestUser()` to `getUserProfile`? Please check if my suggestion below for `User` class makes sense too

lib/src/client.dart
```diff
@@ -955,13 +957,63 @@ class Client extends MatrixApi {
...
+    Duration maxCacheAge = const Duration(days: 1),
+  }) async {
+    final cachedProfile = await database?.getUserProfile(userId);
+    if (cachedProfile != null &&
+        (cachedProfile.outdated != true ||
+            DateTime.now().difference(cachedProfile.updated) >= maxCacheAge)) {
+      return cachedProfile;
+    }
+
+    late final ProfileInformation profile;
```
td-famedly, line 981: does this need to be late?

lib/src/client.dart
```diff
@@ -2295,6 +2347,16 @@ class Client extends MatrixApi {
               content: update.content)));
         }
       }
+
+      // Any kind of member change? We should invalidate the profile then:
```
td-famedly, line 2351: Hm, could we just refresh the cache instead of marking it as outdated? i.e can you list the reasons for not doing so?

lib/src/database/hive_database.dart
```diff
@@ -1403,6 +1403,22 @@ class FamedlySdkHiveDatabase extends DatabaseApi with ZoneTransactionMixin {
 
   @override
   Future<void> delete() => Hive.deleteFromDisk();
+
+  @override
+  Future<void> markUserProfileAsOutdated(userId) async {
+    return;
```
td-famedly, line 1409: probably should be throwing Unimplemented errors here.

lib/src/utils/cached_profile_information.dart
krille-chan: I would say no, because that would be duplicated information. When you request the ProfileInformation by the userId you obviously already have the userId. I would rather prefer to stick as close to the API as possible than implementing convenient methods which could lead to inconsistent data (like profile.userId != theUserIdYouGaveIntoTheFunction)

lib/src/database/matrix_sdk_database.dart
krille-chan: I have not added those as the profiles are just cache, which can easily be loaded again from the server. Those blow up the database dump without a real benefit IMO

lib/src/database/matrix_sdk_database.dart
krille-chan: Ah yes you are right. Thanks. Have changed it

lib/src/client.dart
krille-chan: Good idea. I changed it

lib/src/utils/cached_profile_information.dart
krille-chan: No, the User class is an extension to the StrippedStateEvent and this is definitely not what we have. I chose `CachedProfileInformation` because we already introduced `CachedPresence` for a `Presence` which we have cached

lib/src/client.dart
krille-chan: Actually no. I removed the `late` :D thanks

lib/src/client.dart
krille-chan: I've listed the reason

lib/src/database/hive_database.dart
krille-chan: No, then the database would break on every sync. Just returning nothing leads to having no cache functionality which still makes the database able to migrate to something different

lib/src/client.dart
coder-with-a-bushido: I'm talking about this line -

https://github.com/famedly/matrix-dart-sdk/blob/191397ebb8ba7b93b288365c8cedc5941289140d/lib/src/room.dart#L1720

krille-chan: But this is already changed 🤔

coder-with-a-bushido (approved):
LGTM :)

PR #1885, opened 2024-07-23
Feature/get space hierarchy cache

WilsonLe:


nico-famedly:
This includes changes from authors, that haven't signed the CLA. Could you ask them to sign the CLA or remove their contribution?

lib/src/client.dart
```diff
@@ -2315,6 +2316,10 @@ class Client extends MatrixApi {
       }
       onEvent.add(update);
 
+      if (event.type == EventTypes.SpaceChild) {
```
nico-famedly, line 2319: You also need to trigger this, when a parent is changed, since you don't need both relations.

lib/src/client.dart
```diff
@@ -3323,6 +3328,55 @@ class Client extends MatrixApi {
       waitUntilLoadCompletedLoaded: false,
     );
   }
+
+  @override
+  Future<GetSpaceHierarchyResponse> getSpaceHierarchy(String roomId,
```
nico-famedly, line 3333: You seem to just be caching the whole response. This can introduce some weirdness if you mix cached and uncached responses. Additionally such responses shouldn't be cached indefinitely, I would say.

You also aren't indexing your cache entries by the other parameters for the request, which means for spaces with more than "limit" entries in the response, you will be returning the wrong response. You will also return the wrong response if any of the parameters (apart from the roomid) changes. Additionally you aren't invalidating the cache if a subspace member changes, which means for any depth > 1 this will yield wrong results, if the subspace changes.

I think this needs a lot more tests and quite a bit of work on the cache invalidation logic. While it would certainly be useful to have more efficient access to the space hierarchy, we don't want to return wrong results.

td-famedly:
I will close this for now

PR #1886, opened 2024-07-24
fix: also delete db on logout

td-famedly:
to be paired with https://github.com/famedly/app/pull/5890

Tested ^ on android, ios and web, also tested migration from main to that branch on android seems to work.

### DIFF
--- a/lib/src/database/indexeddb_box.dart
+++ b/lib/src/database/indexeddb_box.dart
@@ -83,8 +83,14 @@ class BoxCollection with ZoneTransactionMixin {
     return _db.close();
   }
 
-  static Future<void> delete(String path, [dynamic factory]) =>
-      (factory ?? window.indexedDB!).deleteDatabase(path);
+  @Deprecated('use collection.deleteDatabase now')
+  static Future<void> delete(String name, [dynamic factory]) =>
+      (factory ?? window.indexedDB!).deleteDatabase(name);
+
+  Future<void> deleteDatabase(String name, [dynamic factory]) async {
+    await close();
+    await (factory ?? window.indexedDB).deleteDatabase(name);
+  }
 }
 
 class Box<V> {
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -41,5 +41,6 @@ dev_dependencies:
   file: ">=6.1.1 <8.0.0"
   import_sorter: ^4.6.0
   lints: ^4.0.0
+  path: ^1.9.0
   sqflite_common_ffi: 2.3.2+1 # v2.3.3 doesn't support dart v3.2.x
   test: ^1.15.7
--- a/test/box_test.dart
+++ b/test/box_test.dart
@@ -100,9 +100,9 @@ void main() {
     });
 
     test(
-      'Box.delete',
+      'Collection.deleteDatabase',
       () async {
-        await BoxCollection.delete(
+        await collection.deleteDatabase(
           db?.path ?? '',
           isWeb ? null : databaseFactoryFfi,
         );
--- a/test/client_test.dart
+++ b/test/client_test.dart
@@ -18,11 +18,13 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 import 'dart:typed_data';
 
 import 'package:canonical_json/canonical_json.dart';
 import 'package:collection/collection.dart';
 import 'package:olm/olm.dart' as olm;
+import 'package:path/path.dart' show join;
 import 'package:test/test.dart';
 
 import 'package:matrix/matrix.dart';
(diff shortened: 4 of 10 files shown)


lib/src/database/indexeddb_box.dart
```diff
@@ -70,21 +70,23 @@ class BoxCollection with ZoneTransactionMixin {
...
 
-  Future<void> close() async {
+  Future<void> closeDb() async {
     assert(_txnCache == null, 'Database closed while in transaction!');
     return _db.close();
   }
 
-  static Future<void> delete(String path, [dynamic factory]) =>
-      (factory ?? window.indexedDB!).deleteDatabase(path);
+  Future<void> deleteDb(String name, [dynamic factory]) async {
```
nico-famedly, line 86: Is there any context for why those got renamed?

test/box_test.dart
```diff
@@ -96,13 +96,13 @@ void main() {
...
     test('Box.close', () async {
-      await collection.close();
+      await collection.closeDb();
     });
 
     test(
       'Box.delete',
       () async {
-        await BoxCollection.delete(
+        await collection.deleteDb(
```
nico-famedly, line 105: NIT: test name doesn't match the function name anymore.

lib/src/database/indexeddb_box.dart
td-famedly: cleaned up

test/box_test.dart
td-famedly: cleaned up

lib/src/database/indexeddb_box.dart
```diff
@@ -83,8 +83,14 @@ class BoxCollection with ZoneTransactionMixin {
...
   }
 
+  @Deprecated('use collection.deleteDatabase now')
   static Future<void> delete(String path, [dynamic factory]) =>
       (factory ?? window.indexedDB!).deleteDatabase(path);
+
+  Future<void> deleteDatabase(String name, [dynamic factory]) async {
+    await close();
+    await (factory ?? window.indexedDB).deleteDatabase(name);
+  }
```
nico-famedly, line 93: what's the difference between path and name?

td-famedly: none, indexedDb just doesn't have a path so I wanted to avoid any confusion here https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/deleteDatabase

test/client_test.dart
```diff
@@ -31,16 +33,35 @@ import 'fake_client.dart';
...
   // key @test:fakeServer.notExisting
   const pickledOlmAccount =
       'N2v1MkIFGcl0mQpo2OCwSopxPQJ0wnl7oe7PKiT4141AijfdTIhRu+ceXzXKy3Kr00nLqXtRv7kid6hU4a+V0rfJWLL0Y51+3Rp/ORDVnQy+SSeo6Fn4FHcXrxifJEJ0djla5u98fBcJ8BSkhIDmtXRPi5/oJAvpiYn+8zMjFHobOeZUAxYR0VfQ9JzSYBsSovoQ7uFkNks1M4EDUvHtuyg3RxViwdNxs3718fyAqQ/VSwbXsY0Nl+qQbF+nlVGHenGqk5SuNl1P6e1PzZxcR0IfXA94Xij1Ob5gDv5YH4UCn9wRMG0abZsQP0YzpDM0FLaHSCyo9i5JD/vMlhH+nZWrgAzPPCTNGYewNV8/h3c+VyJh8ZTx/fVi6Yq46Fv+27Ga2ETRZ3Qn+Oyx6dLBjnBZ9iUvIhqpe2XqaGA1PopOz8iDnaZitw';
   const identityKey = '7rvl3jORJkBiK4XX1e5TnGnqz068XfYJ0W++Ml63rgk';
   const fingerprintKey = 'gjL//fyaFHADt9KBADGag8g7F8Up78B/K1zXeiEPLJo';
 
+  group('client path', () {
+    late Client clientOnPath;
+
+    final dbPath = join(dirname(Platform.script.path), 'test.sqlite');
```
nico-famedly, line 45: why not use the current working directory?

pubspec.yaml
```diff
@@ -26,6 +26,7 @@ dependencies:
   markdown: ^7.1.1
   mime: ^1.0.0
   olm: ^2.0.2
+  path: ^1.9.0
```
nico-famedly, line 29: I don't think it is worth adding this dependency just for some tests.

test/client_test.dart
td-famedly: Directory.current does look easier thanks.

pubspec.yaml
td-famedly: changed to dev dep

lib/src/database/indexeddb_box.dart
nico-famedly: Then why not rename the parameter for delete?

test/fake_client.dart
```diff
@@ -26,12 +26,15 @@ const ssssKey = 'EsT9 RzbW VhPW yqNp cC7j ViiW 5TZB LuY4 ryyv 9guN Ysmr WDPH';
 const pickledOlmAccount =
     'N2v1MkIFGcl0mQpo2OCwSopxPQJ0wnl7oe7PKiT4141AijfdTIhRu+ceXzXKy3Kr00nLqXtRv7kid6hU4a+V0rfJWLL0Y51+3Rp/ORDVnQy+SSeo6Fn4FHcXrxifJEJ0djla5u98fBcJ8BSkhIDmtXRPi5/oJAvpiYn+8zMjFHobOeZUAxYR0VfQ9JzSYBsSovoQ7uFkNks1M4EDUvHtuyg3RxViwdNxs3718fyAqQ/VSwbXsY0Nl+qQbF+nlVGHenGqk5SuNl1P6e1PzZxcR0IfXA94Xij1Ob5gDv5YH4UCn9wRMG0abZsQP0YzpDM0FLaHSCyo9i5JD/vMlhH+nZWrgAzPPCTNGYewNV8/h3c+VyJh8ZTx/fVi6Yq46Fv+27Ga2ETRZ3Qn+Oyx6dLBjnBZ9iUvIhqpe2XqaGA1PopOz8iDnaZitw';
 
-Future<Client> getClient(
-    {Duration sendTimelineEventTimeout = const Duration(minutes: 1)}) async {
+Future<Client> getClient({
+  Duration sendTimelineEventTimeout = const Duration(minutes: 1),
+  String? databasePath,
```
nico-famedly, line 31: This could use a comment, that you should not pass this unless you have a special need for it, since by default tests should use the in memory db, I think?

lib/src/database/indexeddb_box.dart
td-famedly: done, altho that's deprecated already!

test/fake_client.dart
td-famedly: done

nico-famedly (approved):
I would like to see the app tests pass as well, but otherwise this looks good.

td-famedly:
https://github.com/famedly/app/actions/runs/10114831185?pr=5890

PR #1887, opened 2024-07-25
ci: create release job fix

coder-with-a-bushido:


.github/workflows/publish.yml
```diff
@@ -28,16 +28,16 @@ jobs:
           releaseRegex="^v[0-9]+\.[0-9]+\.[0-9]+$"
           releaseCandidateRegex="^v[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$"
           notes=$(./scripts/extract_changelog.sh $version)
-          if [ -z $notes ]; then
+          if [[ -n "${notes/[ ]*\n/}" ]]; then
```
nico-famedly, line 31: I think this could use some explanation of what you are matching for. It seems to me that you are replacing all the spaces and a trailing newline in ABC and then checking if the string is not empty. But I don't understand why you would be doing that at a glance. I did learn that the extract version script currently prints one empty line if there is no changelog, but in that case just adding the quotes around $notes should be enough.

```suggestion
          if [ -z "$notes" ]; then
```

But even without that it seems to work fine for me, so yeah, I don't understand that change.

Also your brackets seem to be redundant, ` *\n` is the exact same regex as `[ ]*\n`, the brackets are only needed if you have more than one Element.

nico-famedly: Maybe you ran into https://stackoverflow.com/questions/16355213/string-check-in-bash-with-z-and-n, which can be fixed by either using [[ or by quoting the variable?

nico-famedly (approved):

PR #1891, opened 2024-07-28
build: Release v0.31.0

coder-with-a-bushido:


nico-famedly (approved):
Seems fine to me

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## [0.31.0] 28th July 2024
+
```
nico-famedly, line 2: What's the braking change?

coder-with-a-bushido: > <@deepbluev7:neko.dev> that should probably be mentioned, because it is not clear from the changelog, is it?

Sorry for using the term BREAKING in chat, because we're only deprecating methods and not removing anything. But I want the version bump to be major because we're changing the underlying logic of deprecated ones(for profile fetching) to use the new one (which in case the user is using an old DB could mean that cache doesn't work no more for profiles)

td-famedly: new localization in https://github.com/famedly/matrix-dart-sdk/commit/2e51b450bf8002cdbe6a1b3a4b3dbafd6641f1be so breaking change makes sense

td-famedly (approved):
lg

PR #1892, opened 2024-07-29
feature: room timeline reference

WilsonLe:
Improve run time of `getTimeline` by introducing a mutable reference to it inside a room instance. Only create new timeline instance if the chunks are different, otherwise reuse existing one.


WilsonLe:
@nico-famedly Sorry for late response, I have removed all the space hierarchy changes in this PR.

td-famedly:
I will close this for now

PR #1893, opened 2024-07-29
refactor: Do not set default timeout for initialSync

krille-chan:


nico-famedly (approved):
I think the impact of the timeout passed to /sync is not that big, because if the server has a response, it will start returning something and the timeout only applies when there is nothing to return. However increasing the timeout for the response might actually make sense. I am not sure if it should actually be unlimited, since it might hang on network errors if the network connection changes. However, I guess we can just test that and see, how it behaves?

krille-chan:
> I think the impact of the timeout passed to /sync is not that big, because if the server has a response, it will start returning something and the timeout only applies when there is nothing to return. However increasing the timeout for the response might actually make sense. I am not sure if it should actually be unlimited, since it might hang on network errors if the network connection changes. However, I guess we can just test that and see, how it behaves?

There are still users who say that they cannot login because the server never response fast enough before FluffyChat retries it

nico-famedly:
Oh, I guess that could be happening, if the sync response just takes longer than 2 minutes to send? But I guess this change is fine then, thanks for the answer!

PR #1895, opened 2024-07-31
fix: Call logout on soft logout fail

krille-chan:
If we fail to handle the
soft logout, we should call
the logout endpoint to
delete the device from
the server instead of just
clear the client.

Closes https://github.com/famedly/product-management/issues/2292


nico-famedly (approved):
I'm a bit concerned if logging the user out on any exception might be overly eager (i.e. you could fail the relogin if there is currently no network available), but in general this seems to be fine.

PR #1896, opened 2024-07-31
fix: leave->invite in the same sync would hide the invite

nico-famedly:
Synapse includes the room in both sections if you have both an invite->leave and a leave->invite transition in one sync response. Transitions in the other order are only included once (in the leave section) it seems, so this should work correctly in all cases.

Fixes https://github.com/famedly/product-management/issues/2283


mErrenst (approved):

mErrenst:
LGTM

PR #1897, opened 2024-08-01
fix: maintain event relation data on redact

td-famedly:
alternative fix where we just ignore empty content unredacted messages https://github.com/famedly/app/pull/6100

proper solution would be saving the lastEvents with a different id (or atleast a different cache key so that they don't pollute the timeline events, this will need more investigation)

ideally only the original event should be redacted when we call setRedact, but incase an edited event is the lastEvent, setRedact is also called on the edited version of the message, this causes the db memory cache to have a wrong version of the message for the timeline, theis then causes the m.room.message as we have no content, another solution would be to just ignore empty content bodies

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -251,6 +251,8 @@ class Event extends MatrixEvent {
     _originalSource = null;
     final contentKeyWhiteList = <String>[];
     switch (type) {
+      case EventTypes.Message:
+        contentKeyWhiteList.add('m.relates_to');
       case EventTypes.RoomMember:
         contentKeyWhiteList.add('membership');
         break;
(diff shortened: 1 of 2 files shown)


nico-famedly:
The server doesn't keep relation data on redaction, so wouldn't this still break for redacted events sent by the server to us?

td-famedly:
> The server doesn't keep relation data on redaction, so wouldn't this still break for redacted events sent by the server to us?

Well the spec does not mention if we are supposed to redact aggregated events when deleting a event with edits, but we could just instead use the alt fix which ignores any unredacted m.room.message where the content body is empty

The proper fix for this like stated in the issue body would be storing the lastEvent with a different cache key but that needs more investigation. Do we want to go that route?

td-famedly:
we decided to go with the alt fix of ignoring empty content undredacted messages

PR #1900, opened 2024-08-09
build: Release v0.32.0

krille-chan:


coder-with-a-bushido (approved):

td-famedly (changes requested):
shouldn't this be a breaking release because of the new localization in the cancelSend PR?

td-famedly (approved):
lg

PR #1901, opened 2024-08-15
feat: support for authenticated media

td-famedly:
The ideal way to fix this would be autogen with the v1.11 spec. But because it changed from openapi v2 to v3 that looks like a bit more work than expected. Separating out auth media stuff from that because of urgency.

To test these changes please make sure your server properly registers the new media endpoints. Famedly servers do not atm because of worker misconfiguration. Tracking issue: https://github.com/famedly/infra-meta/issues/1688

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -25,12 +25,14 @@ import 'dart:typed_data';
 import 'package:async/async.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:http/http.dart' as http;
+import 'package:http/http.dart';
 import 'package:mime/mime.dart';
 import 'package:olm/olm.dart' as olm;
 import 'package:random_string/random_string.dart';
 
 import 'package:matrix/encryption.dart';
 import 'package:matrix/matrix.dart';
+import 'package:matrix/matrix_api_lite/generated/fixed_model.dart';
 import 'package:matrix/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart';
 import 'package:matrix/src/models/timeline_chunk.dart';
 import 'package:matrix/src/utils/cached_stream_controller.dart';
@@ -41,6 +43,7 @@ import 'package:matrix/src/utils/run_benchmarked.dart';
 import 'package:matrix/src/utils/run_in_root.dart';
 import 'package:matrix/src/utils/sync_update_item_count.dart';
 import 'package:matrix/src/utils/try_get_push_rule.dart';
+import 'package:matrix/src/utils/versions_comparator.dart';
 
 typedef RoomSorter = int Function(Room a, Room b);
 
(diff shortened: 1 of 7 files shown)


lib/src/utils/uri_extension.dart
```diff
@@ -21,7 +21,79 @@ import 'dart:core';
...
+  /// Transforms this `mxc://` Uri into a `http` resource, which can be used
+  /// to download the content.
+  ///
+  /// Throws an exception if the scheme is not `mxc` or the homeserver is not
+  /// set.
+  ///
+  /// Important! To use this link you have to set a http header like this:
+  /// `headers: {"authorization": "Bearer ${client.accessToken}"}`
+  Future<Uri> getDownloadUri(Client client) async {
+    if (!isScheme('mxc')) throw Exception('Uri has unknown scheme "mxc"');
```
td-famedly, line 33: not sure if we should throw a exception or return a empty uri here atm

td-famedly: decided not to

lib/src/client.dart
```diff
@@ -1158,12 +1160,213 @@ class Client extends MatrixApi {
     _archivedRooms.add(ArchivedRoom(room: archivedRoom, timeline: timeline));
   }
 
+  GetVersionsResponse? _versionsCache;
+
+  Future<bool> authenticatedMediaSupported() async {
+    _versionsCache ??= await getVersions();
+    return _versionsCache?.versions.contains('v1.11') ?? false;
```
krille-chan, line 1167: Actually it is supported if a version of "v1.11" or higher is supported, right? So what if a server claims to support v1.12 but not v1.11, could this happen?

krille-chan (approved):
Making the getDownloadLink and getThumbnail methods async seems to be okay. I checked it in FluffyChat and there it always looks like it is not that big of a deal to use it this way. So we can be sure we always have the correct link. Thank you very much

lib/src/client.dart
td-famedly: Hm I do not see the spec mentioning anything against this so it is possible yes. Nevertheless, In practice I don't think it's something we need to worry about. Fixing this ofc would mean some simple regex and checks, do we want to go that route?

lib/src/client.dart
```diff
@@ -1158,12 +1161,222 @@ class Client extends MatrixApi {
...
 
+  GetVersionsResponse? _versionsCache;
+
+  Future<bool> authenticatedMediaSupported() async {
+    _versionsCache ??= await getVersions();
+    return _versionsCache?.versions.any(
+          (v) => isVersionGreaterThanOrEqualTo(v, 'v1.11'),
+        ) ??
+        false;
+  }
```
nico-famedly, line 1172: I know I am late, but imo it makes more sense to just try the authenticated media endpoints and see, if they fail. There are servers that claim 1.11 support for example, where the proxies are misconfigured and the endpoints just fail. We shouldn't fail to download media in those cases, but fall back to trying the old endpoints imo.

td-famedly: moved to https://github.com/famedly/matrix-dart-sdk/pull/1903

PR #1903, opened 2024-08-16
feat: alt way for auth media (BREAKING)

td-famedly:
This removes the version checks, and expects clients to manually return both endpoints using the `useAuthentication` flag

Breaking change but tries both new and old endpoints. Would require clients to implement a retry mech for their media if they fail.

Particularly useful when servers are misconfigured so they return on the old endpoints but not on the new ones. (not a fan, just configure them correctly!)


lib/src/utils/uri_extension.dart
```diff
@@ -59,19 +62,22 @@ extension MxcUriExtension on Uri {
   ///
   /// Important! To use this link you have to set a http header like this:
   /// `headers: {"authorization": "Bearer ${client.accessToken}"}`
-  Future<Uri> getThumbnailUri(Client client,
-      {num? width,
-      num? height,
-      ThumbnailMethod? method = ThumbnailMethod.crop,
-      bool? animated = false}) async {
+  Uri getThumbnailUri(
```
nico-famedly, line 65: I think we should deprecate the functions to get a raw URL as an SDK user. They might be useful internally, but SDK users will need to manually set auth headers when using any of these URIs. This means you basically want to use the thumbnail and download functions from the SDK anyway, never really the *Uri functions, I think.

nico-famedly: Okay, so some clients are using this to feed Image.network and similar, so those would benefit from keeping these around. However I think we can at least deprecate the extensions on Uri and only expose 2 functions returning a Uri instead if 4+.

nico-famedly (approved):

lib/src/client.dart
```diff
@@ -1189,21 +1178,25 @@ class Client extends MatrixApi {
...
+        unexpectedResponse(response, responseBody);
+      }
+      final responseString = utf8.decode(responseBody);
+      final json = jsonDecode(responseString);
+      return ServerConfig.fromJson(json as Map<String, Object?>);
+    }
+
+    try {
+      return request('_matrix/client/v1/media/config');
+    } catch (e) {
```
krille-chan, line 1197: Do we know the exact error message that would be returned here? Then we should listen to only this and rethrow other errors

lib/src/client.dart
```diff
@@ -1222,27 +1215,30 @@ class Client extends MatrixApi {
...
+        unexpectedResponse(response, responseBody);
+      }
+      return FileResponse(
+          contentType: response.headers['content-type'], data: responseBody);
+    }
+
+    try {
+      return request(
+          '_matrix/client/v1/media/download/${Uri.encodeComponent(serverName)}/${Uri.encodeComponent(mediaId)}');
+    } catch (e) {
```
krille-chan, line 1238: Same here

lib/src/client.dart
```diff
@@ -1266,26 +1262,30 @@ class Client extends MatrixApi {
...
+        unexpectedResponse(response, responseBody);
+      }
+      return FileResponse(
+          contentType: response.headers['content-type'], data: responseBody);
+    }
+
+    try {
+      return request(
+          '_matrix/client/v1/media/download/${Uri.encodeComponent(serverName)}/${Uri.encodeComponent(mediaId)}/${Uri.encodeComponent(fileName)}');
+    } catch (e) {
```
krille-chan, line 1285: And here

krille-chan (changes requested):
Don't get it yet... why should we implement such a hacky workaround because of a misconfigured server? Shouldn't we fix the server instead?

PR #1904, opened 2024-08-18
fix: SQLCipher dylib location on macOS builds

TheOneWithTheBraid:
I guess no one ever noticed, but obviously the path of the SQLCipher build on macOS is relative to the build app on macOS. I double checked in both release and debug mode, this is the correct path for the SQLCipher dylib on macOS. The `Current` in the path defines it automatically chooses the matching one whether on aarch64 or amd64.

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -60,10 +60,6 @@ jobs:
           architecture: "arm64"
       - name: Run tests
         run: |
-          # Prevent restarting the github actions service on package upgrades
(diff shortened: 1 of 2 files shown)


nico-famedly (approved):
Looks reasonable to me, thanks!

TheOneWithTheBraid:
@nico-famedly CI fails but on a completely unrelated test, could you have a look at it ?

nico-famedly:
@TheOneWithTheBraid, CI failures are unrelated:
- the coverage report does not work for external PRs apparently
- The dendrite job is marked as allowed to fail

PR #1907, opened 2024-08-21
fix: dedup /versions calls

td-famedly:


lib/src/client.dart
```diff
@@ -1161,10 +1161,11 @@ class Client extends MatrixApi {
     _archivedRooms.add(ArchivedRoom(room: archivedRoom, timeline: timeline));
   }
 
-  GetVersionsResponse? _versionsCache;
+  final _versionsCache =
+      AsyncCache<GetVersionsResponse>(const Duration(hours: 1));
```
nico-famedly, line 1165: this needs to be reset on logout, I don't really see that happening anywhere?

td-famedly: good catch done

PR #1908, opened 2024-08-26
fix: race condition between getState and requestUser updates

td-famedly:
Fixes: https://github.com/famedly/product-management/issues/2319
 
tldr  notification > requestUser > requestState fails  > requestProfile > updates RoomMember stateKey with bad data (data with no longer is_direct or the membership)

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1694,10 +1694,9 @@ class Room {
     required bool requestProfile,
   }) async {
     // Is user already in cache?
-    final userFromState = getState(EventTypes.RoomMember, mxID)?.asUser(this);
 
     // If not in cache, try the database
-    var foundUser = userFromState;
+    User? foundUser = getState(EventTypes.RoomMember, mxID)?.asUser(this);
 
     // If the room is not postloaded, check the database
     if (partial && foundUser == null) {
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -1739,11 +1741,13 @@ class Room {
     }
 
     if (foundUser == null) return null;
+    // make sure we didn't actually store anything by the time we did those requests
+    final userFromStateFR = getState(EventTypes.RoomMember, mxID)?.asUser(this);
```
rmoradi88, line 1745: Can you describe what ```userFromStateFR``` stands for?

rmoradi88 (commented):
Tested and works as expected. Also I think this will fix https://github.com/famedly/product-management/issues/2318 issue as well.

lib/src/room.dart
```diff
@@ -1739,11 +1741,13 @@ class Room {
...
 
     // Set user in the local state if the state changed.
     // If we set the state unconditionally, we might end up with a client calling this over and over thinking the user changed.
-    if (userFromState == null ||
-        userFromState.displayName != foundUser.displayName) {
+    if (userFromStateFR == null ||
+        userFromStateFR.displayName != foundUser.displayName) {
```
nico-famedly, line 1750: Make this the userFromCurrentState and change the one above to "foundUser"

nico-famedly (approved):

PR #1910, opened 2024-08-28
fix: actually make sure clientBox has a account

td-famedly:


lib/src/client.dart
```diff
@@ -1880,7 +1880,11 @@ class Client extends MatrixApi {
 
       final account = await this.database?.getClient(clientName);
       newRefreshToken ??= account?.tryGet<String>('refresh_token');
-      if (account != null) {
+      // can have discovery_information so make sure it also has the proper
+      // account creds
+      if (account != null &&
+          account['homeserver_url'] != null &&
+          account['token'] != null) {
```
nico-famedly, line 1887: Maybe also check the client id?

nico-famedly (approved):

lib/src/client.dart
td-famedly: also added user_id

PR #1912, opened 2024-09-03
fix: Delete empty dm room on createRoom error

mErrenst:
Closes https://github.com/famedly/product-management/issues/2385


lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+        }
       }
-    }
 
-    await Room(id: roomId, client: this).addToDirectChat(mxid);
+      await Room(id: roomId, client: this).addToDirectChat(mxid);
 
-    return roomId;
+      return roomId;
+    } catch (e) {
```
nico-famedly, line 819: This is catching any error, not just the room creation error. You should narrowly scope the try catch and handle specific errors, not all of them and assume it is the one error you intended to handle.

lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+      /// This looks up the room id of the room that was just created.
+      /// The room is empty since createRoom threw an error. To make sure not to
+      /// delete other empty rooms by accident, the creation time is checked.
+      /// Should be fixed in Synapse and then be removed
```
nico-famedly, line 823: You should really document, why this throws an error, because the synapse code tries to handle all errors ahead of time, so it is not clear, why you are getting an error on room creation. If the problem is invites, we can for example move those out of the room creation.

lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+      final sortedRooms = rooms.sortedBy((room) => room.timeCreated).reversed;
+      final emptyRoom = sortedRooms.firstWhereOrNull((room) =>
+          room.name.isEmpty &&
+          !room.isDirectChat &&
+          room.summary.mJoinedMemberCount == 1 &&
+          room.summary.mInvitedMemberCount == 0 &&
+          room.timeCreated
+              .isAfter(DateTime.now().subtract(Duration(seconds: 20))));
+      final roomId = emptyRoom?.id;
```
nico-famedly, line 832: it's probably more robust to handle the room id list differences from before and after the room creation instead of going purely by timestamp?

lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+
+      /// If we can find an empty unnamed room that was recently created, we leave and delete it
+      if (roomId != null) {
+        await leaveRoom(roomId);
+        await forgetRoom(roomId);
```
nico-famedly, line 837: This can throw and needs error handling to not cause weird error messages.

lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+      await Room(id: roomId, client: this).addToDirectChat(mxid);
 
-    return roomId;
+      return roomId;
+    } catch (e) {
+      /// This looks up the room id of the room that was just created.
+      /// The room is empty since createRoom threw an error. To make sure not to
+      /// delete other empty rooms by accident, the creation time is checked.
+      /// Should be fixed in Synapse and then be removed
+      final sortedRooms = rooms.sortedBy((room) => room.timeCreated).reversed;
```
nico-famedly, line 824: How do you know, that the room is already synced back to the client?

lib/src/client.dart
```diff
@@ -795,27 +795,49 @@ class Client extends MatrixApi {
...
+      final roomId = emptyRoom?.id;
+
+      /// If we can find an empty unnamed room that was recently created, we leave and delete it
+      if (roomId != null) {
+        await leaveRoom(roomId);
+        await forgetRoom(roomId);
+      }
+      rethrow;
+    }
   }
```
nico-famedly, line 841: This code needs tests imo.

nico-famedly:
I think I agree with td now, that if this only happens for invites rejected by the invite checker, that we should fix this in Synapse.

PR #1914, opened 2024-09-05
feat: BREAKING CHANGE v1.11 support

td-famedly:


test/client_test.dart
```diff
@@ -80,7 +80,7 @@ void main() {
       await client.init();
       expect(client.isLogged(), false);
       await client.login(
-        LoginType.mLoginPassword,
```
nico-famedly, line 83: We could provide our own enum, that has those strings defined to not break usercode and not have to retype that everytime

td-famedly: ye but do we want to? @mrmcoding88 mentioned extending those to support new login methods was a pain

lib/src/room.dart
```diff
@@ -469,8 +470,8 @@ class Room {
   static Tag _tryTagFromJson(Object o) {
     if (o is Map<String, dynamic>) {
       return Tag(
-          order: o.tryGet<num>('order', TryGet.silent)?.toDouble(),
-          additionalProperties: Map.from(o)..remove('order'));
+        order: o.tryGet<num>('order', TryGet.silent)?.toDouble(),
```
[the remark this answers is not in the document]
td-famedly: the spec doesn't mention them tho? https://spec.matrix.org/v1.11/client-server-api/#put_matrixclientv3useruseridroomsroomidtagstag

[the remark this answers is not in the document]
nico-famedly: It kinda does, because clients are supposed to use them to store extra data for rooms: https://spec.matrix.org/v1.11/client-server-api/#events-13

test/client_test.dart
nico-famedly: I think we should have our own defines for the officially known login methods still, yes. The functions should take a string though.

lib/matrix_api_lite/generated/api.dart
```diff
@@ -4591,13 +5056,18 @@ class Api {
...
+  /// get all the features available for the logged-in user. If no
+  /// authentication is provided, the server should only return the features
+  /// available to all users. Servers may wish to keep advertising features
+  /// here after they've been released into the spec to give clients a chance
+  /// to upgrade appropriately. Additionally, clients should avoid using
+  /// unstable features in their stable releases.
   Future<GetVersionsResponse> getVersions() async {
     final requestUri = Uri(path: '_matrix/client/versions');
     final request = Request('GET', baseUri!.resolveUri(requestUri));
+    request.headers['authorization'] = 'Bearer ${bearerToken!}';
```
nico-famedly, line 5070: This looks wrong

lib/src/room.dart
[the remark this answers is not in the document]
nico-famedly: Okay, I tracked this down:
- https://github.com/matrix-org/matrix-spec/pull/1793 removed the additional properties flag
- We still default to additionalProperties: false, but openapi changed the default in version 3: https://github.com/OAI/OpenAPI-Specification/pull/1548/files

So I guess we now get to decide, if we want to default to additionalProperties: true everywhere or not? I think it is useful for tags at least.

lib/matrix_api_lite/generated/api.dart
nico-famedly: Yes, but that should be optional. We call this before login!

td-famedly: oh right!

lib/src/room.dart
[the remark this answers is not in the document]
td-famedly: added it back

test/client_test.dart
td-famedly: ehh that really does not sound worth the effort :/

td-famedly: if you want we could do it on the sdk side manually

td-famedly: done

lib/matrix_api_lite/generated/api.dart
```diff
@@ -1834,7 +2296,7 @@ class Api {
         path: '_matrix/client/v3/knock/${Uri.encodeComponent(roomIdOrAlias)}',
         queryParameters: {
           if (serverName != null)
-            'server_name': serverName.map((v) => v).toList(),
+            for (final query in serverName) 'server_name': query
```
nico-famedly, line 2299: This seems to be exploded wrong

lib/matrix_api_lite/generated/api.dart
```diff
@@ -4616,6 +5124,7 @@ class Api {
   /// between the client and the server may affect the apparent behaviour of content
   /// repository APIs, for example, proxies may enforce a lower upload size limit
   /// than is advertised by the server on this endpoint.
+  @Deprecated('message')
```
nico-famedly, line 5127: This is still ugly, can you maybe at least manually edit the message?

nico-famedly (approved):

lib/src/client.dart
```diff
@@ -1223,206 +1222,15 @@ class Client extends MatrixApi {
...
-  /// [mediaId] The media ID from the `mxc://` URI (the path component)
-  ///
-  ///
-  /// [allowRemote] Indicates to the server that it should not attempt to fetch the media if it is deemed
-  /// remote. This is to prevent routing loops where the server contacts itself. Defaults to
-  /// true if not provided.
-  ///
-  @override
-  // TODO: remove once we are able to autogen this
-  Future<FileResponse> getContent(String serverName, String mediaId,
```
nico-famedly, line 1262: Doesn't this break TI Messenger compatibility as well as compatibility with older servers?

td-famedly: hmm, it's a breaking change and we should really remove these asap. I think we already decided to not autogen these?

td-famedly: hm breaking change, but the older servers part makes sense brb

nico-famedly: It is fine to break the API, it is not really okay to just have no media support now on TIM and other older servers. That would affect our users as well.

lib/src/client.dart
```diff
@@ -1253,124 +1236,125 @@ class Client extends MatrixApi {
...
+  /// uploaded. The default value is 20000 (20 seconds). The content
+  /// repository SHOULD impose a maximum value for this parameter. The
+  /// content repository MAY respond before the timeout.
+  ///
+  ///
+  /// [allowRedirect] Indicates to the server that it may return a 307 or 308 redirect
+  /// response that points at the relevant media content. When not explicitly
+  /// set to `true` the server must return the media content itself.
+  ///
+  Future<FileResponse> getMediaContent(
```
nico-famedly, line 1256: Maybe we shouldn't rename it and regress all upgrading clients to the old behaviour? You can still call the original function using super.getContent.

nico-famedly (approved):

td-famedly (commented):
probably fine

PR #1915, opened 2024-09-09
fix: clear legacy db on logout

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/2399

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2182,14 +2182,22 @@ class Client extends MatrixApi {
   /// Resets all settings and stops the synchronisation.
   Future<void> clear() async {
     Logs().outputEvents.clear();
+    DatabaseApi? legacyDatabase;
+    if (legacyDatabaseBuilder != null) {
+      // If there was data in the legacy db, it will never let the SDK
+      // completely log out as we migrate data from it, everytime we `init`
+      legacyDatabase = await legacyDatabaseBuilder?.call(this);
+    }
     try {
       await abortSync();
       await database?.clear();
+      await legacyDatabase?.clear();
       _backgroundSync = true;
     } catch (e, s) {
       Logs().e('Unable to clear database', e, s);
     } finally {
       await database?.delete();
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -2058,14 +2058,22 @@ class Client extends MatrixApi {
   /// Resets all settings and stops the synchronisation.
```
td-famedly, line 2058: Hm, I don't see _migrateFromLegacyDatabase calling .clear(), when is clear expected to be called?

coder-with-a-bushido: It depends on the implementation but calling `delete` should have been enough. We're calling `clear` just to be sure.

In the case of hive db, we should always call `clear` even if we call delete because it doesn't really delete all the boxes. We could also just fix the `delete` for hive db implementation but that'd mean if someone uses a custom db and does this exact mistake, it'd be hard to track.

I chose to go this way because we already do the same for `database` in logout.

td-famedly: Yes but _migrateFromLegacyDatabase only calls delete, so when does the clear get called? 🤔

coder-with-a-bushido: It doesn't get called and that is the issue. So, I called it in there as well.

td-famedly: found it

lib/src/client.dart
```diff
@@ -2098,14 +2098,22 @@ class Client extends MatrixApi {
...
   Future<void> clear() async {
     Logs().outputEvents.clear();
+    DatabaseApi? legacyDatabase;
```
td-famedly, line 2101: lg, can you add a test for this? should be fairly straight forward

td-famedly: @coder-with-a-bushido PR looks stuck ping :3

coder-with-a-bushido: @td-famedly pls check now

PR #1917, opened 2024-09-12
fix: also lazy load members for archive

nico-famedly:
This should speed up loading the archived rooms. One of the reasons it was so slow, is because we were loading all room members!

Additionally this may work around a bug in Synapse, where rooms stuck in their partial state may block sync indefinitely.

Relates to https://github.com/famedly/product-management/issues/2250


mErrenst (approved):
Probably I can't see a difference because the account I tested it with was too small but at least it doesn't look worse + the code looks good

mErrenst:
Oh, was already merged XD

Issue #1916, opened 2024-09-09
Old left rooms reappear in broken state


emgrav:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [X] I already asked for help in the chat

### In which Project did the bug appear?

Famedly App

### On which platform did the bug appear?

Firefox

### SDK Version

Unknown, but app v2.39.1-0-g0b66f4b1 

### Describe the problem caused by this bug

I logged into my account using element-desktop. After this, a dozen old left rooms from the archive appeared at the top of my room list in Famedly. The rooms are not visible in Element, and appear to be in a broken state. The "last activity" indicator in the room list is always the current time, meaning the rooms are pinned to the top of the room list sorted by activity. The rooms are all called "Empty chat", and attempting to leave them does nothing. Attempting to send a message in the rooms set the last activity to an actual timestamp, unpinning them from the room list. Sending the even did however not succeed.

Nico theorized that element did something with the room archive and account data which caused this to happen.

### Steps To Reproduce

1. Have old rooms from 2022 and 2023 in the archive
2. Log into element-desktop

nico-famedly:
I have also experienced that sometimes when opening the archive, possibly some account data events cause the room to be populated? Needs some debugging.

Clemens-Toegel:
We also experience this problem with our client. In our setup no other client is involved which may cause a problem, we only tested with the matrix-dart-sdk.
The behaviour is the following: In the same session no bug occurs, we can leave a room (lets say _Room A_) and find it then in the room list as archived room with membership leave. After a new login, however, we see  _Room A_ as 'Empty chat' room in our rooms list instead of in the archived rooms list. After debugging I have noticed that the left room _Room A_ is in the archived rooms list as expected, but also in the rooms list (with membership joined!), which is why we see _Room A_ in the rooms list with title 'Empty chat' and membership joined.

I have pinned this behaviour down to the version V30.0, before this version we don't see such a behaviour.

Having a look at the changelog I may think that https://github.com/famedly/matrix-dart-sdk/pull/1866 introduced this behaviour. Specifically, this change:
https://github.com/famedly/matrix-dart-sdk/blob/4fd616bcbf483104d6a1be3c5b3b295b7687b87c/lib/encryption/key_manager.dart#L204

The `JoinedRoomUpdate` may cause the room to be added in the rooms list where it does not belong.

nico-famedly:
[...]

PR #1918, opened 2024-09-17
fix archive rooms becoming joined

nico-famedly:
When decrypting the last event for archived rooms when a room key is received, we used to send a synthetic SyncUpdate. This however put the archived room into the join section, which converted the room to a joined room. We need to respect what section the room was in when sending synthetic SyncUpdates.

fixes https://github.com/famedly/matrix-dart-sdk/issues/1916

### DIFF
--- a/lib/src/database/sqflite_box.dart
+++ b/lib/src/database/sqflite_box.dart
@@ -66,7 +66,7 @@ class BoxCollection with ZoneTransactionMixin {
         },
       );
 
-  Future<void> close() => _db.close();
+  Future<void> close() => zoneTransaction(() => _db.close());
 
   @Deprecated('use collection.deleteDatabase now')
   static Future<void> delete(String path, [dynamic factory]) =>
--- a/test/room_archived_test.dart
+++ b/test/room_archived_test.dart
@@ -23,7 +23,7 @@ import 'package:test/test.dart';
 import 'package:matrix/matrix.dart';
 import 'fake_client.dart';
 
-void main() {
+void main() async {
   group('Timeline', () {
     Logs().level = Level.error;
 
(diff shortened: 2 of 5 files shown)


nico-famedly:
(Also includes a drive by fix for closing the db racing with running transactions)

lib/src/database/sqflite_box.dart
```diff
@@ -66,7 +66,7 @@ class BoxCollection with ZoneTransactionMixin {
         },
       );
 
-  Future<void> close() => _db.close();
+  Future<void> close() => zoneTransaction(() => _db.close());
```
td-famedly, line 69: does the indexedDb impl not need this for some reason?

nico-famedly: I don't really know, it has an assertion, that it is not in a transaction...

nico-famedly: Ah, those were different transactions.

td-famedly: ye but the assert basically throws when we close the db while in a txn, wouldn't it be better to close the db after txn was done? (so in a zonetxn?)

test/room_archived_test.dart
```diff
@@ -115,6 +127,66 @@ void main() {
...
+      // import the inbound session
+      await client.encryption!.keyManager.setInboundGroupSession(
+          roomid,
+          inboundSession.sessionId,
+          inboundSession.senderKey,
+          inboundSession.content);
+
+      expect(client.getArchiveRoomFromCache(roomid) != null, true);
+      expect(client.getRoomById(roomid)?.membership, Membership.leave);
+    }, skip: !encryptionSupported());
```
td-famedly, line 188: just tag the test as olm?

nico-famedly: Yeah, I figured that out on my own! :p

test/room_archived_test.dart
```diff
@@ -115,6 +115,66 @@ void main() {
...
+      // import the inbound session
+      await client.encryption!.keyManager.setInboundGroupSession(
+          roomid,
+          inboundSession.sessionId,
+          inboundSession.senderKey,
+          inboundSession.content);
+
+      expect(client.getArchiveRoomFromCache(roomid) != null, true);
+      expect(client.getRoomById(roomid)?.membership, Membership.leave);
+    }, tags: 'olm');
```
td-famedly, line 176: you still need to init olm iirc

```dart
    setUpAll(() async {
      await olm.init();
      olm.get_library_version();
    });
    ```

nico-famedly: Nope, otherwise how do the tests pass? The client already initializes olm.

td-famedly (approved):
lg

PR #1920, opened 2024-09-21
chore: update Emote-only expressions

TheOneWithTheBraid:
- update the emote base regex
- update documentation about emote computation
- add source for the emote regex
- explain use of regex replace instead of `calcUnlocalizedBody`
- split the RegEx sequences into smaller parts in order to increase maintainability and readability

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -1001,37 +1001,55 @@ class Event extends MatrixEvent {
       content['formatted_body'] is String;
 
   // regexes to fetch the number of emotes, including emoji, and if the message consists of only those
-  // to match an emoji we can use the following regex:
-  // (?:\x{00a9}|\x{00ae}|[\x{2600}-\x{27bf}]|[\x{2b00}-\x{2bff}]|\x{d83c}[\x{d000}-\x{dfff}]|\x{d83d}[\x{d000}-\x{dfff}]|\x{d83e}[\x{d000}-\x{dfff}])[\x{fe00}-\x{fe0f}]?
-  // we need to replace \x{0000} with \u0000, the comment is left in the other format to be able to paste into regex101.com
+  // to match an emoji we can use the following regularly updated regex : https://stackoverflow.com/a/67705964
   // to see if there is a custom emote, we use the following regex: <img[^>]+data-mx-(?:emote|emoticon)(?==|>|\s)[^>]*>
-  // now we combind the two to have four regexes:
+  // now we combined the two to have four regexes and one helper:
+  // 0. the raw components
+  //   - the pure unicode sequence from the link above and
+  //   - the padded sequence with whitespace, option selection and copyright/tm sign
+  //   - the matrix emoticon sequence
   // 1. are there only emoji, or whitespace
   // 2. are there only emoji, emotes, or whitespace
   // 3. count number of emoji
-  // 4- count number of emoji or emotes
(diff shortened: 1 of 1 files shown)


lib/src/event.dart
```diff
@@ -1019,10 +1017,7 @@ class Event extends MatrixEvent {
   /// This is useful to determine if stand-alone emotes should be displayed bigger.
   bool get onlyEmotes {
     if (isRichMessage) {
-      final formattedTextStripped = formattedText.replaceAll(
-          RegExp('<mx-reply>.*</mx-reply>',
-              caseSensitive: false, multiLine: false, dotAll: true),
-          '');
+      final formattedTextStripped = calcUnlocalizedBody(hideReply: true);
```
nico-famedly, line 1020: Can you explain, what exactly this is fixing? Because I think it might not actually remove the reply fallback now?

lib/src/event.dart
```diff
@@ -1035,10 +1030,7 @@ class Event extends MatrixEvent {
   /// WARNING: This does **not** test if there are only emotes. Use `event.onlyEmotes` for that!
   int get numberEmotes {
```
nico-famedly, line 1031: this is now failing in tests and that seems to be a bug in the regex, I think?

lib/src/event.dart
nico-famedly: Actually, no, it seems to strip the reply fallback, the html to text convert does that, it just isn't obvious....

nico-famedly: https://github.com/famedly/matrix-dart-sdk/pull/1924 is somewhat related, where I just try to figure out what that function actually does :p

lib/src/event.dart
TheOneWithTheBraid: Where do you see it failing 👀  ?

lib/src/event.dart
TheOneWithTheBraid: Here we are, I added a comment string on why this is failing right now.

lib/src/event.dart
TheOneWithTheBraid: Should work now, please restart tests.

TheOneWithTheBraid (commented):
Poke @nico-famedly !

lib/src/event.dart
```diff
@@ -1001,37 +1001,50 @@ class Event extends MatrixEvent {
...
+        _unicodeSequences +
+        r')[\ufe00-\ufe0f]?)',
+    caseSensitive: false,
+    multiLine: false,
+  );
   static final RegExp _countEmojiEmoteRegex = RegExp(
-      r'((?:\u00a9|\u00ae|[\u2600-\u27bf]|[\u2b00-\u2bff]|\ud83c[\ud000-\udfff]|\ud83d[\ud000-\udfff]|\ud83e[\ud000-\udfff])[\ufe00-\ufe0f]?|<img[^>]+data-mx-(?:emote|emoticon)(?==|>|\s)[^>]*>)',
-      caseSensitive: false,
-      multiLine: false);
+    r'((?:\x{00a9}|\x{00ae}|' +
```
nico-famedly, line 1035: I don't think there is any reason to switch these to x instead of u, right? Could you change them back for consistency? Otherwise this looks good to me, thanks!

nico-famedly (approved):

PR #1923, opened 2024-09-27
feat: Encrypt file in different isolate

krille-chan:


krille-chan:
I've tried it out but there was no benefit in the performance 🤔 closing it for now

PR #1927, opened 2024-09-30
feat: BREAKING CHANGE add equality, hashcode and toString helper overrides to autogenerated code

td-famedly:
autogen script change: https://github.com/famedly/dart_openapi_codegen/pull/50

### DIFF
--- a/lib/matrix_api_lite/generated/model.dart
+++ b/lib/matrix_api_lite/generated/model.dart
@@ -1,3 +1,6 @@
+import 'dart:core' as dart;
+import 'dart:core';
+
 import 'package:enhanced_enum/enhanced_enum.dart';
 
 import 'package:matrix/matrix_api_lite/model/children_state.dart';
@@ -26,6 +29,16 @@ class HomeserverInformation {
 
   /// The base URL for the homeserver for client-server connections.
   Uri baseUrl;
+
+  @dart.override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      (other is HomeserverInformation &&
+          other.runtimeType == runtimeType &&
+          other.baseUrl == baseUrl);
+
+  @dart.override
+  int get hashCode => baseUrl.hashCode;
 }
 
 ///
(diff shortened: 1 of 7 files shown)


lib/matrix_api_lite/generated/model.dart
```diff
@@ -82,6 +106,20 @@ class DiscoveryInformation {
...
   Map<String, Map<String, Object?>> additionalProperties;
+
+  @override
+  bool operator ==(Object other) =>
+      other is DiscoveryInformation &&
+      other.mHomeserver == mHomeserver &&
+      other.mIdentityServer == mIdentityServer;
+
+  @override
+  int get hashCode => mHomeserver.hashCode ^ mIdentityServer.hashCode;
```
nico-famedly, line 117: Pretty sure this should be

```suggestion
  int get hashCode => Object.hash(mHomeserver, mIdentityServer);
```

td-famedly: done

lib/matrix_api_lite/generated/model.dart
```diff
@@ -209,6 +276,20 @@ class GenerateLoginTokenResponse {
 
   /// The login token for the `m.login.token` login flow.
   String loginToken;
+
+  @override
+  bool operator ==(Object other) =>
+      other is GenerateLoginTokenResponse &&
+      other.expiresInMs == expiresInMs &&
+      other.loginToken == loginToken;
```
nico-famedly, line 284: This should probably not be in the logs by default?

lib/matrix_api_lite/generated/model.dart
```diff
@@ -834,6 +1118,22 @@ class ThreePidCredentials {
 
   /// The session identifier given by the identity server.
   String sid;
+
+  @override
+  bool operator ==(Object other) =>
+      other is ThreePidCredentials &&
+      other.clientSecret == clientSecret &&
+      other.idAccessToken == idAccessToken &&
```
nico-famedly, line 1126: Should we really log this via toString by default?

lib/matrix_api_lite/generated/model.dart
```diff
@@ -1703,6 +2291,26 @@ class LoginResponse {
   /// optionally validating the URLs within. This object takes the same
   /// form as the one returned from .well-known autodiscovery.
   DiscoveryInformation? wellKnown;
+
+  @override
+  bool operator ==(Object other) =>
+      other is LoginResponse &&
+      other.accessToken == accessToken &&
```
nico-famedly, line 2298: We probably shouldn't log this via toString by default either.

nico-famedly (commented):
I'm not sure about the toString stuff. We just had an access token leak and we also now might log user data more easily. I think we really should have to opt in to log stuff like that by default?

lib/matrix_api_lite/generated/model.dart
td-famedly: removed toString

nico-famedly (approved):

PR #1928, opened 2024-10-03
feat: Allow room previews with history visibility world_readable

krille-chan:


lib/src/timeline.dart
```diff
@@ -79,7 +79,10 @@ class Timeline {
   bool _fetchedAllDatabaseEvents = false;
 
   bool get canRequestHistory {
-    if (events.isEmpty) return true;
+    if (events.isEmpty) {
+      return room.prev_batch != null ||
+          room.historyVisibility == HistoryVisibility.worldReadable;
+    }
```
nico-famedly, line 85: I don't think this is correct. The PR sounds like you want to enable paginating the history for unjoined rooms when they are world readable. But this code actually restricts the visibility more. And I think this might even break fragmented timelines? Otoh, the check below also is sketchy in that regard, since we use the chunk.prev_batch for fragmented timelines...

I think we should at least have some tests for this, that verify this actually allows pagination on world readable rooms, it is very hard for me to keep the pagination rules in my head and make an accurate statement about if this code is correct or not :D

krille-chan: Can you explain this more please?

> But this code actually restricts the visibility more.

Why? It returns true when the room history visiblity is world readable which was not the case before. So this un-ristricts it 🤔

nico-famedly: Before it always returned true, which is more often than only if the prevBatch is null or the history visibility is worldReadable. Now, I am not sure how much that matters, but that might break pagination for timeline fragments or loading data from the db in the first step?

krille-chan: hmm, okay I will try to rethink this and reopen it then. Thanks for your feedback

PR #1929, opened 2024-10-04
fix: Change power level without changing memory

krille-chan:
The problem here is that we
have not created a deep copy
of the power level map. By using .copy() we create a
deep copy now.

Closes https://github.com/famedly/product-management/issues/2412

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1189,22 +1189,31 @@ class Room {
   /// Set the power level of the user with the [userID] to the value [power].
   /// Returns the event ID of the new state event. If there is no known
   /// power level event, there might something broken and this returns null.
-  Future<String> setPower(String userID, int power) async {
-    final powerMap = Map<String, Object?>.from(
-      getState(EventTypes.RoomPowerLevels)?.content ?? {},
-    );
+  /// Please note, that you need to await the power level state from sync before
+  /// the changes are actually applied. Especially if you want to set multiple
+  /// power levels at once, you need to await each change in the sync, to not
+  /// override those.
+  Future<String> setPower(String userId, int power) async {
+    final powerLevelMapCopy =
+        getState(EventTypes.RoomPowerLevels)?.content.copy() ?? {};
 
-    final usersPowerMap = powerMap['users'] is Map<String, Object?>
-        ? powerMap['users'] as Map<String, Object?>
-        : (powerMap['users'] = <String, Object?>{});
+    var users = powerLevelMapCopy['users'];
 
-    usersPowerMap[userID] = power;
+    if (users is! Map<String, Object?>) {
+      if (users != null) {
+        Logs().v(
+            'Repairing Power Level "users" has the wrong type "${powerLevelMapCopy['users'].runtimeType}"');
+      }
+      users = powerLevelMapCopy['users'] = <String, Object?>{};
+    }
+
+    users[userId] = power;
 
     return await client.setRoomStateWithKey(
       id,
       EventTypes.RoomPowerLevels,
       '',
-      powerMap,
+      powerLevelMapCopy,
     );
   }
 
--- a/test/room_test.dart
+++ b/test/room_test.dart
@@ -787,6 +787,20 @@ void main() {
       await room.invite('Testname');
     });
 
+    test('setPower', () async {
+      final powerLevelMap =
+          room.getState(EventTypes.RoomPowerLevels, '')!.content.copy();
+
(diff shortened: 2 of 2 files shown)


lib/src/room.dart
```diff
@@ -1187,22 +1187,26 @@ class Room {
   /// Set the power level of the user with the [userID] to the value [power].
   /// Returns the event ID of the new state event. If there is no known
   /// power level event, there might something broken and this returns null.
-  Future<String> setPower(String userID, int power) async {
-    final powerMap = Map<String, Object?>.from(
-      getState(EventTypes.RoomPowerLevels)?.content ?? {},
-    );
+  Future<String> setPower(String userId, int power) async {
```
td-famedly, line 1190: a test would be nice to have :D

lib/src/room.dart
```diff
@@ -1187,22 +1187,26 @@ class Room {
...
+    final powerLevelMapCopy =
+        getState(EventTypes.RoomPowerLevels)?.content.copy() ?? {};
+
+    final usersPowerMap =
+        powerLevelMapCopy.tryGetMap<String, int>('users') ?? {};
 
-    final usersPowerMap = powerMap['users'] is Map<String, Object?>
-        ? powerMap['users'] as Map<String, Object?>
-        : (powerMap['users'] = <String, Object?>{});
+    if (power == getDefaultPowerLevel(powerLevelMapCopy)) {
```
td-famedly, line 1197: Couldn't find anything in the spec which suggests we should do this, but probably fine. Can we add a test for this as well? :D

lib/src/room.dart
krille-chan: Added a test :)

td-famedly (approved):
lg

nico-famedly (approved):
Strictly speaking I guess there could be an edge case if the users map is set to a list, where we won't set the permissions ever now. I think those rooms probably fail to federate and so this is fine and we don't need to care about it :D

lib/src/room.dart
```diff
@@ -1187,22 +1187,18 @@ class Room {
...
-    final powerMap = Map<String, Object?>.from(
-      getState(EventTypes.RoomPowerLevels)?.content ?? {},
-    );
-
-    final usersPowerMap = powerMap['users'] is Map<String, Object?>
-        ? powerMap['users'] as Map<String, Object?>
-        : (powerMap['users'] = <String, Object?>{});
+  Future<String> setPower(String userId, int power) async {
+    final powerLevelMapCopy =
+        getState(EventTypes.RoomPowerLevels)?.content.copy() ?? {};
```
nico-famedly, line 1192: Can you also apply this fix to `enableGroupCalls()` (and possibly a few others)?

lib/src/room.dart
```diff
@@ -1187,22 +1187,18 @@ class Room {
...
 
-    usersPowerMap[userID] = power;
+    powerLevelMapCopy['users'] ??= <String, Object?>{};
+    powerLevelMapCopy.tryGetMap<String, Object?>('users')?[userId] = power;
```
nico-famedly, line 1195: This operates on a copy of `powerLevelMapCopy`, so this actually never modifies the power levels of that user! That's probably also why the tests hang, but for me at least the command test with the op command explicitly fails. Maybe we can add a proper test for this?

lib/src/room.dart
```diff
@@ -1187,22 +1187,18 @@ class Room {
...
 
     return await client.setRoomStateWithKey(
```
nico-famedly, line 1197: this actually now means that the changes will only be reflected in the state event later. I would bet there is some code that depends on this function having immediately reflected the changes. Should we consider this a breaking change or maybe wait for the event to come down in sync? Maybe we can just ignore that, but I think it makes sense to consider the impact of that. For example the following code would now only set the permissions for one of the users:

```dart
await room.setPower('@a:example.org', 10);
await room.setPower('@b:example.org', 10);
```

lib/src/room.dart
krille-chan: Would prefer to do this in a different PR 🤔 one thing after another

lib/src/room.dart
krille-chan: this is also a copy? 😳 oopsie...

lib/src/room.dart
nico-famedly: We discussed this and agreed that just documenting it is the best solution for now.

PR #1932, opened 2024-10-07
chore: Lower loglevel for call event with unexpected sender

krille-chan:
Closes https://github.com/famedly/product-management/issues/2142


td-famedly (approved):
probably fine

PR #1934, opened 2024-10-07
Speedup tests

nico-famedly:
This does various fixes throughout our tests to speed them up as well as prevents us from doing redundant work on sync.

1. Removes a 5 second sleep from the initial sync. There is no need for that and if we want to test something with such a delay, we should do that explicitly instead of in every test.
2. Deduplicates OTK uploads. This means we don't redundantly create OTKs when our upload key request is slower than the next sync. This might cause delays in OTK uploads until the next sync after an upload succeeds, but might also fix a few edge cases, where we may have overwritten our own OTKs multiple times and caused UTDs.
3. Removes a few key uploads in the common tests. We used to only ever return 20 stored OTKs (or null) in sync, which caused us to generate new keys and waste CPU cycles. This only affects our tests.
4. The last commit fixes a few tests that failed now regularly, because the next sync isn't delayed by 5s. I mostly replaced those cases with explicitly cancelling the background sync loop, which is pretty much equivalent to delaying the next sync for 5s in the common case of tests taking less than 5s. There are probably a few more race conditions still hiding, so possibly the tests will be a bit unstable for a while, but those tend to be rather easy to fix and should make the tests more stable in general in the long term.

All in all this makes the tests complete in 45s where before it took around 5 minutes on my machine.

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -50,14 +50,15 @@ jobs:
           retention-days: 1
 
   coverage:
-    runs-on: arm-ubuntu-latest-16core
+    #runs-on: arm-ubuntu-latest-16core
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - run: cat .github/workflows/versions.env >> $GITHUB_ENV
       - uses: dart-lang/setup-dart@a57a6c04cf7d4840e88432aad6281d1e125f0d46
         with:
           sdk: ${{ env.dart_version }}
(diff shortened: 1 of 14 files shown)


nico-famedly:
Okay, switching the runners to x86 makes the tests exit properly, should we just do that? I think this is a weird dart bug, but I can't figure out how to fix it otherwise and this would also make our CI a bit cheaper (and not much slower).

krille-chan (approved):

PR #1937, opened 2024-10-12
feat: improve commands_extension

TheOneWithTheBraid:
While implementing some custom commands, I stumbled across some issues with the current command runner :

1. Currently, even Client-level commands such as `/join`, `/dm` etc. require to be entered with a `Room` present as parameter
2. Currently, commands have no way to provide output, such as a success message, some reference to e.g. a newly created room or any other information since the return parameter is currently usually an eventId used by the clients to confirm sending a message was successful.
3. Currently, \[matrix\] clients have no way to catch exceptions specific to the command execution, such as a missing or invalid parameter etc.

I have the following proposals to address these issues in the SDK's command execution :

- unify behavior of all message sending related command
- add a `StringBuffer` as stdout-like output buffer for commands
- create a `typedef` for the command function signature
- create a common exception type for command execution
- enable commands to run on Client-level rather than Room-level
- BREAKING: `Client.addCommand` signature now takes an optional `StringBuffer` as second parameter

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -98,7 +98,7 @@ class Client extends MatrixApi {
   DateTime? _accessTokenExpiresAt;
 
   // For CommandsClientExtension
-  final Map<String, FutureOr<String?> Function(CommandArgs)> commands = {};
+  final Map<String, CommandExecutionCallback> commands = {};
   final Filter syncFilter;
 
   final NativeImplementations nativeImplementations;
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -625,6 +625,7 @@ class Room {
     String msgtype = MessageTypes.Text,
     String? threadRootEventId,
     String? threadLastEventId,
+    StringBuffer? commandStdout,
   }) {
     if (parseCommands) {
       return client.parseAndRunCommand(
@@ -635,6 +636,7 @@ class Room {
         txid: txid,
         threadRootEventId: threadRootEventId,
         threadLastEventId: threadLastEventId,
+        stdout: commandStdout,
       );
     }
     final event = <String, dynamic>{
--- a/lib/src/utils/commands_extension.dart
+++ b/lib/src/utils/commands_extension.dart
@@ -21,31 +21,45 @@ import 'dart:convert';
 
 import 'package:matrix/matrix.dart';
 
+/// callback taking [CommandArgs] as input and a [StringBuffer] as standard output
+/// optionally returns an event ID as in the [Room.sendEvent] syntax.
+/// a [CommandException] should be thrown if the specified arguments are considered invalid
+typedef CommandExecutionCallback = FutureOr<String?> Function(
+  CommandArgs,
+  StringBuffer? stdout,
+);
+
 extension CommandsClientExtension on Client {
   /// Add a command to the command handler. `command` is its name, and `callback` is the
   /// callback to invoke
-  void addCommand(
-    String command,
-    FutureOr<String?> Function(CommandArgs) callback,
-  ) {
+  void addCommand(String command, CommandExecutionCallback callback) {
     commands[command.toLowerCase()] = callback;
   }
 
-  /// Parse and execute a string, `msg` is the input. Optionally `inReplyTo` is the event being
-  /// replied to and `editEventId` is the eventId of the event being replied to
+  /// Parse and execute a command on Client level
+  /// - `room`: a [Room] to run the command on. Can be null unless you execute a command strictly requiring a [Room] to run on
+  /// - `msg`: the complete input to process
+  /// - `inReplyTo`: an optional [Event] the command is supposed to reply to
+  /// - `editEventId`: an optional event ID the command is supposed to edit
+  /// - `txid`: an optional transaction ID
+  /// - `threadRootEventId`: an optional root event ID of a thread the command is supposed to run on
+  /// - `threadLastEventId`: an optional most recent event ID of a thread the command is supposed to run on
+  /// - `stdout`: an optional [StringBuffer] the command can write output to. This is meant as tiny implementation of https://en.wikipedia.org/wiki/Standard_streams in order to process advanced command output to the matrix client. See [DefaultCommandOutput] for a rough idea.
   Future<String?> parseAndRunCommand(
-    Room room,
+    Room? room,
     String msg, {
     Event? inReplyTo,
     String? editEventId,
     String? txid,
     String? threadRootEventId,
     String? threadLastEventId,
+    StringBuffer? stdout,
   }) async {
     final args = CommandArgs(
       inReplyTo: inReplyTo,
       editEventId: editEventId,
       msg: '',
+      client: this,
       room: room,
       txid: txid,
       threadRootEventId: threadRootEventId,
@@ -55,7 +69,7 @@ extension CommandsClientExtension on Client {
       final sendCommand = commands['send'];
       if (sendCommand != null) {
         args.msg = msg;
-        return await sendCommand(args);
+        return await sendCommand(args, stdout);
       }
       return null;
     }
@@ -71,14 +85,14 @@ extension CommandsClientExtension on Client {
     }
     final commandOp = commands[command];
     if (commandOp != null) {
-      return await commandOp(args);
+      return await commandOp(args, stdout);
     }
     if (msg.startsWith('/') && commands.containsKey('send')) {
       // re-set to include the "command"
       final sendCommand = commands['send'];
       if (sendCommand != null) {
         args.msg = msg;
-        return await sendCommand(args);
+        return await sendCommand(args, stdout);
       }
     }
     return null;
@@ -91,8 +105,12 @@ extension CommandsClientExtension on Client {
 
   /// Register all default commands
   void registerDefaultCommands() {
-    addCommand('send', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('send', (args, stdout) async {
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendTextEvent(
         args.msg,
         inReplyTo: args.inReplyTo,
         editEventId: args.editEventId,
@@ -102,8 +120,12 @@ extension CommandsClientExtension on Client {
         threadLastEventId: args.threadLastEventId,
       );
     });
-    addCommand('me', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('me', (args, stdout) async {
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendTextEvent(
         args.msg,
         inReplyTo: args.inReplyTo,
         editEventId: args.editEventId,
@@ -114,21 +136,44 @@ extension CommandsClientExtension on Client {
         threadLastEventId: args.threadLastEventId,
       );
     });
-    addCommand('dm', (CommandArgs args) async {
+    addCommand('dm', (args, stdout) async {
       final parts = args.msg.split(' ');
-      return await args.room.client.startDirectChat(
-        parts.first,
+      final mxid = parts.first;
+      if (!mxid.isValidMatrixId) {
+        throw CommandException('You must enter a valid mxid when using /dm');
+      }
+
+      final roomId = await args.client.startDirectChat(
+        mxid,
         enableEncryption: !parts.any((part) => part == '--no-encryption'),
       );
-    });
-    addCommand('create', (CommandArgs args) async {
-      final parts = args.msg.split(' ');
-      return await args.room.client.createGroupChat(
-        enableEncryption: !parts.any((part) => part == '--no-encryption'),
+      stdout?.write(
+        DefaultCommandOutput(
+          rooms: [roomId],
+          users: [mxid],
+        ).toString(),
       );
+      return null;
     });
-    addCommand('plain', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('create', (args, stdout) async {
+      final groupName = args.msg.replaceFirst('--no-encryption', '').trim();
+
+      final parts = args.msg.split(' ');
+
+      final roomId = await args.client.createGroupChat(
+        groupName: groupName.isNotEmpty ? groupName : null,
+        enableEncryption: !parts.any((part) => part == '--no-encryption'),
+        waitForSync: false,
+      );
+      stdout?.write(DefaultCommandOutput(rooms: [roomId]).toString());
+      return null;
+    });
+    addCommand('plain', (args, stdout) async {
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendTextEvent(
         args.msg,
         inReplyTo: args.inReplyTo,
         editEventId: args.editEventId,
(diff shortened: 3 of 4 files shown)


lib/src/utils/commands_extension.dart
```diff
@@ -21,29 +21,38 @@ import 'dart:convert';
 
 import 'package:matrix/matrix.dart';
 
+/// callback taking [CommandArgs] as input and a [StringBuffer] as standard output
+/// optionally returns an event ID as in the [Room.sendEvent] syntax.
+/// a [CommandException] should be thrown if the specified arguments are considered invalid
+typedef CommandExecutionCallback = FutureOr<String?> Function(
+  CommandArgs,
+  StringBuffer? stdout,
```
coder-with-a-bushido, line 29: I don't see the `stdout` used in any of the default commands. Even though having an output buffer sounds like an important thing to have, I doubt how useful it'll be. I just can't think of a really good use-case for it right now, could you pls tell me a good way it can be used (to output useful data) with the default commands?

coder-with-a-bushido:
This PR looks really good to me except that one thing I commented about :)
Thank you so much!

lib/src/utils/commands_extension.dart
TheOneWithTheBraid: Yes, it's true, I didn't change the behavior of the default commands. Mostly, since they do not really have anything to output because they are pretty static commands without any kind of user interaction.

I personally use the stdout to modify the message text field : I e.g. implemented a `/poll` command assisting in sending composing Element polls, I added a `/search` command returning a protobuf content stream. I intentionally didn't specify any high-level format to use for stdout since I can imagine infinite different use cases how clients would like to interact with the output of commands that is not an eventId - that's why ; a simple `StreamBuffer` and nothing further.

TheOneWithTheBraid: One could consider adding JSON output of the `/dm` and `/create` commands containing the room ID though. If you like, I will add this.

TheOneWithTheBraid: @coder-with-a-bushido See latest comment. I added a helper class for command output and now add the room IDs of created/searched rooms to the stdout in the default commands.

krille-chan (changes requested):
May I ask where the issue is for this refactoring? Such a big change should go through a refinement first IMO

TheOneWithTheBraid:
> May I ask where the issue is for this refactoring? Such a big change should go through a refinement first IMO

I'm not bound to any work processes when I'm contributing in my free time. The PR has half a page of documentation why and how the change was made. If you have question or would like to discuss the implementation, feel free to leave a constructive review comment. Since the API was not changed in a way relevant for client developers, I do not see the need of outlining a dedicated change proposal in advance.

krille-chan:
> > May I ask where the issue is for this refactoring? Such a big change should go through a refinement first IMO
> 
> I'm not bound to any work processes when I'm contributing in my free time. The PR has half a page of documentation why and how the change was made. If you have question or would like to discuss the implementation, feel free to leave a constructive review comment. Since the API was not changed in a way relevant for client developers, I do not see the need of outlining a dedicated change proposal in advance.

Well it would still have been nice to communicate such a change in advance instead of just expecting that we would spend our time in reviewing hundreds of code lines without knowing if it would have any benefit

TheOneWithTheBraid:
Yeah, I see the problem. It's quite hard to have work processes on an open source open contribution project. Maybe some outline on starting from what amount of change (e.g. everything which is a new feature, changes API etc.) one should write an issue for discussion ? Might also help other external contributors ?

Currently the [CONTRIBUTING.md](https://github.com/famedly/matrix-dart-sdk/blob/main/CONTRIBUTING.md) only sais to open a merge request and discuss over there, that's why I so far did that for personal contributions, maybe you can update this (btw. it still sais GitLab too ^^).

krille-chan:
> Yeah, I see the problem. It's quite hard to have work processes on an open source open contribution project. Maybe some outline on starting from what amount of change (e.g. everything which is a new feature, changes API etc.) one should write an issue for discussion ? Might also help other external contributors ?
> 
> Currently the [CONTRIBUTING.md](https://github.com/famedly/matrix-dart-sdk/blob/main/CONTRIBUTING.md) only sais to open a merge request and discuss over there, that's why I so far did that for personal contributions, maybe you can update this (btw. it still sais GitLab too ^^).

okay maybe I should rephrase my comments so far. Of course it is okay that everyone opens Pull Requests and we will always do our best to review them. The problem is only that we cannot guarantuee that the changes get merged. Especially for changes which change the API. Maybe we are not satisfied with how it looks like, the approach to fix it and so on.

The interest conflict here is mostly because external contributors have the desire to fix one specific thing, but usually don't have to deal with the code base after it, while we have to deal with the code base as it is our daily work. So we have a strong desire to keep the code base in a level of quality, which fits our needs.

 For small changes this shouldn't be a big deal. For big changes, which apply to hundreds of code lines, it could mean, that we either request to change a lot of things so that it matches our interests as well, or we completely decline it at all. Then the whole work on the Pull Request would be nearly gone with the potential to leave the author back with some frustration.
Therefore we **suggest** to propose bigger changes in an issue first. Opening an issue, explaining the need and the idea how to solve it. Then we could comment if we think this is a good idea or give feedback before the implementation starts. This could drastically reduce the dangers that the Pull Request in the end gets merged.

In the end one could argue that it is not our problem but leaving pull request authors back with frustration might lead to shit talking about us. Especially from you I received a lot of FluffyChat pull requests where I felt a lot of pressure to merge them and even frustration when it took longer. This is something which could have been avoided.

I admit that my previous comment does not really reflected this and we should update our contribution guideline.

lib/src/room.dart
```diff
@@ -600,22 +600,29 @@ class Room {
...
+    String message, {
+    String? txid,
+    Event? inReplyTo,
+    String? editEventId,
+    bool parseMarkdown = true,
+    bool parseCommands = true,
+    String msgtype = MessageTypes.Text,
+    String? threadRootEventId,
+    String? threadLastEventId,
+    StringBuffer? commandStdout,
```
krille-chan, line 613: Can we add a little bit dartdoc comments to make it more intuitive how to use the `StringBuffer`? Or at least a link to the documentation. For example I see this type for the first time and would have to look it up before using it.

lib/src/client.dart
```diff
@@ -97,7 +97,7 @@ class Client extends MatrixApi {
   DateTime? _accessTokenExpiresAt;
 
   // For CommandsClientExtension
-  final Map<String, FutureOr<String?> Function(CommandArgs)> commands = {};
+  final Map<String, CommandExecutionCallback> commands = {};
```
krille-chan, line 100: I'm a little bit confused how to actually trigger a command by using a Client without a Room. Can you add a test for it?

krille-chan: Ah, probably by using `Client.parseAndRunCommand()`, right? A test for a command without a room could still be useful

lib/src/room.dart
krille-chan: We might want to have more information from the output as just a String, as errors could also be interesting, while we also might get errors and values at the same time. Could be probably solved by introducing a CommandResult type which contains the success data, (Matrix-)errors as well as the output of the command

krille-chan (commented):
We looked at it on the refinement meeting together and we agree with the technical approach and the API changes. The usual review will follow. Thank you very much for the contribution 😊

lib/src/room.dart
TheOneWithTheBraid: Hmmmm, in my implementation I simply used the regular error catching around the `Client.parseAndRunCommand()` and `Room.send*` methods for error handling. Since there are decent, unified Error types available I don't see an advantage in a dedicated error collection mechanism ; but if you have any proposal on how to implement some improved error handling, I'm curious for suggestions !

The advantage of a `StringBuffer` is that it basically allows to perform random access (similar to a `File.open` call) and object-oriented append syntax. I simply considered this as way more convenient for this use case - since a String could easily accidentally lose it's reference and doesn't support such a beautiful syntax to access its onwardly growing contents during the asynchronous writing into it.

lib/src/client.dart
TheOneWithTheBraid: You are right, I will add a test for a non-room related call.

TheOneWithTheBraid: I added a couple of tests for non-room level calls with some different test cases.

lib/src/room.dart
coder-with-a-bushido: @TheOneWithTheBraid Can we please have dartdoc comment for the `commandStdout` that says it's used for the command output in case a command was executed

lib/src/utils/commands_extension.dart
```diff
@@ -21,31 +21,38 @@ import 'dart:convert';
...
-    String command,
-    FutureOr<String?> Function(CommandArgs) callback,
-  ) {
+  void addCommand(String command, CommandExecutionCallback callback) {
     commands[command.toLowerCase()] = callback;
   }
 
   /// Parse and execute a string, `msg` is the input. Optionally `inReplyTo` is the event being
   /// replied to and `editEventId` is the eventId of the event being replied to
   Future<String?> parseAndRunCommand(
```
coder-with-a-bushido, line 41: Can we also add a dartdoc comment here saying "room object can be null here only when you are very sure the msg doesn't contain room specific commands"

lib/src/utils/commands_extension.dart
```diff
@@ -114,21 +129,43 @@ extension CommandsClientExtension on Client {
...
+        DefaultCommandOutput(
+          rooms: [roomId],
+          users: [mxid],
+        ).toString(),
+      );
+      return null;
     });
-    addCommand('create', (CommandArgs args) async {
+    addCommand('create', (args, stdout) async {
+      final groupName = args.msg.replaceFirst('--no-encryption', '').trim();
```
coder-with-a-bushido, line 152: can we also make sure that `groupName` is null when the result of `replaceFirst` followed by `trim` is an empty string? It would prevent sending a `m.room.name` event with an empty string when room name is not specified.

lib/src/utils/commands_extension.dart
```diff
@@ -91,8 +98,12 @@ extension CommandsClientExtension on Client {
 
   /// Register all default commands
   void registerDefaultCommands() {
-    addCommand('send', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('send', (args, stdout) async {
```
coder-with-a-bushido, line 101: we could add the `eventId` in the stdout here?

lib/src/utils/commands_extension.dart
```diff
@@ -102,8 +113,12 @@ extension CommandsClientExtension on Client {
...
       );
     });
-    addCommand('me', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('me', (args, stdout) async {
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendTextEvent(
```
coder-with-a-bushido, line 121: same here?

lib/src/utils/commands_extension.dart
```diff
@@ -114,21 +129,43 @@ extension CommandsClientExtension on Client {
...
+      return null;
     });
-    addCommand('plain', (CommandArgs args) async {
-      return await args.room.sendTextEvent(
+    addCommand('plain', (args, stdout) async {
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendTextEvent(
```
coder-with-a-bushido, line 168: same here?

lib/src/utils/commands_extension.dart
```diff
@@ -139,168 +176,280 @@ extension CommandsClientExtension on Client {
...
         'body': args.msg,
         'format': 'org.matrix.custom.html',
         'formatted_body': args.msg,
       };
-      return await args.room.sendEvent(
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      return await room.sendEvent(
```
coder-with-a-bushido, line 190: here too?

lib/src/utils/commands_extension.dart
```diff
@@ -139,168 +176,280 @@ extension CommandsClientExtension on Client {
...
+      final room = args.room;
+      if (room == null) {
+        throw RoomCommandException();
+      }
+      final parts = args.msg.split(' ');
+      final reaction = parts.first.trim();
+      if (reaction.isEmpty) {
+        throw CommandException('You must provide a reaction when using /react');
+      }
+      return await room.sendReaction(inReplyTo.eventId, reaction);
```
coder-with-a-bushido, line 211: here too?

lib/src/utils/commands_extension.dart
```diff
@@ -139,168 +176,280 @@ extension CommandsClientExtension on Client {
...
         pl = int.tryParse(parts[1]);
+        if (pl == null) {
+          throw CommandException(
+            'Invalid power level ${parts[1]} when using /op',
+          );
+        }
       }
       final mxid = parts.first;
-      return await args.room.setPower(mxid, pl ?? 50);
+      return await room.setPower(mxid, pl ?? 50);
```
coder-with-a-bushido, line 245: same here?

coder-with-a-bushido: maybe also the mxid

lib/src/utils/commands_extension.dart
```diff
@@ -139,168 +176,280 @@ extension CommandsClientExtension on Client {
...
+      final currentEventJson = room
+              .getState(EventTypes.RoomMember, args.client.userID!)
               ?.content
               .copy() ??
           {};
       currentEventJson['avatar_url'] = args.msg;
-      return await args.room.client.setRoomStateWithKey(
-        args.room.id,
+
+      return await args.client.setRoomStateWithKey(
```
coder-with-a-bushido, line 339: same here?

lib/src/utils/commands_extension.dart
```diff
@@ -139,168 +176,280 @@ extension CommandsClientExtension on Client {
...
+      final currentEventJson = room
+              .getState(EventTypes.RoomMember, args.client.userID!)
               ?.content
               .copy() ??
           {};
       currentEventJson['displayname'] = args.msg;
-      return await args.room.client.setRoomStateWithKey(
-        args.room.id,
+
+      return await args.client.setRoomStateWithKey(
```
coder-with-a-bushido, line 319: here too?

lib/src/utils/commands_extension.dart
```diff
@@ -114,21 +129,43 @@ extension CommandsClientExtension on Client {
         threadLastEventId: args.threadLastEventId,
       );
     });
-    addCommand('dm', (CommandArgs args) async {
+    addCommand('dm', (args, stdout) async {
       final parts = args.msg.split(' ');
-      return await args.room.client.startDirectChat(
-        parts.first,
+      final mxid = parts.first;
```
coder-with-a-bushido, line 134: should we use the same approach like in '/create' to get the mxid here?

test/commands_test.dart
```diff
@@ -527,6 +527,65 @@ void main() {
...
+        '!1234:fakeServer.notExisting',
+      );
+    });
+
+    test('client - create', () async {
+      FakeMatrixApi.calledEndpoints.clear();
+      final stdout = StringBuffer();
+      await client.parseAndRunCommand(
+        null,
+        '/create @alice:example.com --no-encryption',
```
coder-with-a-bushido, line 563: in this case, "@alice:example.com" is the room name?

coder-with-a-bushido (changes requested):
@TheOneWithTheBraid Sorry for the super late review 😖

Code looks really good to me except for those nitpick comments. Now that you've used the `stdout` in some commands, I somehow want to use it all commands possible so it can be used

test/commands_test.dart
coder-with-a-bushido: Hmm, like I guessed the test is failing in CI. I think the command should be something like '/create some random room --no-encryption'. Then you should check the request body.

lib/src/utils/commands_extension.dart
TheOneWithTheBraid: I'd object this since sending an Event is the default "action" on the `Room.send`. The event ID always used to be returned as result of the method, not as command output. I'd rather see the standard output StrinBuffer as a complimentary concept for *additional* information which is exactly *not* the default behavior (aka sending an event).

lib/src/utils/commands_extension.dart
TheOneWithTheBraid: The mxid is already available to the user since it's a user provided argument. I'd consider a MXID rather as standard output in cases where users might e.g. search for an MXID or similar as command output ?

lib/src/utils/commands_extension.dart
TheOneWithTheBraid: Consider the following flow :

As a user, I enter something into a message text field of my matrix client. I send the contents via the `Room.send*` methods.

The matrix client underneath now does not know at first if I just sent a regular message or if a command was executed. As the matrix client, I will know about this by checking the stdout. If there is data in the stdout, I consider the result as command result and show e.g. a dialog or a snack bar with a details button where I can see the stdout.

If the stdout is empty and I simply got back an event id as return value, as a matrix client, I simply know the user just sent a message.

If we'd by default add the event id to the stdout, the entire flow of determination whether a command with special result was executed or just a simple event was sent stops working.

coder-with-a-bushido: I'm just trying to understand the use case for stdout once again, pls correct me if I'm wrong.

I use `Room.send` with a text while I'm unsure if it's actually a command or just a text that needs to be sent. So, what I do is I create a `stdout` object that will return me something (can be any meaningful data) when it's **a command that does something other than sending an event**.

If that's the case, as the user of the SDK I want 2 things to be done:
1. Dartdoc comments that clearly explains this context wherever `stdout` is mentioned.
2. An example usage of this scenario in both README and dartdoc comment above `Room.send`.

If I didn't understand it correctly, please tell me.

lib/src/room.dart
coder-with-a-bushido: @TheOneWithTheBraid ping

> @TheOneWithTheBraid Can we please have dartdoc comment for the `commandStdout` that says it's used for the command output in case a command was executed

coder-with-a-bushido:
> @coder-with-a-bushido Thanks for the feedback ! I (hopefully) fixed the test and added two comments above, I'd appreciate some feedback. Have a good weekend !

Hey, thank you. But looks like 1 test is still broken and I think 2 more comments should be resolved before we can merge this :)

TheOneWithTheBraid:
Tests should now be fixed, @coder-with-a-bushido !

TheOneWithTheBraid:
@coder-with-a-bushido Can you elaborate on what's wrong with the coverage test ?

lib/src/client.dart
```diff
@@ -98,7 +98,7 @@ class Client extends MatrixApi {
...
```
coder-with-a-bushido, line 101: Can you also pls add "BREAKING" in the "feat: improve command_extension" commit message?

coder-with-a-bushido:
> > @coder-with-a-bushido Can you elaborate on what's wrong with the coverage test ?
> 
> I don't understand it either. The test is breaking in two totally different files, let me try it locally first. Meanwhile, can you pls address the 2 other comments left?

I think you should just rebase it to upstream/main. That should work.

Also, make sure to squash your commits and add "BREAKING" right after "feat: " in the commit message.

coder-with-a-bushido:
@TheOneWithTheBraid there was a dependency issue that's fixed in main now. If you rebase to main, that should fix the e2ee tests and for the coverage, you have to do the following change.

Replace the following line at `client_test.dart` line 1276:
```   
expect(client2.rooms.length, 4);
```

All that's left would be to address the 2 left out comments, squash the commits and add a "BREAKING" in the commit message. I'm sorry this got stuck so long.

lib/fake_matrix_api.dart
coder-with-a-bushido, line 1: I tested it and looks like removing this file change will fix the coverage test

coder-with-a-bushido: Probably got to do something with concurrent tests changing the `client.rooms` number in between the "Test the fake store api" test

TheOneWithTheBraid:
@coder-with-a-bushido I finally found the time to add the relevant documentation. Please ensure to add the `BREAKING` when squashing and re-signing the commits.

lib/src/utils/commands_extension.dart
```diff
@@ -319,9 +477,92 @@ class CommandArgs {
...
+
+/// Helper class for normalized command output
+///
+/// This class can be used to provide a default, processable output of commands
+/// containing some generic data.
+///
+/// When simply sending an event to a room, please ensure not to add the
+/// event ID here but simply use `return` instead. Otherwise clients may not be
+/// able to distinguish between an eventual search result or similar output and
+/// just the default [Room.sendTextEvent] behavior.
```
coder-with-a-bushido, line 511: Can you pls rephrase this? I don't quite understand what this means

coder-with-a-bushido:
```
NOTE:
If your command actually sends an event to a room, please do not include the event ID here. The default behavior of the [Room.sendTextEvent] is to return the event ID of the just sent event. The [DefaultCommandOutput.events] field is not supposed to replace/duplicate this behavior.

But if your command performs an action such as search, highlight or anything your matrix client should display different than adding an event to the [Timeline], you can include the event IDs related to the command output here.
```

coder-with-a-bushido (approved):
LGTM :)

PR #1939, opened 2024-10-18
feat: v1.12 spec endpoints support

td-famedly:


lib/matrix_api_lite/generated/api.dart
```diff
@@ -1993,12 +1996,14 @@ class Api {
...
   Future<String> joinRoom(String roomIdOrAlias,
       {List<String>? serverName,
+      List<String>? via,
       String? reason,
       ThirdPartySigned? thirdPartySigned}) async {
     final requestUri = Uri(
         path: '_matrix/client/v3/join/${Uri.encodeComponent(roomIdOrAlias)}',
         queryParameters: {
           if (serverName != null) 'server_name': serverName,
+          if (via != null) 'via': via,
```
nico-famedly, line 2006: You should really just set both when passing one. serverName is deprecated, but should still be sent for compat. I guess you could add a compat shim in the client for it.

nico-famedly (approved):
Most of these changes just look like churn for no benefit, but apart from the serverName stuff, nothing too bad. Probably good the spec is getting cleaned up?

PR #1940, opened 2024-10-21
chore: Upgrade dependencies

nico-famedly:
The library annotations are a new lint and we have no need to specify a custom library name anywhere. The JS annoation on the library also did nothing, since it doesn't default to the library name in dart and only modifies the prefix of functions if you pass in a custom name.

The sqflite_common_ffi I removed the upper constraint, which allows us to use a newer version if our dart version is new enough, but I left it at the original version because there is no need to require a higher minimum currently.

### DIFF
--- a/lib/encryption.dart
+++ b/lib/encryption.dart
@@ -17,7 +17,7 @@
  */
 
 /// Matrix SDK encryption specific extension
-library encryption;
+library;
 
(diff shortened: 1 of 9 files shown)


nico-famedly:
The weird mime constraint is to make it compatible also with dart 3.4.0

krille-chan (approved):

PR #1942, opened 2024-10-21
fix: AsyncCache is not invalidating on error

krille-chan:
This fixes for example that if a
fetch from an asynccache once
failed it never gets retried. So we
only want to cache results, not
exceptions/errors


nico-famedly (approved):
Seems reasonable to me

PR #1944, opened 2024-10-23
feat: Optional authentication for profile requests

krille-chan:
Closes https://github.com/famedly/product-management/issues/2537

Result of https://github.com/famedly/dart_openapi_codegen/pull/53


lib/matrix_api_lite/generated/api.dart
```diff
@@ -14,6 +14,8 @@ import 'package:matrix/matrix_api_lite/model/sync_update.dart';
 
 // ignore_for_file: provide_deprecation_message
 
+
```
td-famedly, line 17: needs formatting, someone should fix this in the autogen :D

td-famedly (approved):
lg

PR #1946, opened 2024-10-23
feat: Add deleteFile() endpoint to database

krille-chan:
Prepares other features

### DIFF
--- a/lib/src/database/database_file_storage_io.dart
+++ b/lib/src/database/database_file_storage_io.dart
@@ -9,13 +9,15 @@ mixin DatabaseFileStorage {
   late final Uri? fileStorageLocation;
   late final Duration? deleteFilesAfterDuration;
 
+  File _getFileFromMxc(Uri mxcUri) => File(
+        '${Directory.fromUri(fileStorageLocation!).path}/${mxcUri.toString().split('/').last}',
+      );
+
   Future<void> storeFile(Uri mxcUri, Uint8List bytes, int time) async {
     final fileStorageLocation = this.fileStorageLocation;
     if (!supportsFileStoring || fileStorageLocation == null) return;
 
-    final dir = Directory.fromUri(fileStorageLocation);
-
-    final file = File('${dir.path}/${mxcUri.toString().split('/').last}');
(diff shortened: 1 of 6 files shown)


nico-famedly:
Is this a duplicate of #1945?

lib/src/database/database_file_storage_io.dart
```diff
@@ -33,6 +33,20 @@ mixin DatabaseFileStorage {
     return null;
   }
 
+  Future<bool> deleteFile(Uri mxcUri) async {
+    final fileStorageLocation = this.fileStorageLocation;
+    if (!supportsFileStoring || fileStorageLocation == null) return false;
+
+    final dir = Directory.fromUri(fileStorageLocation);
+
+    final file = File('${dir.path}/${mxcUri.toString().split('/').last}');
```
nico-famedly, line 42: We repeat this code a lot now. Can we somehow put that in some shared code and add some validation, that the last part isn't `..` or similar? Also, because we don't use the full namespacing, this can clobber files from other servers if the server is malicious, maybe we want to fix that as well? :)

nico-famedly:
The description sounds a bit ominous, what new features does this prepare? :D

test/database_api_test.dart
```diff
@@ -78,11 +78,14 @@ void main() {
...
         expect(toDeviceQueue.isEmpty, true);
       });
-      test('storeFile', () async {
+      test('storeFile and deleteFile', () async {
         await database.storeFile(
             Uri.parse('mxc://test'), Uint8List.fromList([0]), 0);
         final file = await database.getFile(Uri.parse('mxc://test'));
         expect(file != null, database.supportsFileStoring);
+
+        final result = await database.deleteFile(Uri.parse('mxc://test'));
```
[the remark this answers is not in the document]
krille-chan: Done :)

lib/src/database/database_file_storage_io.dart
krille-chan: Done :)

nico-famedly (approved):

Issue #1947, opened 2024-10-28
`PushNotification`'s `devices` field is not compatible with Sygnal


vanyasem:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

`matrix-dart-sdk`

### On which platform did the bug appear?

Android, iOS

### SDK Version

0.34.0

### Describe the problem caused by this bug

`PushNotification` object in `martix-dart-sdk` is not compatible with the reference implementation of a Push Gateway for Matrix called Sygnal

The `PushNotification` object in the SDK has a mandatory `devices` field: https://github.com/famedly/matrix-dart-sdk/blob/v0.34.0/lib/src/utils/push_notification.dart#L20 which is not sent by Sygnal

I have only tested the GCM (FCM) implementation of Sygnal, but by reading the [appropriate source code](https://github.com/matrix-org/sygnal/blob/v0.15.1/sygnal/apnspushkin.py), I can confirm that this issue affects APN notifications as well

See https://github.com/matrix-org/sygnal/pull/319 for comments of the maintainer of Sygnal

As far as I could tell, despite being required, it's not used by the SDK in any capacity

I believe the confusion happened because of the following text in the matrix push gateway API spec: https://spec.matrix.org/v1.12/push-gateway-api/
<img width="624" src="https://github.com/user-attachments/assets/d73fde4b-0796-46c2-83c8-8be25376d43d">

This field is only marked as required for server-side endpoints of the push gateway, and should not be required on the client-side

I made a PR that makes the `devices` field optional: https://github.com/famedly/matrix-dart-sdk/pull/1949

vanyasem:
You can send a fake test notification to a device with `curl`: https://github.com/matrix-org/sygnal/blob/v0.15.1/docs/troubleshooting.md#sending-a-notification-to-sygnal-manually-with-curl

Sending the following notification body:
```json
{
  "notification": {
    "event_id": "\$3957tyerfgewrf384",
    "room_id": "!slw48wfj34rtnrf:example.org",
    "type": "m.room.message",
    "sender": "@exampleuser:example.org",
    "sender_display_name": "Major Tom",
    "room_name": "Mission Control",
    "room_alias": "#exampleroom:example.org",
    "prio": "high",
    "content": {
      "msgtype": "m.text",
      "body": "I'm floating in a most peculiar way."
    },
    "counts": {
      "unread": 2,
[...]

PR #1949, opened 2024-10-28
PushNotification: Make 'devices' field optional

vanyasem:
Fixes https://github.com/famedly/matrix-dart-sdk/issues/1947

### DIFF
--- a/lib/src/utils/push_notification.dart
+++ b/lib/src/utils/push_notification.dart
@@ -4,7 +4,7 @@ import 'dart:convert';
 class PushNotification {
   final Map<String, dynamic>? content;
   final PushNotificationCounts? counts;
-  final List<PushNotificationDevice> devices;
+  final List<PushNotificationDevice>? devices;
   final String? eventId;
   final String? prio;
   final String? roomAlias;
@@ -17,7 +17,7 @@ class PushNotification {
   const PushNotification({
     this.content,
     this.counts,
-    required this.devices,
+    this.devices,
     this.eventId,
     this.prio,
     this.roomAlias,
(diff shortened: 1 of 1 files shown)


krille-chan (changes requested):
According to  https://spec.matrix.org/v1.12/push-gateway-api/#post_matrixpushv1notify this is wrong. The devices field is required

vanyasem:
Btw the correct documentation to link would be https://github.com/matrix-org/sygnal/blob/main/docs/applications.md#firebase-cloud-messaging, which specifies the client-side model for the push:

```json
{
  "event_id": "$3957tyerfgewrf384",
  "type": "m.room.message",
  "sender": "@exampleuser:example.org",
  "room_name": "Mission Control",
  "room_alias": "#exampleroom:example.org",
  "sender_display_name": "Major Tom",
  "content_msgtype": "m.text",
  "content_body": "I'm floating in a most peculiar way.",
  "room_id": "!slw48wfj34rtnrf:example.org",
  "prio": "high",
  "unread": "2",
  "missed_calls": "1"
}
```

krille-chan:
@vanyasem Sygnal is just one push gateway from multiple possible gateways. The format of the data you receive from it is nowhere standartized but the server side model is. That's why we have implemented it here as it is also supported by [Hedwig](https://github.com/famedly/hedwig) and [FluffyGate](https://github.com/krille-chan/fluffygate). The purpose of a push gateway is to just forward the data to the app and that's why these two push gateways are doing.

Sygnal is just one possible use case and if you want to use it in your app you can just bring your own data model. I don't get what the blocker here is for you.

vanyasem:
@krille-chan I'm not sure what part of the phrase "**reference** push gateway" you don't understand, but if you insists on doing it your own incorrect way - so be it

krille-chan:
> @krille-chan I'm not sure what part of the phrase "**reference** push gateway" you don't understand, but if you insists on doing it your own incorrect way - so be it

Well a push gateway is highly bound to the app. That is app-specific implementation detail while this SDK provides matrix-specific things. Even with being a reference push gateway it does not mean, that this data format is an official standard.

PR #1952, opened 2024-11-07
Add require trailing comma lint

rmoradi88:


nico-famedly (approved):
Does what it says on the tin!

PR #1953, opened 2024-11-08
fix: BREAKING Cache members for encrypted rooms in database

krille-chan:
This fixes a huge performance
leak as for every encrypted room
the whole member list is loaded
every time you start the client.
To make sure we don't accidentally
cache 10000 users from a large
public room, I made this
an optional parameter. As the
parameters of this method are
now too large, I made them
named parameters instead of
positional ones. Unfortunately a
breaking change but makes it
more usable.


lib/src/room.dart
```diff
@@ -1560,14 +1560,16 @@ class Room {
...
-      bool suppressWarning = false,
-      bool cache = true]) async {
+  Future<List<User>> requestParticipants({
+    List<Membership> membershipFilter = const [
+      Membership.join,
+      Membership.invite,
+      Membership.knock,
+    ],
+    bool suppressWarning = false,
+    bool cache = true,
```
td-famedly, line 1570: if this is a breaking change, can you also be more verbose with the `cache` parameter? i.e change it to `cacheInMemory` for example?

krille-chan: Good idea. Have added it :)

nico-famedly (approved):
I think this looks good, although I really would like to see some tests for changes like this to ensure we actually prevent a second membership query and such for example. Tests are there to ensure we don't regress in behaviour that we want to keep, so imo it is always good to add them for something as significant as this :)

lib/src/room.dart
```diff
@@ -1600,9 +1605,18 @@ class Room {
...
-    if (cache) {
+    if (cacheInMemory) {
       for (final user in users) {
-        setState(user); // at *least* cache this in-memory
+        setState(user);
+      }
+    }
+    if (cacheInDatabase) {
+      for (final user in users) {
+        await client.database?.storeEventUpdate(
```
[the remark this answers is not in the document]
krille-chan: We already do as this is often called in a sync so it is in a batch there. If we put it into another batch we increase the problem with transaction-inside-transaction. Until we have found a better solution for this, we should try to avoid it

krille-chan:
> I think this looks good, although I really would like to see some tests for changes like this to ensure we actually prevent a second membership query and such for example. Tests are there to ensure we don't regress in behaviour that we want to keep, so imo it is always good to add them for something as significant as this :)

Added tests but also simplified it a little bit.

PR #1955, opened 2024-11-15
feat: Add more SyncStatus updates for init

krille-chan:
This makes it possible to track
what actually happens while
the Client runs the init() method.
The purpose is that clients can
give feedback to the user, what
the client is doing right now.
This is especially useful for
the login where we pipe the
Login, init and first sync method
behind each other.


lib/src/client.dart
```diff
@@ -3880,6 +3888,11 @@ class SyncStatusUpdate {
 }
 
 enum SyncStatus {
+  notLoggedIn,
+  initializing,
+  initializingEncryption,
+  initializingDatabaseMigration,
+  initializingFinalize,
```
td-famedly, line 3895: This could be more descriptive, as a comment or maybe even a better naming scheme.

lib/src/client.dart
```diff
@@ -3880,10 +3888,42 @@ class SyncStatusUpdate {
 }
 
 enum SyncStatus {
```
nico-famedly, line 3890: Another option we could use, is that we use a custom enum like this:

```dart
notLoggedIn(inSync: false, inInit: false, initStatus: null),
initializingStarted(inSync: false, inInit: true, initStatus: started),
initializingEncryption(inSync: false, inInit: true, initStatus: encryption),
...
waitingForResponse(inSync: true, inInit: false, syncStatus: waitingForResponse),
...
```

But in general you seem to have 3 separate states:

- The general app state: notLoggedIn, initializing, syncing
- The sync status: waitingForResponse, processing, cleanUp, finished, error
- The init status: start, encryption, dbMigration, finalize

So another option could be:

```
enum SyncStatus {
waitingForResponse, processing, cleanUp, finished, error
}
enum InitStatus {
start, encryption, dbMigration, finalize
}
enum ClientStatus {
  notLoggedIn(initStatus: null, syncStatus: null),
  initStart(initStatus: start, syncStatus: null),
  ...
  bool get initializing => initStatus != null;
  bool get syncing => syncStatus != null;
}
```

This is only a suggestion, but I think it would be easier to understand, if those separate values were actually separate. Then you can also easily check if the client is initialising or syncing and get more details if you want them. I also think the SyncStatus name isn't that descriptive anymore, once you start overloading it with the init and login state. But again, just a suggestion, your current solution also works fine.

krille-chan: hmm, or we just keep the SyncStatus as it is and add another stream `onInitStatus`? 🤔 This could make it even more non-complex

PR #1956, opened 2024-11-19
fix: BREAKING! missed initial updates for stream listener callbacks in P2P & mesh calls

coder-with-a-bushido:
I noticed that the SDK misses sending initial stream updates (in `onStreamAdd` and `onStreamRemove`) for both P2P and mesh calls in some cases. The easiest to reproduce would be an incoming P2P call where the `onStreamAdd` and `onStreamRemove` listeners are registered in `handleNewCall` VoIP method.

### Steps to reproduce
1. Start the flutter app in VS Code
2. Set a breakpoint within an if condition that check for the stream to not be local, inside the `onStreamAdd` callback
3. Start an outgoing call
4. Once the call is connected and the other person's media is shared, the debugger would stop at the breakpoint. This means that the `onStreamAdd` callback is working as expected.
5. Now, start the call from another device so that this device has an incoming call
6. Answer the call
7. Even after the call is connected, the debugger would NOT stop at the breakpoint. 

This is because, the remote stream is actually added even before the callback is registered inside the `handleNewCall` method.

Now, let's see where the remote stream is added in `CallSession`, in reverse order of invocation -

->  `_addRemoteStream()`
->  `_createPeerConnection()`
-> `_preparePeerConnection()`
-> `placeCallWithStreams()`          OR          -> `initWithInvite()`       OR              -> `initOutboundCall()`

And the top level method `initWithInvite()` is called in an incoming call before `delegate.handleNewCall` is called. The same for `initOutboundCall()` with outgoing call.

I thought of fixing this by adding a new intermediate WebRTCDelegate method where you can register callbacks before these methods are invoked. And that's what `registerListeners` will be used for.

For mesh calls, the same applies because it internally uses the same `CallSession`. So, I separated out the `onStreamAdd` and `onStreamRemove` callback registration and called it before the `placeCallWithStreams` method (in case a new member joins the call). For the case where you join a call, `onIncomingCallStart` stream controller in VoIP is used similar to the `handleNewCall`, so I created a new `onIncomingCallSetup` stream controller that's fired instead of the `registerListeners`. Then, use the `onIncomingCallSetup` callback to do the stream related callback registration.

### DIFF
--- a/lib/src/voip/backend/mesh_backend.dart
+++ b/lib/src/voip/backend/mesh_backend.dart
@@ -20,7 +20,11 @@ class MeshBackend extends CallBackend {
   /// participant:volume
   final Map<CallParticipant, double> _audioLevelsMap = {};
 
-  StreamSubscription<CallSession>? _callSubscription;
+  /// The stream is used to prepare for incoming peer calls like registering listeners
+  StreamSubscription<CallSession>? _callSetupSubscription;
+
+  /// The stream is used to signal the start of an incoming peer call
+  StreamSubscription<CallSession>? _callStartSubscription;
 
   Timer? _activeSpeakerLoopTimeout;
 
@@ -109,14 +113,32 @@ class MeshBackend extends CallBackend {
     );
   }
 
+  /// Register listeners for a peer call to use for the group calls, that is
+  /// needed before even call is added to `_callSessions`.
+  /// We do this here for onStreamAdd and onStreamRemoved to make sure we don't
+  /// miss any events that happen before the call is completely started.
+  void _registerListenersBeforeCallAdd(CallSession call) {
+    call.onStreamAdd.stream.listen((stream) {
+      if (!stream.isLocal()) {
+        onStreamAdd.add(stream);
+      }
+    });
+
+    call.onStreamRemoved.stream.listen((stream) {
+      if (!stream.isLocal()) {
+        onStreamRemoved.add(stream);
+      }
+    });
+  }
+
   Future<void> _addCall(GroupCallSession groupCall, CallSession call) async {
     _callSessions.add(call);
-    await _initCall(groupCall, call);
+    _initCall(groupCall, call);
     groupCall.onGroupCallEvent.add(GroupCallStateChange.callsChanged);
   }
 
   /// init a peer call from group calls.
-  Future<void> _initCall(GroupCallSession groupCall, CallSession call) async {
+  void _initCall(GroupCallSession groupCall, CallSession call) {
     if (call.remoteUserId == null) {
       throw MatrixSDKVoipException(
         'Cannot init call without proper invitee user and device Id',
@@ -141,18 +163,6 @@ class MeshBackend extends CallBackend {
     call.onCallHangupNotifierForGroupCalls.stream.listen((event) async {
       await _onCallHangup(groupCall, call);
     });
-
-    call.onStreamAdd.stream.listen((stream) {
-      if (!stream.isLocal()) {
-        onStreamAdd.add(stream);
-      }
-    });
-
-    call.onStreamRemoved.stream.listen((stream) {
-      if (!stream.isLocal()) {
-        onStreamRemoved.add(stream);
-      }
-    });
   }
 
   Future<void> _replaceCall(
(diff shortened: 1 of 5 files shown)


lib/src/voip/backend/mesh_backend.dart
```diff
@@ -20,7 +20,8 @@ class MeshBackend extends CallBackend {
   /// participant:volume
   final Map<CallParticipant, double> _audioLevelsMap = {};
 
-  StreamSubscription<CallSession>? _callSubscription;
+  StreamSubscription<CallSession>? _callSetupSubscription;
```
td-famedly, line 23: can you please add comments to what each of these streams is for?

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -658,7 +666,29 @@ class MeshBackend extends CallBackend {
     return;
   }
 
-  Future<void> _onIncomingCall(
+  void _onIncomingCallSetup(GroupCallSession groupCall, CallSession newCall) {
+    if (newCall.room.id != groupCall.room.id ||
```
td-famedly, line 670: This basically looks like `_onIncomingCallStart`, can you not merge them together and remove the register listener from the replaceCall as well?

td-famedly (commented):
need to test, but mostly lg! great catch :D

lib/src/voip/backend/mesh_backend.dart
coder-with-a-bushido: done

td-famedly: _callSetupSubscription and _callStartSubscription still seem to be missing a  comment

lib/src/voip/backend/mesh_backend.dart
td-famedly: hm oki

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -658,7 +666,29 @@ class MeshBackend extends CallBackend {
...
+        newCall.state != CallState.kRinging ||
```
td-famedly, line 671: Can you look at _onIncomingCallStart and add comments to all the conditionals and separate them out? It's easier to debug why calls are breaking then. Also make sure to namespace the comments so we know what function has printed them.

lib/src/voip/backend/mesh_backend.dart
coder-with-a-bushido: done

lib/src/voip/backend/mesh_backend.dart
coder-with-a-bushido: you mean logs and not comments right?

coder-with-a-bushido: if it's logs, then done!

rmoradi88:
@td-famedly Do you want to test it?

lib/src/voip/voip.dart
```diff
@@ -479,6 +485,12 @@ class VoIP {
     // by terminate.
     currentCID = VoipId(roomId: room.id, callId: callId);
 
+    if (confId == null) {
+      await delegate.registerListeners(newCall);
```
td-famedly, line 489: What happens if you always register the listeners at this point? Could you potentially remove the second registerListeners call below?

lib/src/voip/voip.dart
```diff
@@ -36,7 +36,13 @@ class VoIP {
   Map<VoipId, GroupCallSession> get groupCalls => _groupCalls;
   final Map<VoipId, GroupCallSession> _groupCalls = {};
 
-  final CachedStreamController<CallSession> onIncomingCall =
+  /// The stream is used to prepare for incoming peer calls in a mesh call
+  /// For example, registering listeners
```
td-famedly, line 40: if these are just used in mesh calls, maybe `onIncomingCallInMeshStart` and `onIncomingCallInMeshSetup` would be better? but i'll leave that decision completely up to you

td-famedly (commented):
apart from last nitpicks LG!

lib/src/voip/voip.dart
coder-with-a-bushido: Well, the other reference is used in a different function `inviteToCall` which gets called for incoming calls. And this usage is for `onCallInvite`.

td-famedly: ah github diff :D

td-famedly (approved):
lg, gw!

PR #1957, opened 2024-11-20
refactor: Do not set the deprecated dont_notify action in push rules

krille-chan:
- [x] Tested in FluffyChat

This has been deprecated since
https://spec.matrix.org/v1.7/client-server-api/#actions
An empty list of actions should now
be equal to a list with the entry
"dont_notify". Clients should
ignore this action and use
empty lists instead. A good
explaination can be found
https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3987-push-actions-clean-up.md

closes https://github.com/famedly/product-management/issues/2575

Should be backwards compatible. This also adds type safety in the push rule getter by using the already existing parser.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -2090,16 +2090,22 @@ class Room {
   /// Returns the [PushRuleState] for this room, based on the m.push_rules stored in
   /// the account_data.
   PushRuleState get pushRuleState {
-    final globalPushRules =
-        client.accountData['m.push_rules']?.content['global'];
-    if (globalPushRules is! Map) {
+    final globalPushRules = client.globalPushRules;
+    if (globalPushRules == null) {
+      // We have no push rules specified at all so we fallback to just notify:
       return PushRuleState.notify;
     }
 
-    if (globalPushRules['override'] is List) {
-      for (final pushRule in globalPushRules['override']) {
-        if (pushRule['rule_id'] == id) {
-          if (pushRule['actions'].indexOf('dont_notify') != -1) {
+    final overridePushRules = globalPushRules.override;
+    if (overridePushRules != null) {
+      for (final pushRule in overridePushRules) {
+        if (pushRule.ruleId == id) {
+          // "dont_notify" and "coalesce" should be ignored in actions since
+          // https://spec.matrix.org/v1.7/client-server-api/#actions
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -2090,27 +2090,38 @@ class Room {
   /// Returns the [PushRuleState] for this room, based on the m.push_rules stored in
   /// the account_data.
   PushRuleState get pushRuleState {
-    final globalPushRules =
-        client.accountData['m.push_rules']?.content['global'];
-    if (globalPushRules is! Map) {
+    final globalPushRules = client.globalPushRules;
+    if (globalPushRules == null) {
```
rmoradi88, line 2094: Just for my understanding could you explain why when the globalpushrules is null we return the notify state?

lib/src/room.dart
```diff
@@ -2090,27 +2090,38 @@ class Room {
   /// Returns the [PushRuleState] for this room, based on the m.push_rules stored in
```
rmoradi88, line 2090: Would you please update the comment as well :)

lib/src/room.dart
```diff
@@ -2090,27 +2090,38 @@ class Room {
...
-        if (pushRule['rule_id'] == id) {
-          if (pushRule['actions'].indexOf('dont_notify') != -1) {
+    final overridePushRules = globalPushRules.override;
+    if (overridePushRules != null) {
+      for (final pushRule in overridePushRules) {
+        if (pushRule.ruleId == id) {
+          // "dont_notify" and "coalesce" should be ignored in actions since
+          // https://spec.matrix.org/v1.7/client-server-api/#actions
+          pushRule.actions
+            ..remove('dont_notify')
```
rmoradi88, line 2105: You are removing this two times. Maybe as a cleaner approach just define a static variable and use it here and below instead of hardcoding two times. Just a personal prefrence.

lib/src/room.dart
krille-chan: Added a comment :)

lib/src/room.dart
krille-chan: To what? It still does exactly the same 🤔

rmoradi88 (approved):

PR #1958, opened 2024-11-20
feat: support filtering events when requesting events history or future

JohannesNevels:
This pull request introduces the StateFilter parameter to several methods within the codebase to allow for more granular control over the types of events being requested. This change enhances the flexibility of event requests, allowing for more specific event type filtering.

**Before:**
When using the requestHistory method, requesting a specific number of events (e.g., 10) could result in fewer events being returned, or none at all, even though there are more relevant chat events available. This happens because the default behavior checks all event types, not just the desired ones. In a chat application, when scrolling backwards or forwards, users expect to receive a consistent number of messages (e.g., 10), and only fewer when reaching the end of the available messages.

**After:**
By adding the StateFilter parameter, we can specify various criteria for the events to be received, such as event types (e.g., [EventTypes.Message]) and other state-related filters. This ensures that the methods return the expected number of relevant events, improving the user experience in scenarios like scrolling through chat messages. The StateFilter object will have lazyLoadMembers set to true by default, but this can be overridden.

Signed-off-by: Johannes Nevels <johannes.nevels@voys.nl>

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1261,18 +1261,26 @@ class Room {
         reason: reason,
       );
 
-  /// Request more previous events from the server. [historyCount] defines how much events should
+  /// Request more previous events from the server. [historyCount] defines how many events should
   /// be received maximum. When the request is answered, [onHistoryReceived] will be triggered **before**
-  /// the historical events will be published in the onEvent stream.
+  /// the historical events will be published in the onEvent stream. [filter] allows you to specify a
+  /// [StateFilter] object to filter the events, which can include various criteria such as event types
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -1282,7 +1284,8 @@ class Room {
       direction,
       from: prev_batch,
       limit: historyCount,
-      filter: jsonEncode(StateFilter(lazyLoadMembers: true).toJson()),
+      filter: jsonEncode(
```
td-famedly, line 1287: why did `lazyLoadMembers` change to false? Also I would just pass in the `Statefilter` as a parameter to the function instead of individual fields

PR #1959, opened 2024-11-26
feat: Add Client.onUploadingKeys stream to track uploading keys in the background

krille-chan:


krille-chan:
I'll try a different approach

PR #1961, opened 2024-11-26
fix: clear cache when clearing DB in MatrixSdkDatabase

coder-with-a-bushido:


lib/src/database/matrix_sdk_database.dart
```diff
@@ -338,7 +338,32 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
+    _userDeviceKeysBox.clearCache();
+    _userDeviceKeysOutdatedBox.clearCache();
+    _userCrossSigningKeysBox.clearCache();
+    _ssssCacheBox.clearCache();
+    _presencesBox.clearCache();
+    _timelineFragmentsBox.clearCache();
+    _eventsBox.clearCache();
+    _seenDeviceIdsBox.clearCache();
+    _seenDeviceKeysBox.clearCache();
+    _userProfilesBox.clearCache();
```
krille-chan, line 363: Can we move this into the Collection implementation? There we have a list of names `BoxCollection.boxNames` which we could use to iterate over them and then we cannot forget a box here

coder-with-a-bushido: I tried to do it but we create the Boxes here, so using the boxNames we can only create a new object that doesn't have this cache

coder-with-a-bushido: Do you think it's okay to do even more refactoring here?

lib/src/database/indexeddb_box.dart
```diff
@@ -213,20 +213,22 @@ class Box<V> {
     return;
   }
 
+  void clearCache() {
```
td-famedly, line 216: Would like a better name here so as to not collide with db.clearCache

lib/src/database/matrix_sdk_database.dart
td-famedly: This is similar to what clearCache does below, so I think this is probably fine

td-famedly: Overall I would like a refactor where we create, clear and delete boxes from a list of boxnames and not have to spell them out everytime, but we can look into that later ig

td-famedly (approved):
lgtm

PR #1962, opened 2024-12-02
feat: (BREAKING) Replace onMigration with advanced callback onInitStateChanged

krille-chan:
This makes it possible to track
the process of initialization in the
client by just enhancing what
the onMigration callback already
does.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -597,6 +597,7 @@ class Client extends MatrixApi {
     bool? refreshToken,
     AuthenticationData? auth,
     AccountKind? kind,
+    void Function(InitState)? onInitStateChanged,
   }) async {
     final response = await super.register(
       kind: kind,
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -3958,3 +3984,30 @@ class _EventPendingDecryption {
...
+  /// The database has been updated. A migration is in progress.
+  migratingDatabase,
+
+  /// The encryption module will be set up now. For the first login this also
+  /// includes uploading keys to the server.
+  settingUpEncryption,
+
+  /// The client is loading rooms, device keys and account data from the
+  /// database.
+  loadingData,
```
td-famedly, line 4001: we could make it more fancy and separate this out into loading rooms, device keys, etc

krille-chan: For this we already have the futures we can await. Also those futures are actually done in parallel and not one after another

td-famedly: Yes, but I would like to listen to one state and know what is going on instead of having to worry about the futures. Do you think this makes sense? 🤔

krille-chan: It does not work with the parallel stuff. We are awaiting the futures here but they are actually parallel

td-famedly: figured, thanks for confirming!

td-famedly (approved):
lg

PR #1965, opened 2024-12-05
refactor: Clarify Room.join() behavior and make sure DM link is purged if room not found

krille-chan:


lib/src/room.dart
```diff
@@ -1150,21 +1150,29 @@ class Room {
...
-      final dmId = directChatMatrixID;
-      if (dmId != null) {
-        await addToDirectChat(dmId);
-      }
+      if (dmId != null) await addToDirectChat(dmId);
 
       // now join
       await client.joinRoomById(id);
     } on MatrixException catch (exception) {
+      if (dmId != null) await removeFromDirectChat();
```
td-famedly, line 1167: Can we put this in the leave room call?

td-famedly: The spec doesn't say so, but it is intuitive to update the m.direct list if you leave the room right? Or does this have any edge cases?

krille-chan: We had this before and removed by a reason that in the archive we still want to see that this was once a DM room with this user.

td-famedly: ah right! thanks

td-famedly (approved):
lg, I don't see `M_NOT_FOUND` used for join events in the spec so this should be fine

Issue #1967, opened 2024-12-06
Rejecting a call doesn't send 'm.call.reject' event


coder-with-a-bushido:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [x] I already asked for help in the chat

### In which Project did the bug appear?

Famedly App

### On which platform did the bug appear?

Android, iOS, Chrome

### SDK Version

0.35.0

### Describe the problem caused by this bug

When a user calls `reject` from `CallSession` for an incoming call, it should end the call and ideally send an 'm.call.reject' event. But with the current implementation 'm.call.reject' event is not sent.

### Steps To Reproduce

1. Call from device A
2. Reject the call from device B
3. You should not get the 'm.call.reject' event, so the UI shows you that the call was not rejected but is actually a missed call

PR #1968, opened 2024-12-06
fix: rejecting a call doesn't send m.call.reject event

coder-with-a-bushido:
Closes https://github.com/famedly/matrix-dart-sdk/issues/1967

### DIFF
--- a/lib/src/voip/call_session.dart
+++ b/lib/src/voip/call_session.dart
@@ -937,7 +937,6 @@ class CallSession {
   /// This used to be done by calling hangup, but is a separate method and protocol
   /// event as of MSC2746.
   Future<void> reject({CallErrorCode? reason, bool shouldEmit = true}) async {
-    setCallState(CallState.kEnding);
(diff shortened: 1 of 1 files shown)


lib/src/voip/call_session.dart
```diff
@@ -946,6 +945,7 @@ class CallSession {
       return;
     }
     Logs().d('[VOIP] Rejecting call: $callId');
+    setCallState(CallState.kEnding);
```
td-famedly, line 948: the sending of the reject event is in `sendCallReject` though, are you sure the commit message makes sense?

coder-with-a-bushido: yep, the actual problem was `setCallState` was called right before checking the state

coder-with-a-bushido: The if condition was always true because the state will be `CallState.kEnding` and only hangup code will be executed

td-famedly: ah makes sense thanks!

td-famedly (approved):

Issue #1969, opened 2024-12-06
Matrix 1.11/MAC calculation – support for hkdf-hmac-sha256.v2


kputh:
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

[The specification of Matrix 1.6+ requires clients to support hkdf-hmac-sha256.v2 for MAC calculation](https://spec.matrix.org/v1.6/client-server-api/#mac-calculation). I examined the current (0.35.0) Matrix Dart SDK's source code – particularly [key_verification.dart, line 1241](https://github.com/famedly/matrix-dart-sdk/blob/c30f77cfb68bbc1ad3b18765284fc5962984d792/lib/encryption/utils/key_verification.dart#L1241) – and concluded that the SDK currently doesn't support hkdf-hmac-sha256.v2. (I apologize if I arrived at a wrong conclusion, and wasted your time.)

### Describe your ideal solution

As a software developer, I want the Matrix Dart SDK to fully support Matrix 1.11 so that I don't have to fork and modify this library. (The SDK has to support hkdf-hmac-sha256.v2 and prefer it over hkdf-hmac-sha256.)

### Version

0.35.0

### Security requirements

None

td-famedly:
msc: https://github.com/matrix-org/matrix-spec-proposals/pull/3783

kputh:
Fixing this issue requires a new binding for `olm_sas_calculate_mac_fixed_base64` in [olm](https://pub.dev/packages/olm). I couldn't create an issue for this, because the library's repository doesn't exist anymore, has been moved or is hidden from the public.

PR #1970, opened 2024-12-06
fix: use macv2 for correct base64 encoding

td-famedly:
Fixes: https://github.com/famedly/matrix-dart-sdk/issues/1969

### DIFF
--- a/lib/encryption/utils/key_verification.dart
+++ b/lib/encryption/utils/key_verification.dart
@@ -1238,7 +1238,10 @@ class QRCode {
 
 const knownKeyAgreementProtocols = ['curve25519-hkdf-sha256', 'curve25519'];
 const knownHashes = ['sha256'];
-const knownHashesAuthentificationCodes = ['hkdf-hmac-sha256'];
+const knownHashesAuthentificationCodes = [
+  'hkdf-hmac-sha256.v2',
+  'hkdf-hmac-sha256',
+];
 
 class _KeyVerificationMethodSas extends _KeyVerificationMethod {
   _KeyVerificationMethodSas({required super.request});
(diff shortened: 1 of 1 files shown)


lib/encryption/utils/key_verification.dart
```diff
@@ -1560,8 +1565,9 @@ class _KeyVerificationMethodSas extends _KeyVerificationMethod {
   }
 
   String _calculateMac(String input, String info) {
-    if (messageAuthenticationCode == 'hkdf-hmac-sha256') {
-      return sas!.calculate_mac(input, info);
+    if ({'hkdf-hmac-sha256', 'hkdf-hmac-sha256.v2'}
```
td-famedly, line 1568: This is probably wrong, we want to use the old function to ensure backwards compatibility

pubspec.yaml
```diff
@@ -43,3 +43,9 @@ dev_dependencies:
   path: ^1.9.1
   sqflite_common_ffi: ^2.3.4+4 # sqflite_common_ffi aggressively requires newer dart versions
   test: ^1.25.13
+
+dependency_overrides:
+  olm:
+    git:
+      url: https://github.com/famedly/dart-olm
+      ref: td/ffigen
```
krille-chan, line 51: Can we wait until this gets released before we merge this?

binghatch:
Hey @krille-chan, @td-famedly , do we have an update on this? It is also an gematik requirement ✌️

binghatch:
See here: https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_Basis/gemSpec_TI-M_Basis_V1.1.1/#A_26196

TI-M clients MUST support the MAC procedure hkdf-hmac-sha256 for device verification and additionally offer hkdf-hmac-sha256.v2 at the start of the verification 1. 1 [Client-Server API/#mac-calculation]

krille-chan (approved):

PR #1975, opened 2024-12-18
refactor: Use Event instead of EventUpdate for pending decryption event queue and for decrypt events in general

krille-chan:
This should removes an
unnecessary step of
json serialization and deserialization and should
therefore improve performance.
Gets rid of some unnecessary
code as well.

Related to https://github.com/famedly/product-management/issues/1875

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2561,16 +2561,16 @@ class Client extends MatrixApi {
       if (room != null) {
         final List<BasicEvent> events = [];
         for (final event in _eventsPendingDecryption) {
-          if (event.event.roomID != roomId) continue;
+          if (event.event.room.id != roomId) continue;
           if (!sessionIds.contains(
-            event.event.content['content']?['session_id'],
+            event.event.content.tryGet<String>('session_id'),
           )) {
             continue;
           }
 
-          final decryptedEvent = await event.event.decrypt(room);
-          if (decryptedEvent.content.tryGet<String>('type') !=
-              EventTypes.Encrypted) {
+          final decryptedEvent =
+              await encryption!.decryptRoomEvent(event.event);
+          if (decryptedEvent.type != EventTypes.Encrypted) {
             events.add(BasicEvent.fromJson(decryptedEvent.content));
           }
         }
@@ -2747,7 +2747,7 @@ class Client extends MatrixApi {
     // we collect them first before we handle them.
     final callEvents = <Event>[];
 
-    for (final event in events) {
+    for (var event in events) {
       // The client must ignore any new m.room.encryption event to prevent
       // man-in-the-middle attacks!
       if ((event.type == EventTypes.Encryption &&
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -2561,16 +2561,16 @@ class Client extends MatrixApi {
...
+            event.event.content.tryGet<String>('session_id'),
           )) {
             continue;
           }
 
-          final decryptedEvent = await event.event.decrypt(room);
-          if (decryptedEvent.content.tryGet<String>('type') !=
-              EventTypes.Encrypted) {
+          final decryptedEvent =
+              await encryption?.decryptRoomEvent(event.event);
```
coder-with-a-bushido, line 2572: should we wrap this into a try-catch block so it doesn't affect the for loop in case an exception occurs?

lib/src/client.dart
```diff
@@ -2760,25 +2760,30 @@ class Client extends MatrixApi {
...
       }
 
-      var update =
-          EventUpdate(roomID: room.id, type: type, content: event.toJson());
-      if (event.type == EventTypes.Encrypted && encryptionEnabled) {
-        update = await update.decrypt(room);
+      if (event is MatrixEvent &&
+          event.type == EventTypes.Encrypted &&
+          encryptionEnabled) {
+        final decrypted = await encryption?.decryptRoomEvent(
```
coder-with-a-bushido, line 2766: same here?

lib/src/client.dart
```diff
@@ -2760,25 +2760,30 @@ class Client extends MatrixApi {
...
+          Event.fromMatrixEvent(event, room),
+          updateType: type,
+        );
 
         // if the event failed to decrypt, add it to the queue
```
coder-with-a-bushido, line 2771: nitpick but this comment should be right above the else case

lib/src/client.dart
krille-chan: decryptRoomEvent() itself is now inside a big try-catch block :)

lib/src/client.dart
krille-chan: Done

lib/src/client.dart
krille-chan: Fixed :)

lib/src/client.dart
```diff
@@ -2561,16 +2561,16 @@ class Client extends MatrixApi {
...
+            event.event.content.tryGet<String>('session_id'),
           )) {
             continue;
           }
 
-          final decryptedEvent = await event.event.decrypt(room);
-          if (decryptedEvent.content.tryGet<String>('type') !=
-              EventTypes.Encrypted) {
+          final decryptedEvent =
+              await encryption!.decryptRoomEvent(event.event);
```
coder-with-a-bushido, line 2572: I have a doubt, do we use `EventUpdateType.timeline` for all to-device events?

krille-chan: How to you mean this? 🤔

coder-with-a-bushido: I was thinking if we should have a seperate enum value for to-device events. Like `EventUpdateType.toDevice` but handle it the same way as `timeline`. Just a nitpick though.

Not a blocker so I'll approve this one :)

coder-with-a-bushido: I just thought it would make it obvious that it's different from usual events when logging and makes code a little easy to understand

coder-with-a-bushido (approved):

lib/src/client.dart
krille-chan: > I was thinking if we should have a seperate enum value for to-device events. Like EventUpdateType.toDevice but handle it the same way as timeline. Just a nitpick though.

Ah okay. We don't handle to-device messages as EventUpdates. Only events which are tied to a room are handled here

PR #1977, opened 2024-12-18
build: Update dev dependencies and remove unused dependencies

krille-chan:


td-famedly:
This is unfortunately wrong, we need those for the autogen stuff, i'll fix it in a subsequent PR

PR #1981, opened 2024-12-19
chore: (BREAKING CHANGE) spec v1.13 autogen

td-famedly:


lib/matrix_api_lite/generated/api.dart
```diff
@@ -4648,7 +4693,7 @@ class Api {
     if (response.statusCode != 200) unexpectedResponse(response, responseBody);
     final responseString = utf8.decode(responseBody);
     final json = jsonDecode(responseString);
-    return ignore(json);
+    return json as Map<String, Object?>;
```
nico-famedly, line 4696: this is weird, I can't find that in the spec text.

nico-famedly (approved):
Not sure what this is breading, but apart from the support url changing types, this does not look that breaking.

lib/matrix_api_lite/generated/api.dart
krille-chan: The spec now returns an event_id like this:

```json
{
  "event_id": "$YUwRidLecu:example.com"
}
```

so returning json is at least not wrong but fetching the string, like we do with the sendMessage method would be better

td-famedly: Hm I can't seem to find it here either https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3roomsroomidreporteventid

td-famedly: Looks like reportContent got replaced to reportEvent and the autogen needed that input, I'll change this back to ignore then because I can't find the spec returning that

PR #1984, opened 2024-12-30
refactor: Do not handle ephemerals as EventUpdates

krille-chan:
This handles room ephemerals
directly and not as room
event updates, which saves
one unnecessary step to json serialize and
deserialize. Handling those
as room event updates had
no benefit anyway so this
should speed up performance.

### DIFF
--- a/test/client_test.dart
+++ b/test/client_test.dart
@@ -282,7 +282,7 @@ void main() {
 
       final eventUpdateList = await eventUpdateListFuture;
 
-      expect(eventUpdateList.length, 20);
+      expect(eventUpdateList.length, 18);
 
       expect(eventUpdateList[0].content['type'], 'm.room.member');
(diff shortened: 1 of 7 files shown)


test/client_test.dart
```diff
@@ -308,36 +308,28 @@ void main() {
       expect(eventUpdateList[5].roomID, '!726s6s6q:example.com');
       expect(eventUpdateList[5].type, EventUpdateType.timeline);
 
-      expect(eventUpdateList[6].content['type'], 'm.typing');
+      expect(eventUpdateList[6].content['type'], LatestReceiptState.eventType);
```
coder-with-a-bushido, line 311: can we somehow also check the content in `room.ephemerals` here?

coder-with-a-bushido (approved):
LGTM :)

test/client_test.dart
krille-chan: Good idea 👍💡

PR #1985, opened 2024-12-30
refactor: Do not unnecessarily serialize and deserialize json for every account data object

krille-chan:
This changes the database
api a little bit so that it does
not unnecessarily serialize
and deserialize all
account data objects. Should
improve the performance of
the SDK.

Please note that in the hive
database(s) we create a
copy of the json map before
storing it. That is not necessary
for the matrix sdk database
as there the json map gets
serialized to a String anyway
inside.


coder-with-a-bushido (approved):
LGTM :)

PR #1986, opened 2024-12-30
fix: Also load room account data in getSingleRoom()

krille-chan:
Looks like room account data
was just forgotten in this
method which explains the
bug in FluffyChat when
I receive a DM and in the
push notification it is not
marked as DM sometimes.


coder-with-a-bushido (approved):
LGTM :)

PR #1987, opened 2024-12-30
refactor: Remove deprecated Hive Database

krille-chan:
The Hive Database is
nowhere in use anymore and
as far as I can see not even
as a legacyDatabaseBuilder.


coder-with-a-bushido (approved):

lib/matrix.dart
```diff
@@ -25,7 +25,6 @@ export 'fake_matrix_api.dart' show FakeMatrixApi;
 
 export 'src/client.dart';
 export 'src/database/database_api.dart';
-export 'src/database/hive_database.dart';
```
coder-with-a-bushido, line 28: Since we're removing a DB option, shouldn't this be a breaking change even though it was deprecated beforehand?

krille-chan: Mentioned it in the commit message

PR #1988, opened 2024-12-31
refactor: Handle Room Account Data outside of Room Event Updates

krille-chan:
This also makes sure that
room account data does not
get unnecessarily serialized
and deserialized before
storing it in the database.
For this it changes the
code flow at multiple
places.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2559,7 +2559,7 @@ class Client extends MatrixApi {
 
       final room = getRoomById(roomId);
       if (room != null) {
-        final List<BasicEvent> events = [];
+        final events = <Event>[];
         for (final event in _eventsPendingDecryption) {
           if (event.event.room.id != roomId) continue;
           if (!sessionIds.contains(
@@ -2571,7 +2571,7 @@ class Client extends MatrixApi {
           final decryptedEvent =
               await encryption!.decryptRoomEvent(event.event);
           if (decryptedEvent.type != EventTypes.Encrypted) {
-            events.add(BasicEvent.fromJson(decryptedEvent.content));
+            events.add(decryptedEvent);
           }
         }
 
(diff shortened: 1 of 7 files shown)


lib/src/database/hive_collections_database.dart
```diff
@@ -1095,6 +1095,15 @@ class HiveCollectionsDatabase extends DatabaseApi {
     return;
   }
 
+  @override
+  Future<void> storeRoomAccountData(BasicRoomEvent event) async {
+    await _roomAccountDataBox.put(
+      TupleKey(event.roomId ?? '', event.type).toString(),
+      copyMap(event.toJson()),
```
coder-with-a-bushido, line 1102: if I'm not wrong, we only need to use `copyMap` when we get something from DB and not to store

lib/src/client.dart
```diff
@@ -2671,12 +2670,10 @@ class Client extends MatrixApi {
...
         final accountData = syncRoomUpdate.accountData;
         if (accountData != null && accountData.isNotEmpty) {
-          await _handleRoomEvents(
-            room,
-            accountData,
-            EventUpdateType.accountData,
-            store: false,
-          );
+          for (final event in accountData) {
+            await database?.storeRoomAccountData(event);
```
coder-with-a-bushido, line 2674: I think we don't need to store it in DB here because it is `LeftRoomUpdate`. Even the old code has `store` set to `false`

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1058,6 +1058,15 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     return;
   }
 
+  @override
+  Future<void> storeRoomAccountData(BasicRoomEvent event) async {
+    await _roomAccountDataBox.put(
+      TupleKey(event.roomId ?? '', event.type).toString(),
+      copyMap(event.toJson()),
```
coder-with-a-bushido, line 1065: same here?

lib/src/database/hive_collections_database.dart
krille-chan: Ah you are right. Thanks

lib/src/client.dart
krille-chan: Thanks, you are right again :D

lib/src/database/hive_collections_database.dart
krille-chan: Okay actually only in the Matrix SDK database. In Hive I think we still need to create a copy before storing it. At least in the other parts we are still doing this

coder-with-a-bushido (approved):
LGTM :)

PR #1992, opened 2025-01-02
refactor: Use Event instead of EventUpdate for storing in db

krille-chan:
Fixes https://github.com/famedly/product-management/issues/1875

### DIFF
--- a/lib/encryption/encryption.dart
+++ b/lib/encryption/encryption.dart
@@ -348,11 +348,9 @@ class Encryption {
           event.room.setState(event);
         }
         await client.database?.storeEventUpdate(
-          EventUpdate(
-            content: event.toJson(),
-            roomID: event.room.id,
-            type: updateType,
-          ),
+          event.room.id,
+          event,
+          updateType,
           client,
         );
       }
--- a/lib/encryption/encryption.dart
+++ b/lib/encryption/encryption.dart
@@ -154,20 +154,20 @@ class Encryption {
     }
   }
 
-  Future<void> handleEventUpdate(EventUpdate update) async {
-    if (update.type == EventUpdateType.history) {
+  Future<void> handleEventUpdate(Event event, EventUpdateType type) async {
+    if (type == EventUpdateType.history) {
       return;
     }
-    if (update.content['type'].startsWith('m.key.verification.') ||
-        (update.content['type'] == EventTypes.Message &&
-            (update.content['content']['msgtype'] is String) &&
-            update.content['content']['msgtype']
-                .startsWith('m.key.verification.'))) {
+    if (event.type.startsWith('m.key.verification.') ||
+        (event.type == EventTypes.Message &&
+            event.content
+                    .tryGet<String>('msgtype')
+                    ?.startsWith('m.key.verification.') ==
+                true)) {
       // "just" key verification, no need to do this in sync
-      runInRoot(() => keyVerificationManager.handleEventUpdate(update));
+      runInRoot(() => keyVerificationManager.handleEventUpdate(event));
     }
-    if (update.content['sender'] == client.userID &&
-        update.content['unsigned']?['transaction_id'] == null) {
+    if (event.senderId == client.userID && event.status.isSynced) {
       // maybe we need to re-try SSSS secrets
       runInRoot(() => ssss.periodicallyRequestMissingCache());
     }
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2801,8 +2801,11 @@ class Client extends MatrixApi {
       if (store) {
         await database?.storeEventUpdate(update, this);
       }
-      if (encryptionEnabled) {
-        await encryption?.handleEventUpdate(update);
+      if (event is MatrixEvent && encryptionEnabled) {
+        await encryption?.handleEventUpdate(
+          Event.fromMatrixEvent(event, room),
+          type,
+        );
       }
       onEvent.add(update);
 
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1647,7 +1647,27 @@ class Client extends MatrixApi {
   /// the app receives a new synchronization, this event is called for every signal
   /// to update the GUI. For example, for a new message, it is called:
   /// onRoomEvent( "m.room.message", "!chat_id:server.com", "timeline", {sender: "@bob:server.com", body: "Hello world"} )
+  // ignore: deprecated_member_use_from_same_package
+  @Deprecated(
+    'Use `onTimelineEvent`, `onHistoryEvent` or `onNotification` instead.',
+  )
   final CachedStreamController<EventUpdate> onEvent = CachedStreamController();
+
+  /// A stream of all incoming timeline events for all rooms **after**
+  /// decryption. The events are coming in the same order as they come down from
+  /// the sync.
+  final CachedStreamController<Event> onTimelineEvent =
+      CachedStreamController();
+
+  /// A stream for all incoming historical timeline events **after** decryption
+  /// triggered by a `Room.requestHistory()` call or a method which calls it.
+  final CachedStreamController<Event> onHistoryEvent = CachedStreamController();
+
+  /// A stream of incoming Events **after** decryption which **should** trigger
+  /// a (local) notification. This includes timeline events but also
+  /// invite states. Excluded events are those sent by the user themself or
+  /// not matching the push rules.
+  final CachedStreamController<Event> onNotification = CachedStreamController();
 
   /// The onToDeviceEvent is called when there comes a new to device event. It is
   /// already decrypted if necessary.
@@ -1889,11 +1909,9 @@ class Client extends MatrixApi {
     if (storeInDatabase) {
       await database?.transaction(() async {
         await database.storeEventUpdate(
-          EventUpdate(
-            roomID: roomId,
-            type: EventUpdateType.timeline,
-            content: event.toJson(),
-          ),
+          roomId,
+          event,
+          EventUpdateType.timeline,
           this,
         );
       });
@@ -2768,12 +2786,6 @@ class Client extends MatrixApi {
         }
       }
 
-      final update = EventUpdate(
-        roomID: room.id,
-        type: type,
-        content: event.toJson(),
-      );
-
       // Any kind of member change? We should invalidate the profile then:
       if (event.type == EventTypes.RoomMember) {
         final userId = event.stateKey;
@@ -2799,7 +2811,7 @@ class Client extends MatrixApi {
       }
       _updateRoomsByEventUpdate(room, event, type);
       if (store) {
-        await database?.storeEventUpdate(update, this);
+        await database?.storeEventUpdate(room.id, event, type, this);
       }
       if (event is MatrixEvent && encryptionEnabled) {
         await encryption?.handleEventUpdate(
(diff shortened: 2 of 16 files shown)


lib/encryption/encryption.dart
```diff
@@ -348,11 +348,9 @@ class Encryption {
...
         }
         await client.database?.storeEventUpdate(
-          EventUpdate(
-            content: event.toJson(),
-            roomID: event.room.id,
-            type: updateType,
-          ),
+          event.room.id,
+          event,
+          EventUpdateType.state,
```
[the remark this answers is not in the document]
krille-chan: Ah you are right. Fixed it :)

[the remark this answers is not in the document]
coder-with-a-bushido: I think the argument is `type` and you just used `updateType`

lib/encryption/encryption.dart
```diff
@@ -348,11 +348,9 @@ class Encryption {
...
         }
         await client.database?.storeEventUpdate(
-          EventUpdate(
-            content: event.toJson(),
-            roomID: event.room.id,
-            type: updateType,
-          ),
+          event.room.id,
+          event,
+          updateType,
```
coder-with-a-bushido, line 353:
```suggestion
          type,
```

lib/src/client.dart
```diff
@@ -2797,23 +2796,41 @@ class Client extends MatrixApi {
...
       }
-      _updateRoomsByEventUpdate(room, update);
+      _updateRoomsByEventUpdate(room, event, type);
       if (store) {
-        await database?.storeEventUpdate(update, this);
+        await database?.storeEventUpdate(room.id, event, type, this);
       }
-      if (encryptionEnabled) {
-        await encryption?.handleEventUpdate(update);
+      if (event is MatrixEvent && encryptionEnabled) {
```
coder-with-a-bushido, line 2803: I checked that `event is MatrixEvent` is false only in the case of `EventUpdateType.inviteState`. I think it could break key verification for invited rooms? I'll have to research more to confirm but I think this is not a required change for this refactoring. Can we pls remove just the `event is MatrixEvent` condition here?

coder-with-a-bushido: It would mean both `Encryption.handleEventUpdate` and `KeyVerificationManager.handleEventUpdate` should accept `StrippedStateEvent` instead of `Event` 😞

coder-with-a-bushido: AH nevermind, key verification can happen only after joining a room.

lib/src/client.dart
```diff
@@ -2797,23 +2796,41 @@ class Client extends MatrixApi {
...
+      // ignore: deprecated_member_use_from_same_package
+      onEvent.add(
+        // ignore: deprecated_member_use_from_same_package
+        EventUpdate(
+          roomID: room.id,
+          type: type,
+          content: event.toJson(),
+        ),
+      );
+      if (event is MatrixEvent) {
```
coder-with-a-bushido, line 2819: should we have an `onInviteStateEvent` stream too? It's the only case where the event is not `MatrixEvent` but a `StrippedStateEvent`

lib/encryption/encryption.dart
coder-with-a-bushido: my bad

lib/src/client.dart
krille-chan: No because this would be a duplication of `onSync` where you can just filter. This is for events **after** potential decryption

lib/src/client.dart
```diff
@@ -2913,23 +2930,34 @@ class Client extends MatrixApi {
...
+        room.setState(eventUpdate);
         break;
       case EventUpdateType.state:
       case EventUpdateType.timeline:
-        final event = Event.fromJson(eventUpdate.content, room);
+        if (eventUpdate is! MatrixEvent) {
+          Logs().wtf(
+            'Passed in a ${eventUpdate.runtimeType} with $type to _updateRoomsByEventUpdate(). This should never happen!',
+          );
+          assert(true);
```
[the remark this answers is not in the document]
krille-chan: oh you are right. Thank you so much. I changed it to the same condition as before to improve the error message there

lib/src/client.dart
coder-with-a-bushido: Ok, but we use `onEvent` for local notifications. If there's no `onInviteStateEvent` then we would miss `inviteState` updates to show local notifications.

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1068,45 +1068,37 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
-      final event =
+    if (event.type == EventTypes.Redaction && event is MatrixEvent) {
+      final redactionEvent = Event.fromMatrixEvent(event, tmpRoom);
+      final eventId = redactionEvent.redacts;
+      final redactedEvent =
           eventId != null ? await getEventById(eventId, tmpRoom) : null;
-      if (event != null) {
-        event.setRedactionEvent(Event.fromJson(eventUpdate.content, tmpRoom));
+      if (redactedEvent != null) {
+        redactedEvent.setRedactionEvent(redactionEvent);
```
coder-with-a-bushido, line 1087:
```suggestion
        redactedEvent.setRedactionEvent(event);
```

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1068,45 +1068,37 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
-              TupleKey(eventUpdate.roomID, event.type, '').toString(),
-              event.toJson(),
-            );
-          } else {
-            await _nonPreloadRoomStateBox.put(
-              TupleKey(eventUpdate.roomID, event.type, '').toString(),
-              event.toJson(),
-            );
-          }
-        }
```
coder-with-a-bushido, line 1099: Just confirming. We removed it because it's already handled below right?

```
    // Store a common state event
    if (stateKey != null &&
        // Don't store events as state updates when paginating backwards.
        (eventUpdate.type == EventUpdateType.timeline ||
            eventUpdate.type == EventUpdateType.state ||
            eventUpdate.type == EventUpdateType.inviteState)) {
      if (eventUpdate.content['type'] == EventTypes.RoomMember) {
        await _roomMembersBox.put(
          TupleKey(
            eventUpdate.roomID,
            eventUpdate.content['state_key'],
          ).toString(),
          eventUpdate.content,
        );
      } else {
        final type = eventUpdate.content['type'] as String;
        final roomStateBox = client.importantStateEvents.contains(type)
            ? _preloadRoomStateBox
            : _nonPreloadRoomStateBox;
        final key = TupleKey(
          eventUpdate.roomID,
          type,
          stateKey,
        ).toString();

        await roomStateBox.put(key, eventUpdate.content);
      }
    }
```

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1190,37 +1172,36 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
     // Store a common state event
     if (stateKey != null &&
         // Don't store events as state updates when paginating backwards.
-        (eventUpdate.type == EventUpdateType.timeline ||
-            eventUpdate.type == EventUpdateType.state ||
-            eventUpdate.type == EventUpdateType.inviteState)) {
-      if (eventUpdate.content['type'] == EventTypes.RoomMember) {
+        (type == EventUpdateType.timeline ||
+            type == EventUpdateType.state ||
+            type == EventUpdateType.inviteState)) {
```
coder-with-a-bushido, line 1185:
```suggestion
        {EventUpdateType.timeline, EventUpdateType.state, EventUpdateType.inviteState}.contains(type)) {
```

lib/src/timeline.dart
```diff
@@ -512,43 +526,35 @@ class Timeline {
...
-      );
+      final status = event.status;
 
       final i = _findEvent(
-        event_id: eventUpdate.content['event_id'],
-        unsigned_txid: eventUpdate.content['unsigned'] is Map
-            ? eventUpdate.content['unsigned']['transaction_id']
-            : null,
+        event_id: event.eventId,
+        unsigned_txid: event.unsigned?.tryGet<String>('transaction_id'),
```
coder-with-a-bushido, line 551: I think there's enough usage of `event.unsigned?.tryGet<String>('transaction_id')` to create a getter `txid` in the `Event` class. Should we create one? I can open a separate PR for that if you agree.

test/client_test.dart
```diff
@@ -278,43 +278,20 @@ void main() {
...
       final eventUpdateList = await eventUpdateListFuture;
 
-      expect(eventUpdateList.length, 13);
+      expect(eventUpdateList.length, 3);
 
-      expect(eventUpdateList[0].content['type'], 'm.room.member');
-      expect(eventUpdateList[0].roomID, '!726s6s6q:example.com');
-      expect(eventUpdateList[0].type, EventUpdateType.state);
+      expect(eventUpdateList[0].type, 'm.room.member');
+      expect(eventUpdateList[0].roomId, '!726s6s6q:example.com');
```
coder-with-a-bushido, line 288: I think we could also have a `event.type` check here for all these 3 cases (just like before)

lib/src/client.dart
krille-chan: That's a good point 🤔 You can use just listen to `Client.onSync` and filter what you need.

Question is, would it be handy to add an `Client.onInvite` stream to listen to or would it be overhead?

lib/src/database/matrix_sdk_database.dart
krille-chan: Correct :)

lib/src/timeline.dart
krille-chan: SGTM :)

test/client_test.dart
krille-chan: Wouldn't this be `expect(eventUpdateList[0].type, 'm.room.member');`? We are already checking the type here

lib/src/database/matrix_sdk_database.dart
coder-with-a-bushido: Hmm, it's either that or the other way around. Definitely not whats there right now

lib/src/client.dart
coder-with-a-bushido: I think it makes sense to have `onInviteStateEvent` since users could have a separate logic for handling invited room events. Even the event type is `StrippedStateEvent` instead of `MatrixEvent`.

1. `onEvent` was used for inviteState events and it's easier to migrate to `onInviteStateEvent` for users than to use `client.onSync` for only that scenario.
2. If we end up not having a separate stream for it, we should make up for it in the documentation and migration guide.

lib/src/database/matrix_sdk_database.dart
coder-with-a-bushido: my bad, sorry!

coder-with-a-bushido (approved):
LGTM :)

lib/src/client.dart
krille-chan: okay changed it accordingly :)

lib/src/client.dart
```diff
@@ -2797,23 +2809,70 @@ class Client extends MatrixApi {
...
+                timelineEvent.senderId != userID &&
+                room.notificationCount > 0 &&
+                pushruleEvaluator.match(timelineEvent).notify) {
+              onNotification.add(timelineEvent);
+            }
+            break;
+          case EventUpdateType.history:
+            onHistoryEvent.add(timelineEvent);
+            break;
+          default:
```
coder-with-a-bushido, line 2847: I think `onNotification.add` should be called for `EventUpdateType.state` as well

lib/src/client.dart
```diff
@@ -2797,23 +2809,70 @@ class Client extends MatrixApi {
...
+        ),
+      );
+      if (event is MatrixEvent) {
+        final timelineEvent = Event.fromMatrixEvent(event, room);
+        switch (type) {
+          case EventUpdateType.timeline:
+            onTimelineEvent.add(timelineEvent);
+            if (prevBatch != null &&
+                timelineEvent.senderId != userID &&
+                room.notificationCount > 0 &&
```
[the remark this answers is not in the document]
krille-chan: to determine if we actually have a notification here. This is the only way (that I'm currently aware of) to ensure that we do not display notifications for already seen events when doing an initial sync for example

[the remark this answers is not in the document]
krille-chan: We are doing the same in FluffyChat and in the Famedly App right now

lib/src/client.dart
krille-chan: onNotification should only display notifications in the same way we receive notifications from push. State events are not triggering notifications

PR #1993, opened 2025-01-03
chore: Trigger coverage timeout

krille-chan:


krille-chan:
Okay so after trial and error I am confident to say that `test/encryption/key_request_test.dart` and `test/encryption/key_manager_test.dart` are causing the bug! Deleting **both** files make the bug disappear

coder-with-a-bushido (approved):

PR #1996, opened 2025-01-07
refactor: Push Notification helper class make all fields optional and migrate dynamics to Object?

krille-chan:


coder-with-a-bushido (approved):

coder-with-a-bushido (changes requested):
this is a breaking change right? can we pls have the commit say it?

coder-with-a-bushido (approved):

PR #1999, opened 2025-01-09
refactor: Remove BasicRoomEvent type

krille-chan:
We actually never use the
roomId of this type and the
matrix spec has changed in
a way that the roomId is
never sent there actually. So
it was super easy to just
replace all BasicRoomEvent
with BasicEvent. The roomId
became nullable anyway.


coder-with-a-bushido (commented):
LGTM but CI is failing for some reason

td-famedly (approved):
weird that we never used `roomId` anywhere

PR #2005, opened 2025-01-20
refactor: Make direct chat getter type safe

krille-chan:
This also makes sure that we
do not accidentally change
it in RAM before the
change comes from the server
when calling
addToDirectChat()


lib/src/room.dart
```diff
@@ -1372,21 +1372,21 @@ class Room {
...
+      return;
     }
 
+    dmRooms.add(id);
+
     await client.setAccountData(
       client.userID!,
       'm.direct',
-      directChats,
+      {
```
coder-with-a-bushido, line 1386: I think the 'markasdm' in commands_test.dart needs to be changed to account for this

krille-chan: After a longer investigation I found the reason why `alice` was no longer in the called endpoints JSON:

In a test before we tested the `/dm` command with alice user ID so we already called this endpoint before and added Alice as a DM user. So she should be in the JSON, right?

No! We shouldn't because we called the endpoint in the fake matrix API which never returned it via sync. In current main she still gets added to the local account data in memory because we have this bug that we did not copied the list before editing it. So we changed it in memory before sending the update to the server. So the whole test case was wrong. Just removing this is therefore correct.

krille-chan: So this PR fixes the following bug:

When marking a room as DM room and this fails because of network connection, the room is nevertheless marked as DM room locally because we edited the account data locally accidentally. This is unintended behavior.

coder-with-a-bushido: Thank you so much :)

PR #2008, opened 2025-01-29
refactor: Add missing versions to supportedVersions and use warning instead of blocking

krille-chan:


lib/src/client.dart
```diff
@@ -1634,7 +1634,7 @@ class Client extends MatrixApi {
     return pushrules != null ? TryGetPushRule.tryFromJson(pushrules) : null;
   }
 
-  static const Set<String> supportedVersions = {'v1.1', 'v1.2'};
+  static final Set<RegExp> supportedVersions = {RegExp(r'^v1\..*$')};
```
td-famedly, line 1637: eh, I don't get this, what if v1.14 has a new endpoint which we probably won't support for a while?

nico-famedly, line 1637: Matrix 1.14 is set to remove the old media endpoints. Until recently we didn't support that. This regex would have connected to that homeserver and then have been incompatible. So imo we should just update that version list when we update the generated API. Seems like that was forgotten, but these version checks validate against specific versions for a reason. Every Matrix spec release is allowed to have breaking changes (and they do), it does not follow semver.

krille-chan: I don't see the problem here. Why should the sdk not connect to the server then? Yes, media would not work anymore but that's not a reason to block the whole app from connecting. The check here is to make sure we can do basic stuff with the server like login and sync. We cannot do this with matrix v2.x right now as we don't have sliding sync or OIDC yet. So if a server only supports v2.x it makes sense to block connecting to the server as it would break anyway. But for everything else why should we block the user at this point? It never helped us so far and only produced problems

nico-famedly: Well, the old sync method might be dropped in 1.18. Nothing in the spec guarantees you semantic versioning, so I don't understand where you v2.x is coming from. The sync endpoint could be removed in a 1.XX version as well! Or the old login endpoint! This version check should really be about checking if one of the API versions we support is supported by the server. Otherwise we can just remove it entirely, I think?

coder-with-a-bushido (approved):
LGTM except that one comment :)

lib/src/client.dart
```diff
@@ -542,10 +542,10 @@ class Client extends MatrixApi {
       final versions = await getVersions();
       if (!versions.versions
           .any((version) => supportedVersions.contains(version))) {
-        throw BadServerVersionsException(
-          versions.versions.toSet(),
-          supportedVersions,
+        Logs().w(
+          'Server supports the versions: ${versions.toString()} but this application is only compatible with ${supportedVersions.toString()}.',
         );
+        assert(true);
```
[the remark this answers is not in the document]
krille-chan: ah thank you so much

PR #2011, opened 2025-02-02
Handle call errors

coder-with-a-bushido:
Fixes https://famedly-ip.sentry.io/issues/6207590176/?project=1448986&query=is%3Aunresolved%20null%20check%20operator&referrer=issue-stream&statsPeriod=7d&stream_index=3

### DIFF
--- a/lib/src/voip/call_session.dart
+++ b/lib/src/voip/call_session.dart
@@ -1209,8 +1209,23 @@ class CallSession {
           }
         }
       };
-    } catch (e) {
-      Logs().v('[VOIP] prepareMediaStream error => $e');
+    } catch (e, s) {
+      Logs().e(
+        '[VOIP] preparePeerConnection => Failed to create peer connection object',
+        e,
+        s,
+      );
+      fireCallEvent(CallStateChange.kError);
+      await terminate(
+        CallParty.kLocal,
+        CallErrorCode.createPeerConnectionFailed,
+        true,
+      );
+      throw CallError(
+        CallErrorCode.createPeerConnectionFailed,
+        'Failed to create peer connection object ',
+        e,
+      );
     }
   }
 
--- a/lib/src/voip/call_session.dart
+++ b/lib/src/voip/call_session.dart
@@ -458,9 +458,14 @@ class CallSession {
         await pc!.setLocalDescription(answer);
       }
     } catch (e, s) {
-      Logs().e('[VOIP] onNegotiateReceived => ', e, s);
-      await _getLocalOfferFailed(e);
-      return;
+      Logs().e('[VOIP] onNegotiateReceived => Failed to get local offer', e, s);
+      fireCallEvent(CallStateChange.kError);
+      await terminate(CallParty.kLocal, CallErrorCode.localOfferFailed, true);
+      throw CallError(
+        CallErrorCode.localOfferFailed,
+        'Failed to get local offer',
+        e,
+      );
     }
 
     final newLocalOnHold = await isLocalOnHold();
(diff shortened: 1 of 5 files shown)


lib/src/voip/call_session.dart
```diff
@@ -207,6 +207,9 @@ class CallSession {
   // outgoing call
   Future<void> initOutboundCall(CallType type) async {
     await _preparePeerConnection();
+    if (pc == null) {
```
td-famedly, line 210: shouldn't `_preparePeerConnection` just rethrow? i.e can you even do anything without a pc?

coder-with-a-bushido: Yes you can't do anything without a pc. That's why it's terminating the call with an error code. Do check the createPeerConnection function

td-famedly: so instead of duplicating the early returns, couldn't we just do it in _preparePeerConnection when createPeerConnection throws and then rethrow?

coder-with-a-bushido: I'm okay with that if this was only about rethrowing the error. 
We are updating `onCallEventChanged` with `CallStateChange.kError` and terminating the call with correct reason. I think what we're doing here right now is very consistent with the existing calls code and only rethrowing the error for just this one doesn't make sense.

td-famedly: But you could also move the call event changed to error logic just in `_preparePeerConnection` right? Am I missing something here?

coder-with-a-bushido: Umm, I think that's what we do here already. Can you pls recheck?

td-famedly (commented):
lg, do you maybe want to add a test for it? I think we could add a flag to `MockWebRTCDelegate` which tells it to throw an error on `createPeerConnection` and then we can confirm if the call is terminated as expected. 

(feel free to modify the suggestion if you have better ideas!)

lib/src/voip/call_session.dart
```diff
@@ -1218,7 +1217,9 @@ class CallSession {
         }
       };
     } catch (e) {
-      Logs().v('[VOIP] prepareMediaStream error => ${e.toString()}');
+      Logs().v('[VOIP] preparePeerConnection error => ${e.toString()}');
+      await _createPeerConnectionFailed(e);
+      rethrow;
```
td-famedly, line 1222: should this rethrow here?

lib/src/voip/call_session.dart
```diff
@@ -592,7 +592,6 @@ class CallSession {
         await addLocalStream(stream, SDPStreamMetadataPurpose.Screenshare);
         return true;
       } catch (err) {
```
td-famedly, line 594: quick question, why was this removed?

coder-with-a-bushido: because it's already handled in `_getDisplayMedia();`

lib/src/voip/call_session.dart
coder-with-a-bushido: sorry, forgot to remove it

lib/src/voip/call_session.dart
```diff
@@ -1313,16 +1314,16 @@ class CallSession {
...
+    return null;
   }
 
   Future<MediaStream?> _getDisplayMedia() async {
     try {
       return await voip.delegate.mediaDevices
           .getDisplayMedia(UserMediaConstraints.screenMediaConstraints);
     } catch (e) {
-      await _getUserMediaFailed(e);
+      await _getDisplayMediaFailed(e);
```
td-famedly, line 1326: hm this is a bit difficult to read, I don't think `_getDisplayMediaFailed` is reused, can we maybe remove the `*failed` methods and move the implementations in this block? it saves us the code jumping + might also remove the `return null`

td-famedly (approved):
lg

Issue #2014, opened 2025-02-03
PushNotifications.fromJson fails without devices


konrad-xtention:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### On which platform did the bug appear?

Android

### SDK Version

0.37.0

### Describe the problem caused by this bug

In 0.37.0, [all fields of `PushNotification` were made optional](https://github.com/famedly/matrix-dart-sdk/pull/1996). But despite this change, `PushNotifications.fromJson` fails when no devices are present.

It should be possible to use `fromJson` on Json from `toJson`, but this throws an exception:

`final PushNotification notification = PushNotification.fromJson(PushNotification().toJson());`

### Steps To Reproduce

Minimal reproduction:

```
final PushNotification notification = PushNotification.fromJson({});
```

Expected result: Get a `PushNotification` with all fields set to null.
Actual result: An exception is thrown.

### Screenshots or Logs

type 'Null' is not a subtype of type 'String' in type cast

PR #2015, opened 2025-02-03
fix: PushNotification fromJson - toJson fails

krille-chan:
closes https://github.com/famedly/matrix-dart-sdk/issues/2014


mErrenst (approved):
LGTM

PR #2018, opened 2025-02-04
feat: (BREAKING) Make share keys with logic configurable

krille-chan:
closes https://github.com/famedly/product-management/issues/2028

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -90,7 +90,7 @@ class Client extends MatrixApi {
 
   final bool mxidLocalPartFallback;
 
-  bool shareKeysWithUnverifiedDevices;
+  ShareKeysWith shareKeysWith;
 
   Future<void> Function(Client client)? onSoftLogout;
 
@@ -219,7 +219,7 @@ class Client extends MatrixApi {
     Duration defaultNetworkRequestTimeout = const Duration(seconds: 35),
     this.sendTimelineEventTimeout = const Duration(minutes: 1),
     this.customImageResizer,
-    this.shareKeysWithUnverifiedDevices = true,
+    this.shareKeysWith = ShareKeysWith.crossVerifiedIfEnabled,
     this.enableDehydratedDevices = false,
     this.receiptsPublicByDefault = true,
 
@@ -4081,4 +4081,26 @@ enum InitState {
 
   /// Initialization has been completed with an error.
   error,
+}
+
+/// Sets the security level with which devices keys should be shared with
+enum ShareKeysWith {
+  /// Keys are shared with all devices if they are not explicitely blocked
+  all,
+
+  /// Once a user has enabled cross signing, keys are no longer shared with
+  /// devices which are not cross verified by the cross signing keys of this
+  /// user. This does not require that the user needs to be verified.
+  crossVerifiedIfEnabled,
+
+  /// Keys are only shared with cross verified devices. If a user has not
+  /// enabled cross signing, then all devices must be verified manually first.
+  /// This does not require that the user needs to be verified.
+  crossVerified,
+
+  /// Keys are only shared with direct verified devices. So either the device
+  /// or the user must be manually verified first, before keys are shared. By
+  /// using cross signing, it is enough to verify the user and then the user
+  /// can verify their devices.
+  directlyVerifiedOnly,
 }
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -4082,3 +4082,21 @@ enum InitState {
...
 }
+
+/// Sets the security level with which devices keys should be shared with
+enum ShareKeysWith {
+  /// Keys are shared with all devices if they are not explicitely blocked
+  all,
+
+  /// Once a user has enabled cross signing, keys are no longer shared with
+  /// devices which are not cross verified.
+  crossVerifiedIfAvailable,
```
coder-with-a-bushido, line 4093: can we pls rename this to `crossVerifiedIfEnabled`? It was a bit tricky to distinguish between `crossVerifiedIfAvailable` and `crossVerified` the first time.

coder-with-a-bushido: I was imagining "available" to the cross-signed devices availability among other devices.

test/device_keys_list_test.dart
```diff
@@ -150,11 +150,15 @@ void main() {
         },
         client,
       );
-      expect(client.shareKeysWithUnverifiedDevices, true);
+      client.shareKeysWith = ShareKeysWith.all;
```
td-famedly, line 153: This could cover more testcases like having no cross signing setup (test all 4 conditions), having cross signing setup but not verified yet (test all 4 conditions), etc.

lib/src/client.dart
```diff
@@ -90,7 +90,7 @@ class Client extends MatrixApi {
```
td-famedly, line 90: I am not sure what the outcome from the refinement was, but what happens if a client sets the sharewith enum to verifiedOnly but then tries to  verify a different user?

td-famedly: (again a test with this would be really helpful!)

test/device_keys_list_test.dart
krille-chan: I added more conditions, verified /unverified the key manually, cross verified it and disabled cross signing for this user. This should cover most cases I think. Adding more would be a lot of complexity...

lib/src/utils/device_keys_list.dart
```diff
@@ -163,7 +163,17 @@ abstract class SignableKey extends MatrixSignableKey {
...
+    switch (client.shareKeysWith) {
+      case ShareKeysWith.all:
+        return true;
+      case ShareKeysWith.crossVerifiedIfAvailable:
+        if (client.userDeviceKeys[userId]?.masterKey == null) return true;
+        return hasValidSignatureChain(verifiedByTheirMasterKey: true);
+      case ShareKeysWith.crossVerified:
+        return hasValidSignatureChain(verifiedByTheirMasterKey: true);
+      case ShareKeysWith.verifiedOnly:
+        return verified;
```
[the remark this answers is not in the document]
krille-chan: Then we should rename `verifiedOnly` to `directlyVerified`?

lib/src/client.dart
krille-chan: Good question but that would have been a problem before as well with the legacy `shareKeysWithUnverifiedDevices`. I would count this as a different issue

coder-with-a-bushido (approved):

lib/src/utils/device_keys_list.dart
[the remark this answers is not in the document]
td-famedly: Why are we not sending to cross verified devices tho? 🤔

lib/src/client.dart
td-famedly: Hm, can you please create an issue then? Feel free to resolve this thread after that!

lib/src/utils/device_keys_list.dart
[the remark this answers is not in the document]
td-famedly: Also `verified` still returns true for the cross verified devices if I am not wrong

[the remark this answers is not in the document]
krille-chan: > Why are we not sending to cross verified devices tho? 🤔

Directly verified is the highest form of trust. Cross verified is only like half trust.

> Also verified still returns true for the cross verified devices if I am not wrong
Yes because if something is directly verified it is the highest form of trust so this always returns true

lib/src/client.dart
krille-chan: https://github.com/famedly/product-management/issues/2845

td-famedly (approved):
lg!

PR #2022, opened 2025-02-06
refactor: Improve linebreak logic for html messages

krille-chan:


lib/src/utils/markdown.dart
```diff
@@ -211,6 +211,15 @@ String markdown(
   String? Function(String)? getMention,
   bool convertLinebreaks = true,
 }) {
+  // Workaround to keep linebreaks. We add a Zero-Width Non-Joiner in all
+  // empty lines to keep the lines as the user added them and workaround the
+  // markdown rules for multiple empty lines.
+  text = text
+      .trim()
+      .split('\n')
+      .map((line) => line.isEmpty ? '\u200C' : line)
```
Victorq10, line 220: You can not used this symbol as workaround because it used in Persian alphabet
https://en.wikipedia.org/wiki/Zero-width_non-joiner

krille-chan: What about the zero width space character U+200B then?

krille-chan: I changed it to marking it with a newly defined `/newline` marker. This gets ignored in markdown itself. The only side effect would be, that the user could type `/newline` everywhere which would be converted to a normal linebreak

lib/src/utils/markdown.dart
```diff
@@ -211,6 +211,16 @@ String markdown(
   String? Function(String)? getMention,
   bool convertLinebreaks = true,
 }) {
+  // Workaround to keep linebreaks. We define a `/newline` command and add this
+  // in all empty lines to keep the lines as the user added them and workaround
+  // the markdown rules for multiple empty lines.
+  text = text
+      .replaceAll('/newline', '\n')
```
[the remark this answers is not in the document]
krille-chan: Then they will have three new lines. I thought as we support commands with /anything anyway this is the best compromise. We already wasted countless hours to find a non hacky solution and havent found one

test/markdown_test.dart
```diff
@@ -70,14 +70,26 @@ void main() {
         'Snape killed <span data-mx-spoiler="">Dumbledoor <strong>bold</strong></span>',
       );
     });
-    test('multiple paragraphs', () {
-      expect(markdown('Heya!\n\nBeep'), '<p>Heya!</p><p>Beep</p>');
+    test('linebreaks', () {
+      expect(markdown('Heya!\nBeep'), 'Heya!<br/>Beep');
+      expect(markdown('Heya!\n\nBeep'), 'Heya!<br/><br/>Beep');
+      expect(markdown('Heya!\n\n\nBeep'), 'Heya!<br/><br/><br/>Beep');
+      expect(markdown('Heya!\n\n\n\nBeep'), 'Heya!<br/><br/><br/><br/>Beep');
```
krille-chan, line 77: Right now this seems to kill the end of lists. Will try to find a solution...

lib/src/utils/markdown.dart
krille-chan: I changed this to a completely new approach. Now it just replaces \n with <br/> if there are at least three \n right next to each other. So that means:

- \n -> unchanged (becomes <br/> later by markdown package)
- \n\n -> unchanged (becomes <p></p><p></p> later by markdown package)
- \n\n\n -> \n\n<br/> (becomes <p></p><p><br/></p> later by markdown package)
- \n\n\n\n -> \n\n<br/><br/> (becomes <p></p><p><br/><br/></p> later by markdown package)

and so on which results in the same amount of linebreaks as the user has desired (as long as the app displays ...</p><p>... as two linebreaks)

rmoradi88 (approved):

PR #2023, opened 2025-02-07
fix: Older state events overwrite newer ones on fetching history

krille-chan:
closes https://github.com/famedly/matrix-dart-sdk/issues/1772

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2665,8 +2665,27 @@ class Client extends MatrixApi {
       if (syncRoomUpdate is JoinedRoomUpdate) {
         final state = syncRoomUpdate.state;
 
+        // If we are receiving states when fetching history we need to check if
+        // we are not overwriting a newer state.
+        if (direction == Direction.b) {
+          await room.postLoad();
+          state?.removeWhere((state) {
+            final existingState =
+                room.getState(state.type, state.stateKey ?? '');
+            if (existingState == null) return false;
+            if (existingState is User) {
+              return existingState.originServerTs
+                      ?.isAfter(state.originServerTs) ??
+                  true;
+            }
+            if (existingState is MatrixEvent) {
+              return existingState.originServerTs.isAfter(state.originServerTs);
+            }
+            return true;
+          });
+        }
+
         if (state != null && state.isNotEmpty) {
-          // TODO: This method seems to be comperatively slow for some updates
           await _handleRoomEvents(
             room,
             state,
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -256,6 +256,7 @@ class Event extends MatrixEvent {
         typeKey: type,
         senderId: senderId,
         room: room,
+        originServerTs: originServerTs,
       );
 
   String get messageType => type == EventTypes.Sticker
--- a/lib/src/user.dart
+++ b/lib/src/user.dart
@@ -22,12 +22,14 @@ import 'package:matrix/matrix.dart';
 class User extends StrippedStateEvent {
   final Room room;
   final Map<String, Object?>? prevContent;
+  final DateTime? originServerTs;
 
   factory User(
     String id, {
     String? membership,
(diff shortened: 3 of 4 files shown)


PhantomRay:
I will test it now.

lib/src/client.dart
```diff
@@ -2650,8 +2650,19 @@ class Client extends MatrixApi {
...
         final state = syncRoomUpdate.state;
 
+        // If we are receiving states when fetching history we need to check if
+        // we are not overwriting a newer state.
+        if (direction == Direction.b) {
+          await room.postLoad();
+          state?.removeWhere((state) {
+            final existingState =
+                room.getState(state.type, state.stateKey ?? '');
+            if (existingState is! MatrixEvent) return false;
```
PhantomRay, line 2660: I noticed that you have MatrixEvent type check here. Perhaps to ensure timestamp is available. However, 

- As shown in screenshot, when an older event type is `m.room.member`, which is not supposed to override newer member info value, the type of this state event is `User`. It's not a MatrixEvent and without timestamp, thus will still override the newer value.
- If I have custom room state, it's will get overridden by older state as well.

I think one way to fix this once for all is to add timestamp to StrippedStateEvent. Then we can compare timestamps to decide whether or not to override the state value.

WechatIMG41073

krille-chan: Thank you very much for the analysis. I needed to do a little trick now as User inherits from StrippedStateEvent whcih does not have the originServerTs field, I had to add it to the User class now. Now it should work. Now it compares the originServerTs if available and it would be ready for another test

PhantomRay:
Testing it now.

PhantomRay:
Thanks for fixing this. I confirm that the followings are working as expected:
- user name/avatar for direct chat
- name/avatar for non-direct chat
- custom room state is not overwritten by older ones, and can be updated by newer ones

lib/src/client.dart
```diff
@@ -2650,8 +2650,27 @@ class Client extends MatrixApi {
...
 
+        // If we are receiving states when fetching history we need to check if
+        // we are not overwriting a newer state.
+        if (direction == Direction.b) {
+          await room.postLoad();
+          state?.removeWhere((state) {
+            final existingState =
+                room.getState(state.type, state.stateKey ?? '');
+            if (existingState == null) return false;
+            if (existingState is User) {
```
td-famedly, line 2661: Does this mean I need to add all state events which extend from StrippedStateEvent manually here? f.ex does `ChildrenState` work as expected?

lib/src/client.dart
```diff
@@ -2650,8 +2650,27 @@ class Client extends MatrixApi {
       if (syncRoomUpdate is JoinedRoomUpdate) {
         final state = syncRoomUpdate.state;
 
+        // If we are receiving states when fetching history we need to check if
+        // we are not overwriting a newer state.
+        if (direction == Direction.b) {
+          await room.postLoad();
```
td-famedly, line 2656: I am guessing this is fine because the only time you would be getting sync backwards is when you have already opened a room, correct?

PhantomRay: That's my experience and checking older messages.

lib/src/client.dart
krille-chan: that are only two types right now. `MatrixEvent` (include all other which inherit from this anyway) and `User` which is a special case and will be the only one forever. It is just to know if we have a originServerTs or not

lib/src/client.dart
krille-chan: yes, if this happens that you are usually requesting the history and then you have already opened the room. So this should not affect the performance.

PhantomRay:
Gentlemen, any plan to merge this?

td-famedly:
nitpick but would it be possible to add a test for this as well? Given that we have proper reproducible steps now?

td-famedly (approved):
lg

PR #2024, opened 2025-02-07
BREAKING: implement Native OIDC as per MSC 3861

TheOneWithTheBraid:
BREAKING: changes to the database API to better reflect the OIDC state

This implementation is based on the draft of native OIDC I implemented in < polycule >. It can be found [here](https://gitlab.com/polycule_client/polycule/-/blob/v0.1.1/lib/src/utils/matrix/oidc_delegation_extension.dart?ref_type=tags). Unlike in my initial draft, this implementation of course does not rely on any Flutter package but ships a minimal OIDC state machine relying on the matrix client to provide the native callback result via a `Completer`.

This implementation does not aim to comply to the full OIDC standard but only the subset required as per the MSCs stated below.

This PR does not aim to implement OIDC device flow as per [MSC 4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108) nor the changes for the application services as per [MSC 4190](https://github.com/matrix-org/matrix-spec-proposals/pull/4190).

Roadmap:

- [x] initial viable implementation
- [x] migrate the [< polycule > implementation](https://gitlab.com/polycule_client/polycule/-/merge_requests/245) to this OIDC stack
- [x] excessive testing
- [ ] figure out a test concept

Implements:

- [x] [MSC 3861](https://github.com/matrix-org/matrix-spec-proposals/pull/3861) - Next-generation auth for Matrix, based on OAuth 2.0/OIDC
  - [x] [MSC 1597](https://github.com/matrix-org/matrix-spec-proposals/pull/1597) - Better spec for matrix identifiers
  - [x] [MSC 2964](https://github.com/matrix-org/matrix-spec-proposals/pull/2964) - Usage of OAuth 2.0 authorization code grant and refresh token grant
  - [ ] [MSC 2965](https://github.com/matrix-org/matrix-spec-proposals/pull/2965) - OAuth 2.0 Authorization Server Metadata discovery
  - [x] [MSC 2966](https://github.com/matrix-org/matrix-spec-proposals/pull/2966) - Usage of OAuth 2.0 Dynamic Client Registration in Matrix
  - [x] [MSC 2967](https://github.com/matrix-org/matrix-spec-proposals/pull/2967) - API scopes
  - [x] [MSC 3824](https://github.com/matrix-org/matrix-spec-proposals/pull/3824) - OIDC aware clients
  - [ ] [MSC 4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108) - Mechanism to allow OIDC sign in and E2EE set up via QR code - **not planned for this PR**
  - [ ] [MSC 4190](https://github.com/matrix-org/matrix-spec-proposals/pull/4190) - Device management for application services - **not planned for this PR**
  - [x] [MSC 4191](https://github.com/matrix-org/matrix-spec-proposals/pull/4191) - Account management deep-linking


TheOneWithTheBraid:
A comment on the changes to the database API:

I initially intended to just keep all OIDC-related state (device ID, OAuth2.0 client ID, auth metadata) in memory until the OIDC flow completes. I sadly figured out the Dart process might get killed by the platform for battery optimization. In order to complete the flow after the main process was killed, I decided to store the OAuth2 client ID and the homeserver's auth metadata in the database as well as to split the device ID from the remaining client store since now the device ID is present a long time before the rest of the account is available.

Even though I so far already implemented the database API changes with this in mind, there is no way to resume an OIDC flow from a new thread yet.

TheOneWithTheBraid:
Review notice: So far, the default behavior of the updated `Client.getWellknown` is to only check the brand new [`/auth_metadata`](https://github.com/element-hq/synapse/pull/18093) endpoint. This might require Synapse nightly builds to work. Client developers however are free to use the previous revision of the MSC suggesting to use the [`/auth_issuer`](https://github.com/sandhose/matrix-spec-proposals/blob/msc/sandhose/oidc-discovery/proposals/2965-auth-metadata.md#discovery-via-openid-connect-discovery) endpoint or the `.well-known` entry containing the [`org.matrix.msc2965.authentication`](https://github.com/sandhose/matrix-spec-proposals/blob/msc/sandhose/oidc-discovery/proposals/2965-auth-metadata.md#discovery-via-the-well-known-client-discovery) field. All three possible implementations are supported in this PR. Since the default is definitely going to be the new `/auth_metadata`, the implementation already suggests this at this point.

TheOneWithTheBraid:
The < polycule > side implementation is now merged and works like a charm. Maybe that's helpful for testing this PR in real world: https://gitlab.com/polycule_client/polycule/-/merge_requests/245

TheOneWithTheBraid:
Since I unaskedly opened this PR, I'd kindly offer to take care of the OIDC code's maintenance in future too. You folks all have my MXID and can always poke me if there are future issues about the OIDC code.

TheOneWithTheBraid:
Clients might require the following workaround for #2027 : https://github.com/famedly/matrix-dart-sdk/issues/2027#issuecomment-2651178057

lib/encryption/utils/bootstrap.dart
```diff
@@ -494,12 +494,12 @@ class Bootstrap {
         }
       }
       if (newSsssKey != null) {
-        final storeFutures = <Future<void>>[];
+        Logs().v('Store new SSSS key entries...');
+        // NOTE(TheOneWithTheBraid): do not use Future.wait due to rate limits
+        // and token refresh trouble
         for (final entry in secretsToStore.entries) {
-          storeFutures.add(newSsssKey!.store(entry.key, entry.value));
+          await newSsssKey!.store(entry.key, entry.value);
```
krille-chan, line 501: Is this related?

TheOneWithTheBraid: Yes, the token rotation might otherwise happen during the requests as mentioned in the comment in the line above.

TheOneWithTheBraid (commented):
Is there any update on review of this PR ? It's now been months and months of waiting and I still did not even get a response on whether you are actually willing to ever accept this contribution. Instead only vaguely related comments.

lib/src/client.dart
```diff
@@ -304,24 +318,44 @@ class Client extends MatrixApi {
...
+      this.accessToken = accessToken = tokenResponse.accessToken;
+      newRefreshToken = tokenResponse.refreshToken;
+      expiresInMs = tokenResponse.expiresIn;
+    } else {
+      final tokenResponse = await refreshWithCustomRefreshTokenLifetime(
+        refreshToken,
+        refreshTokenLifetimeMs: customRefreshTokenLifetime?.inMilliseconds,
+      );
+
+      this.accessToken = accessToken = tokenResponse.accessToken;
```
td-famedly, line 343: nitpick repeated bit, would it be possible to deduplicate it?

lib/msc_extensions/msc_3861_native_oidc/msc4191_account_management/msc4191_account_management.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:matrix/matrix.dart';
+
+extension Msc4191AccountManagementExtension on Client {
```
td-famedly, line 3: only a little bit concerned that this is a draft msc

lib/msc_extensions/msc_3861_native_oidc/msc2966_oidc_dynamic_client_registration/msc2966_oidc_dynamic_client_registration.dart
```diff
@@ -0,0 +1,198 @@
...
+    required Uri url,
+    required Set<String> contacts,
+    required LocalizedOidcClientMetadata defaultLocale,
+    required Set<Uri> redirect,
+    String applicationType = 'native',
+    Map<String, LocalizedOidcClientMetadata> localizations = const {},
+  }) {
+    return OidcDynamicRegistrationData(
+      clientName: {
+        null: defaultLocale.clientName,
```
td-famedly, line 99: nitpick but having a map key as null doesn't really look right. Would {clientName: locale?} be better here? A little wrapper class would probably be a bit cleaner

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
```diff
@@ -0,0 +1,310 @@
...
+
+    Uri authEndpoint;
+    Uri tokenEndpoint;
+
+    try {
+      final authData = oidcAuthMetadata!;
+      authEndpoint = Uri.parse(authData['authorization_endpoint'] as String);
+      tokenEndpoint = Uri.parse(authData['token_endpoint'] as String);
+      // ensure we only hand over permitted prompts
+      if (prompt != null) {
```
td-famedly, line 37: If the prompt doesn't match, do we want to continue?

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
```diff
@@ -0,0 +1,310 @@
...
+        'response_type': 'code',
+        'response_mode': responseMode,
+        'redirect_uri': redirectUri.toString(),
+        'scope': scope.join(' '),
+        // not required per RFC but included due to
+        // https://github.com/element-hq/matrix-authentication-service/issues/2869
+        'state': state,
+        if (prompt != null) 'prompt': prompt,
+        'code_challenge':
+            // remove the "=" padding
```
td-famedly, line 150: Can you make sure to remove ALL the padding?

td-famedly (commented):
very draft review, will try to go through the whole flow and look at the client implementation after this

lib/src/client.dart
```diff
@@ -544,15 +579,22 @@ class Client extends MatrixApi {
...
+        }
+      }
 
       // Check if server supports at least one supported version
       final versions = await getVersions();
       if (!versions.versions
           .any((version) => supportedVersions.contains(version))) {
-        Logs().w(
-          'Server supports the versions: ${versions.toString()} but this application is only compatible with ${supportedVersions.toString()}.',
+        throw BadServerVersionsException(
```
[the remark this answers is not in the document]
TheOneWithTheBraid: Ooooops, yes, cleaned it up.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: Yes. A client might request a prompt the server does not know. In this case, the authentication server will fall back on the default selection showing all possible ways to proceed. The only behavior we need to avoid is passing an invalid prompt since this would cause the authentication server to show an error screen.

TheOneWithTheBraid: (E.g. a client could hard-code the two prompts `login` and `register`. In this case the SDK takes care of not prompting an invalid option to the MAS if e.g. registration is disabled.)

TheOneWithTheBraid: In fact, it's a pure convenience feature making client implementation easier and more reliable.

lib/src/database/matrix_sdk_database.dart
```diff
@@ -866,11 +864,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
         await _clientBox.put('refresh_token', refreshToken);
       }
       await _clientBox.put('user_id', userId);
-      if (deviceId == null) {
-        await _clientBox.delete('device_id');
-      } else {
-        await _clientBox.put('device_id', deviceId);
-      }
+      await _clientBox.delete('device_id');
```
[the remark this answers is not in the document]
TheOneWithTheBraid: Yeah, it's for migrating the device_id to its dedicated storage since with OIDC the device ID might be known before any other account data.

lib/msc_extensions/msc_3861_native_oidc/msc2966_oidc_dynamic_client_registration/msc2966_oidc_dynamic_client_registration.dart
TheOneWithTheBraid: Hmmmm, a null key in the map is sadly exactly what the Dart `Uir` class expects for its `queryParameters` property. I do not see need for introducing a new helper class just in order to later map it to a `Uri.queryParameters` instance again just because the syntax of this parameter is a bit weird. What would be the benefit for end users ? They won't touch this part of the code anyway since they only provide the parameters the SDK will later write into the `Map`.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: Dart's `base64` codec only adds the trailing `=` as padding, doesn't it ? Any suggestions how to better handle the padding ? I'm a bit clueless on how to do this without over engineering the implementation. Help appreciated.

lib/src/database/database_api.dart
```diff
@@ -357,6 +355,18 @@ abstract class DatabaseApi {
...
+
+  Future<String?> getDeviceId();
+
+  Future<void> storeOidcAuthMetadata(Map<String, Object?>? authMetadata);
+
+  Future<Map<String, Object?>?> getOidcAuthMetadata();
+
+  Future<void> storeOidcDynamicClientId(String? oidcClientId);
+
+  Future<String?> getOidcDynamicClientId();
```
krille-chan: I don't think we need this. Just use `getClient()` and pick what you need from it. To have it type safe we could later migrate it to returning a Dart Record.
Yeah that would fetch more data than necessary but I would prefer this over having the database API even more complex.

lib/msc_extensions/msc_3861_native_oidc/msc1597_matrix_identifier_syntax/msc1597_matrix_identifier_syntax.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:math';
+
+import 'package:matrix/matrix.dart';
+
+extension GenerateDeviceIdExtension on Client {
+  /// MSC 2964 & MSC 2967
+  Future<String> oidcEnsureDeviceId([bool enforceNewDevice = false]) async {
+    if (!enforceNewDevice) {
+      final storedDeviceId = await database?.getDeviceId();
```
krille-chan: Why do we need the stored device ID and can't just use `Client.deviceID`? It should be the same

lib/msc_extensions/msc_3861_native_oidc/msc1597_matrix_identifier_syntax/msc1597_matrix_identifier_syntax.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:math';
+
+import 'package:matrix/matrix.dart';
+
+extension GenerateDeviceIdExtension on Client {
+  /// MSC 2964 & MSC 2967
+  Future<String> oidcEnsureDeviceId([bool enforceNewDevice = false]) async {
```
krille-chan, line 12: Can you add more comments what this does? The method name is super unintuitive. Links to the MSCs would also be more helpful than MSC numbers.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
```diff
@@ -0,0 +1,310 @@
...
+import 'dart:convert';
+import 'dart:math';
+
+import 'package:http/http.dart' hide Client;
+
+import 'package:matrix/matrix.dart';
+import 'package:matrix/src/utils/crypto/crypto.dart';
+
+extension OidcOauthGrantFlowExtension on Client {
+  Future<void> oidcAuthorizationGrantFlow({
```
krille-chan: The method name says nothing about what it does? No idea what this is flow. Please switch this to a proper name with dart docs

lib/src/client.dart
```diff
@@ -2134,6 +2194,13 @@ class Client extends MatrixApi {
       _discoveryDataLoading = database.getWellKnown().then((data) {
         _wellKnown = data;
       });
+      _oidcAuthMetadataLoading = database.getOidcAuthMetadata().then((data) {
```
krille-chan: I doubt that this really needs to be waited for. Can't we just make it part of the `Database.getClient` call?

krille-chan (changes requested):
Looks quite good so far, though I'm not that into the MSCs of Matrix. Will test out if it works with oidc login on FluffyChat soon...

krille-chan (changes requested):
Can you add an example how to actually use this?

lib/src/database/database_api.dart
TheOneWithTheBraid: Done.

TheOneWithTheBraid:
> Can you add an example how to actually use this?

I added a README in the MSC folder for client developers to gather an example.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
```diff
@@ -0,0 +1,321 @@
...
+  /// - [oidcClientId]: the either registered or pre-shared OAuth2.0 client ID of the application. This can usually be generated using [Client.oidcRegisterOAuth2Client]
+  /// - [redirectUri]: the OAuth2.0 redirect Uri base the IDP is supposed to call
+  /// - [responseMode]: the OAuth2.0 response mode. This should usually be `fragment` or `query`
+  /// - [launchOAuth2Uri]: a callback for launching the computed OAuth2.0 Uri for authorization in the IDP. This should usually be from `package:url_launcher` or in some cases a handler launching a custom web view.
+  /// - [initialDeviceDisplayName]: the initial deice display name
+  /// - [enforceNewDeviceId]: whether to enforce generating a new device ID even in case the client already knows another device ID
+  /// - [prompt]: the OAuth2.0 authorization prompt. This must be a valid prompt from the [getOidcDiscoveryInformation] as by https://openid.net/specs/openid-connect-prompt-create-1_0.html
+  Future<void> oidcLoginAuthorizationGrantFlow({
+    required Completer<OidcCallbackResponse> nativeCompleter,
```
krille-chan, line 25: Why does this have to be a Completer and not just a async callback? A callback would be easier to use.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
krille-chan: I'm still not happy with this method name. A flow is a thing, not an action but a method is an action so it would better to be more intuitive to use a verb here like: `loginWithOidcAuthGrantFlow()`

lib/src/client.dart
krille-chan: This is still not resolved

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
```diff
@@ -0,0 +1,321 @@
...
+    required String oidcClientId,
```
krille-chan: If `oidc` is already part of the method name and there is no other clientId we can IMO just name this `clientId`. afaik other APIs using OIDC treat it the same

krille-chan (changes requested):
This now contains 22 commits. Can you clean up the commit history please?

lib/src/database/database_api.dart
```diff
@@ -357,6 +355,16 @@ abstract class DatabaseApi {
...
 
+  Future<void> storeDeviceId(String deviceId);
+
+  Future<String?> getDeviceId();
+
+  Future<void> storeOidcAuthMetadata(Map<String, Object?>? authMetadata);
+
+  Future<Map<String, Object?>?> getOidcAuthMetadata();
+
+  Future<void> storeOidcDynamicClientId(String? oidcClientId);
```
krille-chan: I'm not sure if this is the right approach to create methods for every single thing we store in the Client Database. This is now already messed up as we use getClient and updateClient to update single fields but now for other single fields we have methods. What about just adding one generic method `getClientData(ClientData key)` and `putClientData(ClientData key, value)` where `ClientData` is an enum of all possible fields

TheOneWithTheBraid: I migrated the String based key-value access on the client box contents to your proposed enum-based approach. I'd suggest to further extend this to a type safe approach including the values of the box but this is way beyond the scope of this OIDC-related PR.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: Sure, renamed !

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: It's not called `oidcLoginAuthorizationGrantFlow`. All methods related to OIDC use this prefix and I'd avoid having a gap word such as `with` in a method name.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: Sure, renamed !

lib/src/client.dart
TheOneWithTheBraid: Fixed with the migration to an enum-based client box.

lib/msc_extensions/msc_3861_native_oidc/msc2964_oidc_oauth_grants/msc2964_oidc_oauth_grants.dart
TheOneWithTheBraid: Maybe the major confusion comes from the work login : Using OIDC you do not "log in" using the matrix dart SDK but you login to your browser and use a token to *authenticate*. This process is simply not a login. That's why I prefer the technically accurate term `oidcAuthorizationGrantFlow`.

lib/encryption/utils/bootstrap.dart
krille-chan: I don't get it. Can you please explain?

lib/msc_extensions/msc_3861_native_oidc/README.md
```diff
@@ -0,0 +1,56 @@
...
+    },
+    applicationType: kIsWeb ? 'web' : 'native',
+  );
+
+  final oidcClientId = await client.oidcEnsureDynamicClientId(
+    registrationData,
+  );
+
+  // You can e.g. use `package:app_links` to forward the called deep link into this completer
+  final nativeCompleter = Completer<OidcCallbackResponse>();
```
krille-chan, line 43: Seeing this I've no idea how to use it with for example flutter_web_auth(2) package. In the end I get a URL I need to open where I pass a redirect_uri and receive a token back, right? How is this done here?

Is a Completer really the best choice? Would it be possible to use a callback instead? Something like `String Function(Uri uri) onAuth` which we then set like this:

```dart
await client.oidcAuthorizationGrantFlow(
  // ...
  onAuth: (Uri uri) async {
    final token = await FlutterWebAuth2.authenticate(uri);
    return token;
  }
);
```

Another question: In browser apps it might be desired to open it in the same tab which in a flutter context would mean that the app unloads and then starts again. So we need to store information somewhere and continue with the received token later. Is this also possible here?

krille-chan (changes requested):
I have now read into the topic of Native OIDC in Matrix as per MSC 3861 and described a new implementation plan of how the API should look like: https://github.com/famedly/matrix-dart-sdk/issues/2176
There I would vote for a much less complex approach which on the other hand allows clients to end and reopen the tab in between so we can have a same-tab-authentication. If you are willing to refactor your Pull Request that it matches the API described in #2176 and does not store any additional state anywhere (like in `oidcEnsureDeviceId()`, and remove all not related changes like in bootstrap.dart, then I would be willing to merge this PR.

krille-chan:
Has been implemented now already in a different PR

Issue #2028, opened 2025-02-11
Do not use `Client.oneShotSync` in SSSS and bootstrap


TheOneWithTheBraid:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

matrix-dart-sdk

### On which platform did the bug appear?

Firefox

### SDK Version

v0.38.0

### Describe the problem caused by this bug

`SSSS.createKey`, `OpenSSSS.store`, `Bootstrap.askSetupCrossSigning` and `Bootstrap.newSsss` make use of `Client.oneShotSync`. This is a huge slow down of SSSS handling since the homeserver might simply delay the `/sync` request up to the local timeout if there was nothing to sync. I'd propose to rather explicitly match on whether the keys are uploaded and locally processed than waiting for a sync.

### Steps To Reproduce

1. Create an empty account
2. Run SSSS boostrap
3. Wait for a loooooooong time

### Security related

Just do not in-carefully change this.

krille-chan:
Hey thank you for the feedback. I took a look into the code and don't yet see where this delays something. On the one hand it looks very similar to `await client.onSync.stream.where(//Filter here for desired account data key` and actually it was like this before but has been changed (don't know the reason maybe filtering in sync had problems with type escaping?)

However the current approach is just to wait until the next sync is finished and then check again what we have in memory and repeat this. While this is only waiting for the next sync if account data is not updated. This does not delay anything as we have to wait for the sync anyway to continue. I don't see where this would be faster than filtering the sync stream

TheOneWithTheBraid:
Yeah, I get what you mean. I don't have an easy proposal on how to better handle this. But fact is, on a completely new account, this causes SSSS bootstrap to take 5 - 10 minutes since each and every account data change first runs into a network timeout of 40 seconds (default SDK behavior) since there is nothing to sync. One approach might be to add fake syncs to the stream similar to sending an event in a room in order to have the updated account data available immediately after it was uploaded.

krille-chan:
@TheOneWithTheBraid  I tried now several things. I took a look into the code where we use `Client.oneShotSync()`. A good example is probably this one in the ssss.dart:

```dart
    await client.setAccountData(
      client.userID!,
      accountDataTypeKeyId,
      content.toJson(),
    );

[...]

PR #2032, opened 2025-02-13
refactor: Handle case account data arrives before put request finishes

krille-chan:
closes https://github.com/famedly/matrix-dart-sdk/issues/2028

### DIFF
--- a/lib/encryption/ssss.dart
+++ b/lib/encryption/ssss.dart
@@ -211,6 +211,7 @@ class SSSS {
       client.userID!,
       EventTypes.SecretStorageDefaultKey,
       SecretStorageDefaultKeyContent(key: keyId).toJson(),
+      waitForSync: true,
     );
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -3677,6 +3677,25 @@ class Client extends MatrixApi {
...
+    String userId,
+    String type,
+    Map<String, Object?> body, {
+    /// Waits for account data to come down from the sync loop before returns.
+    bool waitForSync = true,
+  }) async {
+    final syncFuture = onSync.stream
+        .where(
+          (syncUpdate) =>
+              syncUpdate.accountData?.any((data) => data.type == type) ?? false,
```
td-famedly, line 3691: this only checks the type while the previous version for checking for the exact data, also not clear how this fixes the original issue 🤔

lib/encryption/ssss.dart
```diff
@@ -211,6 +211,7 @@ class SSSS {
       client.userID!,
       EventTypes.SecretStorageDefaultKey,
       SecretStorageDefaultKeyContent(key: keyId).toJson(),
+      waitForSync: true,
```
coder-with-a-bushido, line 214: I don't understand how this solves "...explicitly match on whether the keys are uploaded and locally processed than waiting for a sync" as written in the linked issue.

PR #2033, opened 2025-02-13
feat: add refresh token handling http client

### LINKED ISSUE #2027: Make http requests soft logout aware
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

matrix-dart-sdk

### On which platform did the bug appear?

Firefox

### SDK Version

v0.38.0

### Describe the problem caused by this bug

(Note : the bug report form is really cursed.)

The `Client._innerSync` method currently seems to be the only place honoring `MatrixError.M_UNKNOWN_TOKEN` with `soft_logout`. As a developer using the SDK, I'd expect all called endpoints to handle a soft log out. The SDK already ships the `FixedTimeoutHttpClient`. I'd propose to implement an an additional layer similar to the [`RetryClient`](https://pub.dev/documentation/http/latest/retry/RetryClient-class.html) on SDK side filtering for soft logout responses and retrying after calling `Client.onSoftLogout`.

As a current workaround, I ship a `RetryClient` on my application side.

### Steps To Reproduce

1. Start SSSS bootstrap with a *really* short lived token (SSSS performs many individual requests)
2. Be slow with user input
3. See how it will fail due to an expired access token

Please note this might take a couple of times to retry since *usually* the `Client.sync` will refresh the access token fast enough.

### Screenshots or Logs

_No response_

### Security related

_No response_

TheOneWithTheBraid:
Sorry for once again coming up with a random PR ... Can I kindly ask you to check whether this implementation matches your expectations about handling soft-logout in more advanced use cases ?

This will especially be relevant for end users once #2024 might be shipped to client developers.

- implement a `MatrixRefreshTokenClient` to await token refresh before dispatching http requests
- improve documentation about soft logout logic
- fix dartdoc locations about soft logout

Fixes: #2027

### DIFF
--- a/lib/src/utils/matrix_refresh_token_client.dart
+++ b/lib/src/utils/matrix_refresh_token_client.dart
@@ -1,0 +1,71 @@
+import 'package:http/http.dart' hide Client;
+
+import 'package:matrix/matrix.dart';
+
+/// a [BaseClient] implementation handling [Client.onSoftLogout]
+///
+/// This client wrapper takes matrix [Client] as parameter and handles token
+/// rotation.
+///
+/// Before dispatching any request, it will check whether the [Client] supports
+/// token refresh by checking [Client.accessTokenExpiresAt]. Token rotation is
+/// done when :
+/// - refresh is supported ([Client.accessTokenExpiresAt])
+/// - we are actually initialized ([Client.onSync.value])
+/// - the request is to the homeserver rather than e.g. IDP ([BaseRequest.url])
+/// - the request is authenticated ([BaseRequest.headers])
+/// - we're logged in ([Client.isLogged])
+///
+/// In this case, [Client.ensureNotSoftLoggedOut] is awaited before running
+/// [BaseClient.send]. If the [Client.bearerToken] was changed meanwhile,
+/// the [BaseRequest] is being adjusted.
+class MatrixRefreshTokenClient extends BaseClient {
+  MatrixRefreshTokenClient({
+    required this.inner,
+    required this.client,
+  });
+
+  /// the matrix [Client] to handle token rotation for
+  final Client client;
+
+  /// the inner [BaseClient] to dispatch requests with
+  final BaseClient inner;
+
+  @override
+  Future<StreamedResponse> send(BaseRequest request) async {
+    Request? req;
+    if ( // only refresh if
+        // refresh is supported
+        client.accessTokenExpiresAt != null &&
+            // we are actually initialized
+            client.onSync.value != null &&
+            // the request is to the homeserver rather than e.g. IDP
+            request.url.host == client.homeserver?.host &&
+            // the request is authenticated
+            request.headers
+                .map((k, v) => MapEntry(k.toLowerCase(), v))
+                .containsKey('authorization') &&
+            // and last but not least we're logged in
+            client.isLogged()) {
+      try {
+        await client.ensureNotSoftLoggedOut();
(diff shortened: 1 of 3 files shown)


lib/src/utils/matrix_refresh_token_client.dart
```diff
@@ -0,0 +1,71 @@
...
+            // the request is to the homeserver rather than e.g. IDP
+            request.url.host == client.homeserver?.host &&
+            // the request is authenticated
+            request.headers
+                .map((k, v) => MapEntry(k.toLowerCase(), v))
+                .containsKey('authorization') &&
+            // and last but not least we're logged in
+            client.isLogged()) {
+      try {
+        await client.ensureNotSoftLoggedOut();
```
krille-chan, line 51: If we handle it here, can we maybe then remove the handling in the _sync() method? Does it then still work?

TheOneWithTheBraid: Well, I've been running into a) race conditions and b) I don't wanna enforce every user of the SDK to use the refresh token client since that one is only necessary when we're dealing with short-lived tokens.

TheOneWithTheBraid:
Is there any update on review of this PR ?

lib/src/utils/matrix_refresh_token_client.dart
```diff
@@ -0,0 +1,71 @@
...
+  @override
+  Future<StreamedResponse> send(BaseRequest request) async {
+    Request? req;
+    if ( // only refresh if
+        // refresh is supported
+        client.accessTokenExpiresAt != null &&
+            // we are actually initialized
+            client.onSync.value != null &&
+            // the request is to the homeserver rather than e.g. IDP
+            request.url.host == client.homeserver?.host &&
```
krille-chan, line 43: Can we also add this check: `request.url.pathSegments.first == '_matrix'`. Someone coud place for example a synapse module at: `/_custom/api/v1/...`

lib/src/utils/matrix_refresh_token_client.dart
```diff
@@ -0,0 +1,71 @@
...
+      } catch (e) {
+        Logs().w('Could not rotate token before dispatching HTTP request.', e);
+      }
+      // in every case ensure we run with the latest bearer token to avoid
+      // race conditions
+      finally {
+        final headers = request.headers;
+        // hours wasted : unknown :facepalm:
+        headers.removeWhere((k, _) => k.toLowerCase() == 'authorization');
+        headers['Authorization'] = 'Bearer ${client.bearerToken!}';
```
krille-chan, line 61: this now **always** adds the Authorization header right? Even if we don't have a token or don't want to set it for some reason.

krille-chan (changes requested):
Not sure if I'm happy with doing it on httpclient level. I really get why we want to do it this way we also check it unnecessarily often.

Doing this check on sync had the reason that we do it around every 30 seconds periodically. If we do it for every single request we would have this check done quite a lot, like multiple times per second.

With refresh tokens the access token has a lifetime of 5 minutes by default. Actually we know when it will be no longer valid so we could refresh it 1 minute before the lifetime ends. In the sync handler we do exactly this. If this is not working, we should investigate the reason for it

td-famedly:
I will close this for now

PR #2038, opened 2025-02-24
feat: Add command to upgrade room

krille-chan:
Related to https://github.com/famedly/product-management/issues/2579

### DIFF
--- a/lib/src/utils/commands_extension.dart
+++ b/lib/src/utils/commands_extension.dart
@@ -460,6 +460,16 @@ extension CommandsClientExtension on Client {
       stdout?.write(DefaultCommandOutput(users: [mxid]).toString());
       return null;
     });
+    addCommand('roomupgrade', (args, stdout) async {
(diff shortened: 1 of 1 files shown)


lib/src/utils/commands_extension.dart
```diff
@@ -460,6 +460,16 @@ extension CommandsClientExtension on Client {
       stdout?.write(DefaultCommandOutput(users: [mxid]).toString());
       return null;
     });
+    addCommand('roomupgrade', (args, stdout) async {
+      final version = args.msg;
+      if (version.isEmpty) {
```
coder-with-a-bushido, line 465: I think we should also validate the version string

krille-chan: Hmm, I think this could in theory be any string. The possible options look like this:

Bildschirmfoto 2025-03-06 um 14 55 05

coder-with-a-bushido (approved):

PR #2039, opened 2025-02-25
feat: Add parameter for allow condition for restricted rooms

krille-chan:


rmoradi88:
Would you give me a context of the problem that will be solved by?

lib/src/room.dart
```diff
@@ -2242,13 +2242,21 @@ class Room {
   }
 
   /// Changes the join rules. You should check first if the user is able to change it.
-  Future<void> setJoinRules(JoinRules joinRules) async {
+  Future<void> setJoinRules(
+    JoinRules joinRules, {
+    /// For restricted rooms, the room ID the user needs to be member:
```
mErrenst, line 2247: That sentence is really hard to understand. I think you mean "For restricted rooms, users can only become a member when they have this room ID".
If that is not meant, it would still make sense to make this comment (easier) understandable

krille-chan:
> Would you give me a context of the problem that will be solved by?

https://spec.matrix.org/latest/client-server-api/#restricted-rooms explains everything about it

rmoradi88 (approved):

PR #2040, opened 2025-02-26
feat: export timeline events

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/2858

### DIFF
--- a/lib/msc_extensions/extension_timeline_export/timeline_export.dart
+++ b/lib/msc_extensions/extension_timeline_export/timeline_export.dart
@@ -1,0 +1,286 @@
+import 'dart:convert';
+
+import 'package:matrix/matrix_api_lite.dart';
+import 'package:matrix/src/event.dart';
+import 'package:matrix/src/timeline.dart';
+
+extension TimelineExportExtension on Timeline {
+  /// Exports timeline events from a Matrix room within a specified date range.
+  ///
+  /// The export process provides progress updates through the returned stream with the following information:
+  /// - Total number of events exported
+  /// - Count of unable-to-decrypt (UTD) events
+  /// - Count of media events (images, audio, video, files)
+  /// - Number of unique users involved
+  ///
+  /// ```dart
+  /// // Example usage:
+  /// final timeline = room.timeline;
+  /// final oneWeekAgo = DateTime.now().subtract(Duration(days: 7));
+  ///
+  /// // Export last week's messages, excluding encrypted events
+  /// await for (final result in timeline.export(
+  ///   from: oneWeekAgo,
+  ///   filter: (event) => event?.type != EventTypes.Encrypted,
+  /// )) {
+  ///   if (result is ExportProgress) {
+  ///     print('Progress: ${result.totalEvents} events exported');
+  ///   } else if (result is ExportComplete) {
+  ///     print('Export completed with ${result.events.length} events');
+  ///   } else if (result is ExportError) {
+  ///     print('Export failed: ${result.error}');
+  ///   }
+  /// }
+  /// ```
+  ///
+  /// [from] Optional start date to filter events. If null, exports from the beginning.
+  /// [until] Optional end date to filter events. If null, exports up to the latest event.
+  /// [filter] Optional function to filter events. Return true to include the event.
+  /// [requestHistoryCount] Optional. The number of events to request from the server at once.
+  ///
+  /// Returns a [Stream] of [ExportResult] which can be:
+  /// - [ExportProgress]: Provides progress updates during export
+  /// - [ExportComplete]: Contains the final list of exported events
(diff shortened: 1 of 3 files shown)


lib/msc_extensions/extension_timeline_export/timeline_export.dart
```diff
@@ -0,0 +1,290 @@
...
+        utdEvents: utdEventsCount,
+        mediaEvents: mediaEventsCount,
+        users: users.length,
+      );
+    }
+  }
+}
+
+/// Base class for export results
+class ExportResult {
```
td-famedly, line 227: this could be a sealed class and then we can do exhaustive logic on it

lib/msc_extensions/extension_timeline_export/timeline_export.dart
```diff
@@ -0,0 +1,290 @@
...
+}
+
+enum ExportSource {
+  timeline,
+  database,
+  server,
+}
+
+/// Represents progress during export
+class ExportProgress extends ExportResult {
```
td-famedly, line 255: all of these could then be `final class`es just for keeping it a bit cleaner

krille-chan (changes requested):
Can we add tests for this method? 🤔

lib/msc_extensions/extension_timeline_export/timeline_export.dart
```diff
@@ -0,0 +1,290 @@
...
+        totalEvents: eventsToExport.length,
+        utdEvents: utdEventsCount,
+        mediaEvents: mediaEventsCount,
+        users: users.length,
+      );
+    }
+  }
+}
+
+/// Base class for export results
```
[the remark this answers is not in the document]
coder-with-a-bushido: It's easy to read and check the type in implementation this way. You can check the PR in app repo. But if you still want to do it, do let me know!

coder-with-a-bushido:
Added tests!

lib/msc_extensions/extension_timeline_export/timeline_export.dart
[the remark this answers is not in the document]
krille-chan: The only downside of using a record here would be that we don't have dart docs. However I would suggest to change the variable names a little bit anyway.

For example:

```dart
return ({totalEventsCount: 5, unableToDecryptEventsCount: 2, mediaEventsCount: 1, usersCount: 2});
```

This way we don't need to specify another class and with this terminology I think this should be intuitive enough that we don't need dart docs. What do you think? Just a suggestion.

test/msc_extensions/timeline_export_test.dart
```diff
@@ -0,0 +1,383 @@
+import 'package:test/test.dart';
+
+import 'package:matrix/matrix.dart';
+import 'package:matrix/msc_extensions/extension_timeline_export/timeline_export.dart';
+import 'package:matrix/src/models/timeline_chunk.dart';
+
+// Mock implementations
```
td-famedly, line 7: Can you add WHY you need to mock these as comments?

coder-with-a-bushido: done

lib/msc_extensions/extension_timeline_export/timeline_export.dart
```diff
@@ -0,0 +1,293 @@
...
+          events: eventsToExport,
+          totalEvents: eventsToExport.length,
+          utdEvents: utdEventsCount,
+          mediaEvents: mediaEventsCount,
+          users: users.length,
+        );
+        return;
+      }
+
+      final requestHistoryCount = 100;
```
krille-chan, line 110: Can we make this configurable in the method as a function parameter?

lib/msc_extensions/extension_timeline_export/timeline_export.dart
```diff
@@ -0,0 +1,293 @@
...
+
+      // From the database
+      var start = events.length;
+      do {
+        final eventsFromStore = await room.client.database?.getEventList(
+              room,
+              start: start,
+              limit: requestHistoryCount,
```
krille-chan, line 118: Do we even need a limit here? I guess we can safely just fetch everything from database at once

lib/msc_extensions/extension_timeline_export/timeline_export.dart
krille-chan: I would also think that for exporting we can use a higher number by default like 500

coder-with-a-bushido: done

lib/msc_extensions/extension_timeline_export/timeline_export.dart
coder-with-a-bushido: done

krille-chan (approved):

Issue #1831, opened 2024-05-28
timeline.startSearch() sometimes gives duplicated events back


krille-chan:
### Checklist

- [X] I could not find a solution in the documentation, the existing issues or discussions.
- [X] I already asked for help in the chat

### In which Project did the bug appear?

Other

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

FluffyChat

### On which platform did the bug appear?

Android, iOS, Chrome, Safari, Firefox, Edge

### SDK Version

latest

### Describe the problem caused by this bug

timeline.startSearch() sometimes gives duplicated entries back

coder-with-a-bushido:
@krille-chan can you pls give me steps to reproduce? I can't reproduce this one

coder-with-a-bushido:
I was able to reproduce this bug yesterday and this affects not only search but also https://github.com/famedly/product-management/issues/2858. I'm picking this up again now since I have a fix!

coder-with-a-bushido:
I can reproduce it again :(

PR #2045, opened 2025-03-11
fix: room prev_batch set incorrectly

coder-with-a-bushido:
Closes https://github.com/famedly/matrix-dart-sdk/issues/1831

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1322,7 +1322,7 @@ class Room {
     );
 
     if (onHistoryReceived != null) onHistoryReceived();
-    this.prev_batch = resp.end;
+    final newPrevBatch = direction == Direction.b ? resp.end : resp.start;
 
     Future<void> loadFn() async {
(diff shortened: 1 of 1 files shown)


lib/src/room.dart
```diff
@@ -1322,7 +1322,8 @@ class Room {
     );
 
     if (onHistoryReceived != null) onHistoryReceived();
-    this.prev_batch = resp.end;
+    final prevBatch = direction == Direction.b ? resp.end : resp.start;
+    this.prev_batch = prevBatch;
```
coder-with-a-bushido, line 1326: This is redundant because the following code also does the same - https://github.com/famedly/matrix-dart-sdk/blob/9a5a44ac661253f77d19904c811b05d831327268/lib/src/client.dart#L2992

Should I remove it?

krille-chan: Would vote for removing it then :)

krille-chan (approved):

PR #2047, opened 2025-03-12
fix: incorrect filename when caption is used in body of event

coder-with-a-bushido:
"body: If filename is not set or the value of both properties are identical, this is the filename of the original upload. Otherwise, this is a caption for the file."

refer https://spec.matrix.org/latest/client-server-api/#mfile


Pat992 (approved):
LGTM

PR #2050, opened 2025-03-13
refactor: Add file info to placeholder before sending file event

krille-chan:
This should display the placeholder for the sending file event better. Especially for images and videos which will directly have the correct width and height


mErrenst (approved):
LGTM

PR #2052, opened 2025-03-17
feat: allow skipping existing chat when calling startDirectChat

td-famedly:


lib/src/client.dart
```diff
@@ -2266,8 +2267,8 @@ class Client extends MatrixApi {
 
   /// Immediately start a sync and wait for completion.
   /// If there is an active sync already, wait for the active sync instead.
-  Future<void> oneShotSync() {
-    return _sync();
+  Future<void> oneShotSync({Duration? timeout}) {
+    return _sync(timeout: timeout);
```
krille-chan, line 2271: Looks unrelated to the commit message 🤔

krille-chan (approved):

lib/src/client.dart
td-famedly: ah yes, but i'll add it in the commit message, I wanted this as well :D

PR #2054, opened 2025-03-17
fix: set prev_batch when seen in sync

td-famedly:


coder-with-a-bushido:
I will test it once and then review today

krille-chan (approved):
@coder-with-a-bushido you may want to check if your fix from https://github.com/famedly/matrix-dart-sdk/commit/0d084d8824b256d127409ae6d3d336184612530c still works

PR #2055, opened 2025-03-18
fix: only set prev_batch once when seen in sync

td-famedly:


lib/src/client.dart
```diff
@@ -2925,6 +2925,13 @@ class Client extends MatrixApi {
   /// stores when we last checked for stale calls
   DateTime lastStaleCallRun = DateTime(0);
 
+  /// we don't want to reset the prev_batch everytime
+  /// because that would break our search and export
+  /// but if we find a room update with just state,
+  /// we make sure to update the prev_batch in the next sync update
+  /// just once.
+  Set<String> justSeenInvites = {};
```
krille-chan, line 2933: I don't like that we add another state here. Do we really need this? This has the danger that it will go out of sync, also this is writeable from the outside. What happens after clear cache, logout/login and all the other possible edge cases? What if we add another clear method and then forgot to add this? Isn't there another solution without a new state?

td-famedly: Changing it to private was easy, clearCache shouldn't really affect this because set and we only really care about this during a subsequent JoinedRoomUpdate. I am not really a fan of this solution either, but I couldn't think of anything better atm.

One other fix would be that when doing search or export, we keep a log of all the prev_batch we have already processed and then if the prev_batch gets reset, we check through this log first. This feels more hacky than this PR though.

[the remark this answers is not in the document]
td-famedly: hm probably, but then this solution gets more hacky

td-famedly:
Will discuss alternative fix for export/search brb

PR #2056, opened 2025-03-18
fix: set prev_batch first time from sync response correctly

coder-with-a-bushido:
Should fix what https://github.com/famedly/matrix-dart-sdk/pull/2055 intended to do. Thanks @td-famedly for the test files.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2984,14 +2984,18 @@ class Client extends MatrixApi {
                 (chatUpdate.unreadNotifications?.highlightCount ?? 0) ||
             chatUpdate.summary != null ||
             chatUpdate.timeline?.prevBatch != null)) {
+      /// [InvitedRoomUpdate] doesn't have prev_batch, so we want to set it in case
+      /// the room first appeared in sync update when membership was invite.
+      if (rooms[roomIndex].membership == Membership.invite &&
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -2984,14 +2984,18 @@ class Client extends MatrixApi {
                 (chatUpdate.unreadNotifications?.highlightCount ?? 0) ||
             chatUpdate.summary != null ||
             chatUpdate.timeline?.prevBatch != null)) {
+      /// [InvitedRoomUpdate] doesn't have prev_batch, so we want to set it in case
+      /// the room first appeared in sync update when membership was invite.
+      if (rooms[roomIndex].membership == Membership.invite &&
```
krille-chan, line 2989: this is not gonna work unfortunately :-( Not all rooms have membership invite before. If you join a public room for example or you accepted the invite on a different client and then open another client for the same account, the room is directly in join

coder-with-a-bushido: In case of public room or accepting invite from different client, the code works already afaik.

https://github.com/famedly/matrix-dart-sdk/blob/3f4c9da90b075b49ad92e233d6441cddda413866/lib/src/client.dart#L2948

krille-chan: ahhh now I got it. Thank you so much. This way it totally makes sense.

PR #2058, opened 2025-03-20
fix: create outboundsession with foreign userIds correctly

td-famedly:
TODO: add testcase

### DIFF
--- a/lib/encryption/key_manager.dart
+++ b/lib/encryption/key_manager.dart
@@ -303,7 +303,8 @@ class KeyManager {
         continue;
       }
       final userDeviceKeyIds = deviceKeyIds[device.userId] ??= <String, bool>{};
-      userDeviceKeyIds[deviceId] = !device.encryptToDevice;
+      userDeviceKeyIds[deviceId] =
+          !device.encryptToDevice && device.userId == client.userID!;
     }
     return deviceKeyIds;
   }
@@ -440,7 +441,9 @@ class KeyManager {
           'session_key': sess.outboundGroupSession!.session_key(),
         };
         try {
-          devicesToReceive.removeWhere((k) => !k.encryptToDevice);
+          devicesToReceive.removeWhere(
+            (k) => (!k.encryptToDevice && k.userId == client.userID!),
+          );
           if (devicesToReceive.isNotEmpty) {
             // update allowedAtIndex
             for (final device in devicesToReceive) {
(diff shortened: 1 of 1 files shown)


lib/encryption/key_manager.dart
```diff
@@ -552,7 +555,8 @@ class KeyManager {
 
     final deviceKeys = await room.getUserDeviceKeys();
     final deviceKeyIds = _getDeviceKeyIdMap(deviceKeys);
-    deviceKeys.removeWhere((k) => !k.encryptToDevice);
+    deviceKeys
+        .removeWhere((k) => !k.encryptToDevice && k.userId == client.userID!);
```
krille-chan, line 559: Shouldn't this be an OR? 🤔 We want to remove devices where `device.encryptToDevice` is `false` OR the device is not from the own user

lib/encryption/key_manager.dart
```diff
@@ -440,7 +441,9 @@ class KeyManager {
           'session_key': sess.outboundGroupSession!.session_key(),
         };
         try {
-          devicesToReceive.removeWhere((k) => !k.encryptToDevice);
+          devicesToReceive.removeWhere(
+            (k) => (!k.encryptToDevice && k.userId == client.userID!),
```
krille-chan, line 445: Why don't we share the new keys with other users here? This is for sharing the own session with other users afterwards. I think what you actually want is, disable key requests->forwarding from other users. But not sharing keys for normal e2ee

lib/encryption/key_manager.dart
krille-chan: Also same here. If we don't share the outbound group session with other users, we can not communicate with them at all

lib/encryption/key_manager.dart
```diff
@@ -303,7 +303,8 @@ class KeyManager {
         continue;
       }
       final userDeviceKeyIds = deviceKeyIds[device.userId] ??= <String, bool>{};
-      userDeviceKeyIds[deviceId] = !device.encryptToDevice;
+      userDeviceKeyIds[deviceId] =
+          !device.encryptToDevice && device.userId == client.userID!;
```
krille-chan, line 307: I'm not sure what this method actually does. It has no description and a non intuitive name. But I guess it is to convert a list of devices keys to a map to know if we want to share keys with the specific device of a user. However we don't know for waht purpose we use this. If this is for normal key sharing we need to send the keys to other users as well

PR #2059, opened 2025-03-21
refactor: user -> member

Yash-Garg:
Closes https://github.com/famedly/matrix-dart-sdk/issues/305


td-famedly (changes requested):
Given that this is such an old issue, I would like to bring this back to the team to see if we still want to do this. I really don't think this is worth the breaking change.

lib/src/database/database_api.dart
```diff
@@ -104,9 +104,9 @@ abstract class DatabaseApi {
 
   Future<void> clear();
 
-  Future<User?> getUser(String userId, Room room);
+  Future<Member?> getUser(String userId, Room room);
```
td-famedly, line 107: well now this is confusing, are we getting a user or a member? cc: @krille-chan this might be more breaking than we first thought if we also start changing all the function names

krille-chan: I agree to create new methods where the term `User` could be changed to `Member` in the function name, and mark the old ones as deprecated would be the best here @Yash-Garg  could you do this?

PR #2066, opened 2025-04-04
chore: Check version endpoint before using dehydrated devices

krille-chan:
closes https://github.com/famedly/product-management/issues/2165


lib/msc_extensions/msc_3814_dehydrated_devices/msc_3814_dehydrated_devices.dart
```diff
@@ -31,6 +31,13 @@ extension DehydratedDeviceHandler on Client {
         return;
       }
 
+      final versions = await this.getVersions();
+      if (versions.unstableFeatures?.tryGet<bool>('org.matrix.msc3814') !=
```
krille-chan, line 35: @nico-famedly do you know if this is the correct key to look for? I haven't found it on any of our servers

PR #2067, opened 2025-04-08
Deleting last message is edited

rmoradi88:
close #https://github.com/famedly/app/issues/6398

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2854,7 +2854,7 @@ class Client extends MatrixApi {
           room.setState(user);
         }
       }
-      _updateRoomsByEventUpdate(room, event, type);
+      await _updateRoomsByEventUpdate(room, event, type);
       if (store) {
         await database?.storeEventUpdate(room.id, event, type, this);
       }
@@ -3020,11 +3020,11 @@ class Client extends MatrixApi {
     return room;
   }
 
-  void _updateRoomsByEventUpdate(
+  Future<void> _updateRoomsByEventUpdate(
     Room room,
     StrippedStateEvent eventUpdate,
     EventUpdateType type,
-  ) {
+  ) async {
     if (type == EventUpdateType.history) return;
 
     switch (type) {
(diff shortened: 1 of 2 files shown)


td-famedly (dismissed):
blocking comment because this should have a testcase, feel free to dismiss once there's one :D

lib/src/client.dart
```diff
@@ -3060,13 +3060,27 @@ class Client extends MatrixApi {
...
           break;
         }
+        // Is this event redacting the last event which is a edited event.
+        if (event.type == EventTypes.Redaction &&
+            ({
+              room.lastEvent?.relationshipEventId,
+            }.contains(
+              event.redacts ?? event.content.tryGet<String>('redacts'),
+            )) &&
+            room.lastEvent?.relationshipType == RelationshipTypes.edit) {
```
krille-chan, line 3076: I think we can simplify this a little bit to:

```dart
if (event.type == EventTypes.Redaction && room.lastEvent?.relationshipEventId == (event.redacts ?? event.content.tryGet<String>('redacts')) {
// ...
```
The relationshipType check do we really need it? I would think that for all relationship types this would be the best behavior.

krille-chan: Maybe also before do a null check on `relationshipEventId` beforehand

lib/src/client.dart
```diff
@@ -3060,13 +3060,27 @@ class Client extends MatrixApi {
...
+          final originalEvent =
+              await room.getEventById(room.lastEvent!.relationshipEventId!);
```
krille-chan, line 3078: I would suggest to use `(await client.database?.getEventById(eventId, room)) ?? room.lastEvent;` here instead and . `room.getEventById` does a network request if not found which would block the sync and that I would like to avoid. If not found we can just fallback to redact the edited one I would say

lib/src/client.dart
rmoradi88: The relationship type is necessary. Assume last event is a reply event on an edited event. If you remove the edited event we have wrong last event in this case.

rmoradi88:
Fun, before changes if you have replied message to and edited message as last message, after deleting the edited message your replied one also will be deleted after clearing cache :D

lib/src/client.dart
rmoradi88: Fixed. Thanks

lib/src/client.dart
rmoradi88: Fix, thanks

krille-chan: ah right, I forgot that a reply could also be this

PR #2068, opened 2025-04-09
chore: flutter upgrade

asabri97:


lib/matrix_api_lite/utils/print_logs_web.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
td-famedly, line 1: hm this might lead to hidden deprecations, can you be more explict instead of disabling them all over the file please?

lib/src/database/indexeddb_box.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
krille-chan, line 1: Hmm, not sure if it wouldn't be better to directly migrate everything to web. In the SDK actually we don't need to rush for the dart upgrade. Maybe worth another issue I guess? @asabri97

Issue #2051, opened 2025-03-16
Calls hang up instantly after sending invite to call


Airyzz:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

https://github.com/commetchat/commet

### On which platform did the bug appear?

Android

### SDK Version

v0.32.4

### Describe the problem caused by this bug

When placing a call, after the `m.call.invite` event is sent, the call is automatically hung up.

I think this is because during `onCallInvite`, the call is not being detected as already existing, and so a new `CallSession` gets created, resetting `_inviteOrAnswerSent` to false. This causes `initWithInvite` to hangup the call

I noticed calls are only being detected as already existing if `call.state` is ended, which doesn't appear correct to me, is it meant to be if `call.state` != ended?

https://github.com/famedly/matrix-dart-sdk/blob/609ee1d4a72c409041f929fa4d1da4da013d7dd2/lib/src/voip/voip.dart#L387 

Changing this line to:
```
if (call != null && call.state != CallState.kEnded) {
```

Seems to behave as I would expect

coder-with-a-bushido:
Confirmed that https://github.com/commetchat/matrix-dart-sdk/pull/5 needs to be done upstream as well

alexantropoff:
I have the same issue, fix helped me too

PR #2070, opened 2025-04-11
fix: handling of existing calls in `onCallInvite`

yash-garg:
Closes https://github.com/famedly/matrix-dart-sdk/issues/2051

### DIFF
--- a/lib/src/voip/voip.dart
+++ b/lib/src/voip/voip.dart
@@ -384,7 +384,7 @@ class VoIP {
       '[glare] got new call ${content.tryGet('call_id')} and currently room id is mapped to ${incomingCallRoomId.tryGet(room.id)}',
     );
 
-    if (call != null && call.state == CallState.kEnded) {
+    if (call != null && call.state != CallState.kEnded) {
       // Session already exist.
       Logs().v('[VOIP] onCallInvite: Session [$callId] already exist.');
       return;
--- a/test/calls_test.dart
+++ b/test/calls_test.dart
@@ -932,5 +932,57 @@ void main() {
       expect(room.groupCallParticipantCount('participants_count'), 2);
       expect(room.hasActiveGroupCall, true);
     });
+
+    test('call persists after sending invite', () async {
+      CallSession? incomingCall;
+
+      // incoming call should not be created yet
+      incomingCall = voip.calls[voip.currentCID];
+      expect(incomingCall, isNull);
+      expect(incomingCall?.pc, isNull);
+
+      // send invite for the call
+      final outgoingCall = await voip.inviteToCall(
+        room,
+        CallType.kVoice,
+        userId: '@alice:testing.com',
+      );
+
+      // acknowledge the invite
+      await matrix.handleSync(
+        SyncUpdate(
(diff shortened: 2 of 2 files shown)


td-famedly (changes requested):
can you also add a testcase? I am not sure why we missed this but should be easy to test

test/calls_test.dart
```diff
@@ -932,5 +932,59 @@ void main() {
...
+    test('call persists after sending invite', () async {
+      expect(voip.currentCID, null);
+
+      final outgoingCall = await voip.inviteToCall(
+        room,
+        CallType.kVoice,
+        userId: '@alice:testing.com',
+      );
+
+      await Future.delayed(Duration(seconds: 1));
```
td-famedly, line 945: hm do we need this? if yes why? can you add that as a comment please?

test/calls_test.dart
```diff
@@ -932,5 +932,59 @@ void main() {
...
+                    ),
+                  ],
+                ),
+              ),
+            },
+          ),
+        ),
+      );
+
+      await Future.delayed(Duration(seconds: 2));
```
td-famedly, line 976: same here

test/calls_test.dart
```diff
@@ -932,5 +932,55 @@ void main() {
...
+      );
+
+      expect(
+        voip.calls.containsKey(
+          VoipId(roomId: room.id, callId: outgoingCall.callId),
+        ),
+        true,
+      );
+
+      expect(outgoingCall.state, CallState.kCreateOffer);
```
td-famedly, line 981: can you also check whether `initWithInvite` was triggered?  I think checking whether pc was null before the call and then you have a pc should be enough, and maybe add comments on why you are testing it like this

lib/src/voip/voip.dart
```diff
@@ -860,10 +860,10 @@ class VoIP {
...
       throw MatrixSDKVoipException(
         '''
-        User ${client.userID}:${client.deviceID} is not allowed to join famedly calls in room ${room.id}, 
-        canJoinGroupCall: ${room.canJoinGroupCall}, 
-        groupCallsEnabledForEveryone: ${room.groupCallsEnabledForEveryone}, 
-        needed: ${room.powerForChangingStateEvent(EventTypes.GroupCallMember)}, 
+        User ${client.userID}:${client.deviceID} is not allowed to join famedly calls in room ${room.id},
+        canJoinGroupCall: ${room.canJoinGroupCall},
+        groupCallsEnabledForEveryone: ${room.groupCallsEnabledForEveryone},
+        needed: ${room.powerForChangingStateEvent(EventTypes.GroupCallMember)},
```
coder-with-a-bushido, line 866: I think removing this space change should fix the codecov/patch job

coder-with-a-bushido:
@Yash-Garg - Also, either @Airyzz could to sign the CLA or we can credit them in commit message description so we can merge this without being blocked by the license/cla check.

Airyzz:
I'd prefer not to sign the CLA at this point, remove me from commit history if you like, i don't mind.

PR #2076, opened 2025-05-04
fix: Do not load timeline while sync is processed

krille-chan:
This sometimes lead to the
problem that the timeline misses
the newest messages. Especially
when opening the app from a
notification this can happen.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1525,14 +1525,16 @@ class Room {
   }) async {
     await postLoad();
 
-    List<Event> events;
+    var events = <Event>[];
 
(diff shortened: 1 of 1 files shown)


coder-with-a-bushido (approved):

td-famedly:
@krille-chan would it be possible to add tests for this?

krille-chan:
> @krille-chan would it be possible to add tests for this?

I fear not. The thing is we could try to trigger the bug, but that is very random to get. In the same millisecond we need to call a sync and getEventList and then it is still random...

PR #2078, opened 2025-05-06
fix: make sure prev_batch is updated after a limited timeline

td-famedly:
fixes: https://github.com/famedly/app/issues/6465

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2991,8 +2991,9 @@ class Client extends MatrixApi {
             chatUpdate.timeline?.prevBatch != null)) {
       /// [InvitedRoomUpdate] doesn't have prev_batch, so we want to set it in case
       /// the room first appeared in sync update when membership was invite.
-    if ((rooms[roomIndex].membership == Membership.invite ||
-              chatUpdate.timeline?.limited == true) &&
+      if ((rooms[roomIndex].membership == Membership.invite ||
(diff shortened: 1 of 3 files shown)


rmoradi88 (approved):

lib/src/client.dart
```diff
@@ -2991,7 +2991,8 @@ class Client extends MatrixApi {
             chatUpdate.timeline?.prevBatch != null)) {
       /// [InvitedRoomUpdate] doesn't have prev_batch, so we want to set it in case
       /// the room first appeared in sync update when membership was invite.
-      if (rooms[roomIndex].membership == Membership.invite &&
+    if ((rooms[roomIndex].membership == Membership.invite ||
+              chatUpdate.timeline?.limited == true) &&
```
coder-with-a-bushido, line 2995: this should be updated even if prevBatch is null

krille-chan: are we sure?

krille-chan: If prev_batch is `null` then `requestHistory` will throw an exception. It has worked years by only do the `!= null` check. So I would not update it, if prev_batch is null

coder-with-a-bushido: yes, we had that prevBatch != null condition chained here because we didn't want to update the prevBatch in the first sync after joining IF the room already has it set

Issue #2025, opened 2025-02-11
Events with status error sometimes still get sent


konrad-xtention:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

Other

### On which platform did the bug appear?

iOS

### SDK Version

0.38.0

### Describe the problem caused by this bug

Sometimes messages that have error status get sent regardless. This is bad UX as messages get sent that no user expects to be sent, and that they maybe don't even want to be sent.

I think the main problem here is that the `Event` constructor and the sending queue have different logic to determine when to set the error status.

The [Event constructor](https://github.com/famedly/matrix-dart-sdk/blob/beef67bd0219c6acaa213a4cf46f38b050f015f5/lib/src/event.dart#L126-L130) sets the Event to error when `originServerTs` is more than `Client.sendTimelineEventTimeout` ago. In [`Room.sendEvent`](https://github.com/famedly/matrix-dart-sdk/blob/beef67bd0219c6acaa213a4cf46f38b050f015f5/lib/src/room.dart#L1108) the Event is set to error when the _time it started sending_ is more than `Client.sendTimelineEventTimeout` ago. But the time it started sending can be way later than `originServerTs`, depending on how many other events are in the sending queue and how long they take to send or fail.

I think those two places need to be aligned so they mark an Event as error at the same time, and Events marked as error should never be sent unless `Event.sendAgain` is called. 

### Steps To Reproduce

1. Go offline
2. Send multiple messages (with `Room.sendTextEvent`)
3. Wait
4. Close and reopen the room (so that `Room.getTimeline` gets called)
5. Maybe wait some more
6. Notice that all messages sent before now have `Event.status == EventStatus.error`
7. Go online again
8. Notice that some of the messages that already had error status are getting sent

krille-chan:
I can reproduce it. Messages are suddenly sent even after they were marked as error

PR #2082, opened 2025-05-08
fix: incorrect Event status update in constructor

coder-with-a-bushido:
Closes https://github.com/famedly/matrix-dart-sdk/issues/2025


lib/src/room.dart
```diff
@@ -1106,11 +1111,14 @@ class Room {
         },
       ),
     );
+    // we need to add the transaction ID to the set of events that are currently queued to be sent
+    // even before the fake sync is called, so that the event constructor can check if the event is in the sending state
+    sendingQueueEventsByTxId.add(messageID);
```
krille-chan, line 1116: Do we really need to duplicate this state? Can't we just check by `sendingQueue.any((event) => event.transactionId == txid)`?

coder-with-a-bushido: the problem is that `sendingQueue` is `<Completer>[]`

coder-with-a-bushido: and we need to block it before the fake sync that sends the event with sending status

lib/src/room.dart
```diff
@@ -72,7 +72,12 @@ class Room {
   /// Key-Value store for private account data only visible for this user.
   Map<String, BasicEvent> roomAccountData = {};
 
-  final _sendingQueue = <Completer>[];
+  /// Queue of sending events
+  /// NOTE: This shouldn't be modified directly, use [sendEvent] instead. This is only used for testing.
+  final sendingQueue = <Completer>[];
+
+  /// Set of transaction IDs of events that are currently queued to be sent
+  final sendingQueueEventsByTxId = <String>[];
```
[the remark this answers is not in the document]
coder-with-a-bushido: I'll fix it

lib/src/room.dart
krille-chan: ah got it, okay then it's fair

krille-chan (approved):

PR #2083, opened 2025-05-08
feat: Support fallback for threads in Event.getReplyEvent()

krille-chan:
Supports fallback for threads according to https://spec.matrix.org/v1.14/client-server-api/#fallback-for-unthreaded-clients

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -481,13 +481,35 @@ class Event extends MatrixEvent {
   Future<String?> redactEvent({String? reason, String? txid}) async =>
       await room.redactEvent(eventId, reason: reason, txid: txid);
 
-  /// Searches for the reply event in the given timeline.
+  /// Searches for the reply event in the given timeline. Also returns the
+  /// event fallback if the relationship type is `m.thread`.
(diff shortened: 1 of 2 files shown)


lib/src/event.dart
```diff
@@ -481,8 +481,33 @@ class Event extends MatrixEvent {
...
+              .tryGetMap<String, Object?>('m.in_reply_to')
+              ?.tryGet<String>('event_id');
+        } else {
+          relationshipEventId = this.relationshipEventId;
+        }
+        return relationshipEventId == null
+            ? null
+            : await timeline.getEventById(relationshipEventId);
+    }
     if (relationshipType != RelationshipTypes.reply) return null;
```
coder-with-a-bushido, line 511: nitpick but can we remove this line and add a `default` case in `switch` to return null?

coder-with-a-bushido: no, I think we should remove the rest of the code

krille-chan: ah right, done :)

coder-with-a-bushido (approved):
lgtm

PR #2085, opened 2025-05-08
fix: Fetch well-known from domain from userId, not from homeserver do…

krille-chan:
…main

as described here: https://spec.matrix.org/v1.14/client-server-api/#well-known-uri

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -585,7 +585,13 @@ class Client extends MatrixApi {
   /// The result of this call is stored in [wellKnown] for later use at runtime.
   @override
   Future<DiscoveryInformation> getWellknown() async {
-    final wellKnown = await super.getWellknown();
+    final wellKnownResponse = await httpClient.get(
+      Uri.https(userID!.domain!, '/.well-known/matrix/client'),
+    );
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -585,7 +585,13 @@ class Client extends MatrixApi {
   /// The result of this call is stored in [wellKnown] for later use at runtime.
   @override
   Future<DiscoveryInformation> getWellknown() async {
-    final wellKnown = await super.getWellknown();
+    final wellKnownResponse = await httpClient.get(
+      Uri.https(userID!.domain!, '/.well-known/matrix/client'),
```
coder-with-a-bushido, line 589: should we also consider changing `baseUri` of `MatrixApi` to this instead of `homeserver`?

krille-chan: No that is correct. It should be set to the base_uri from well-known while well-known itself **should** be fetched from the domain of the userID

coder-with-a-bushido: aahhhh ok, I got it

coder-with-a-bushido (approved):

PR #2092, opened 2025-05-14
refactor: (BREAKING) Make database required

krille-chan:
Closes https://github.com/famedly/product-management/issues/110

As described in the issue this makes the database in the Client no more nullable, which makes a lot of app code easier as we no longer need to do any null checks. To init a client, we now define a `database` parameter. The `databaseBuilder` does no longer exist. It looks like this:

```dart
import 'package:sqflite/sqflite.dart';

final client = Client(
    '<Client Name>',
    database: await MatrixSdkDatabase.init(
        '<Database Name>',
        database: await openDatabase('/path/to/database.sqlite'),
    ),
);
```

I also made the default constructor for MatrixSdkDatabase private, so the user is forced to use the `init()` method. This way it is no longer possible to create a database and forget to open it which leads to an error anyway.

This change is a major refactoring but makes a lot of code much easier I would say.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -67,11 +67,9 @@ class Client extends MatrixApi {
 
   int? get id => _id;
 
-  final FutureOr<DatabaseApi> Function(Client)? databaseBuilder;
   final FutureOr<DatabaseApi> Function(Client)? legacyDatabaseBuilder;
-  DatabaseApi? _database;
 
-  DatabaseApi? get database => _database;
+  final DatabaseApi database;
 
   Encryption? get encryption => _encryption;
   Encryption? _encryption;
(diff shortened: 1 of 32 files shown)


lib/src/client.dart
```diff
@@ -67,11 +67,9 @@ class Client extends MatrixApi {
 
   int? get id => _id;
 
-  final FutureOr<DatabaseApi> Function(Client)? databaseBuilder;
   final FutureOr<DatabaseApi> Function(Client)? legacyDatabaseBuilder;
-  DatabaseApi? _database;
 
-  DatabaseApi? get database => _database;
+  DatabaseApi database;
```
coder-with-a-bushido, line 72: nitpick but having a getter here sounds right to me. Any reason we want to expose the variable?

Making this variable final also sounds good to me

krille-chan: Good point. I'll make it final

coder-with-a-bushido (approved):
lgtm :)

PR #2093, opened 2025-05-15
refactor: event-getter

asabri97:
For: https://github.com/famedly/product-management/issues/2210


lib/src/event.dart
```diff
@@ -557,18 +557,18 @@ class Event extends MatrixEvent {
 
   /// Gets the mimetype of the attachment of a file event, or a blank string if not present
   String get attachmentMimetype => infoMap['mimetype'] is String
-      ? infoMap['mimetype'].toLowerCase()
+      ? (infoMap['mimetype'] as String).toLowerCase()
```
krille-chan, line 560: Suggestion: By using tryGet we could make this like:

```dart
String get attachmentMimetype => infoMap.tryGet<String>('mimetype')?.toLowerCase() ?? //...
```

lib/src/event.dart
```diff
@@ -557,18 +557,18 @@ class Event extends MatrixEvent {
...
       : (content
               .tryGetMap<String, Object?>('file')
               ?.tryGet<String>('mimetype') ??
           '');
 
   /// Gets the mimetype of the thumbnail of a file event, or a blank string if not present
   String get thumbnailMimetype => thumbnailInfoMap['mimetype'] is String
-      ? thumbnailInfoMap['mimetype'].toLowerCase()
+      ? (thumbnailInfoMap['mimetype'] as String).toLowerCase()
```
krille-chan, line 568: Same here

lib/src/event.dart
```diff
@@ -557,18 +557,18 @@ class Event extends MatrixEvent {
...
       : (infoMap['thumbnail_file'] is Map &&
-              infoMap['thumbnail_file']['mimetype'] is String
-          ? infoMap['thumbnail_file']['mimetype']
+              (infoMap['thumbnail_file'] as Map)['mimetype'] is String
+          ? (infoMap['thumbnail_file'] as Map)['mimetype'] as String
```
krille-chan, line 571: also here with tryGetMap: `infoMap.tryGetMap<String, Object?>('thumbnail_file')?.tryGet<String>('mimetype')`

lib/src/event.dart
```diff
@@ -582,7 +582,7 @@ class Event extends MatrixEvent {
   /// Gets the underlying mxc url of a thumbnail of a file event, or null if not present
   Uri? get thumbnailMxcUrl {
     final url = isThumbnailEncrypted
-        ? infoMap['thumbnail_file']['url']
+        ? (infoMap['thumbnail_file'] as Map)['url']
```
krille-chan, line 585: Here the cast is even dangerous as it would break if this wasn't a Map. The JSON inside of the event content is fully user generated so an attacker could send `"thumbnail_file": 0.0` which would be a double and cause the whole app to crash here.

We would be safe by using tryGetMap here as well:
```dart
final url = isThumbnailEncrypted
  ? infoMap.tryGetMap<String, Object?>('thumbnail_file')?.tryGet<String>('url')
  : infoMap.tryGet<String>('thumbnail_url');
```

krille-chan (changes requested):
Could you check the other lines where you using `as ??` type casts as well? In the best case we would try to avoid them everywhere

lib/src/event.dart
```diff
@@ -802,16 +800,17 @@ class Event extends MatrixEvent {
 
     // Decrypt the file
     if (isEncrypted) {
-      final fileMap =
-          getThumbnail ? infoMap['thumbnail_file'] : content['file'];
-      if (!fileMap['key']['key_ops'].contains('decrypt')) {
+      final fileMap = getThumbnail
+          ? infoMap['thumbnail_file'] as Map
```
td-famedly, line 804: all of these could probably use the tryGet now right?

td-famedly (changes requested):

td-famedly:
I will close this for now, please reopen it if it's still needed

PR #2096, opened 2025-05-23
refactor: Do not handle leave rooms on initial sync

krille-chan:
This saves some ressources as
we do not need to delete rooms
as long as the database is empty anyway.
Also for some reason this fixes
an edge case where the keys
cache in the sqlite box gets
corrupted. It is unfortunate
that I wasn't able to find the
root cause or be able to reproduce
it. Wasted 4 hours now.

Closes https://github.com/famedly/product-management/issues/3190

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2483,7 +2483,7 @@ class Client extends MatrixApi {
       // room both in invite and leave. If you get an invite and then leave, it
       // will only be included in leave.
       final leave = sync.rooms?.leave;
-      if (leave != null) {
+      if (leave != null && prevBatch != null) {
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -2483,7 +2483,7 @@ class Client extends MatrixApi {
       // room both in invite and leave. If you get an invite and then leave, it
       // will only be included in leave.
       final leave = sync.rooms?.leave;
-      if (leave != null) {
+      if (leave != null && prevBatch != null) {
```
td-famedly, line 2486: According to https://github.com/famedly/product-management/issues/3190, if you can reproduce it in the sdk, can you add a test for it please?

PR #2100, opened 2025-05-28
feat: Migrate to vodozemac

krille-chan:


README.md
```diff
@@ -4,15 +4,15 @@ Matrix (matrix.org) SDK written in dart.
...
+For E2EE, vodozemac must be provided.
 
 Additionally, OpenSSL (libcrypto) must be provided on native platforms for E2EE.
 
 For flutter apps you can easily import it with the [flutter_olm](https://pub.dev/packages/flutter_olm) and the [flutter_openssl_crypto](https://pub.dev/packages/flutter_openssl_crypto) packages.
 
 ```sh
 flutter pub add matrix
-flutter pub add flutter_olm
+flutter pub add dart_vodozemac
```
coder-with-a-bushido: should this be flutter_vodozemac?

doc/get-started.md
```diff
@@ -9,7 +9,6 @@ In your `pubspec.yaml` file add the following dependencies:
   # If you plan to use the SDK in a Flutter application on IO:
   sqflite: <latest-version>
   # For end to end encryption:
-  flutter_olm: <latest-version>
   flutter_openssl_crypto: <latest-version>
```
coder-with-a-bushido, line 15: do we not want to add flutter_vodozemac here?

doc/web.md
```diff
@@ -1,25 +0,0 @@
-To use end to end encryption in web you have to download the olm javascript/wasm library:
```
coder-with-a-bushido, line 1: can we also have new web docs here? I think it'll be similar

lib/encryption/key_manager.dart
```diff
@@ -669,54 +657,57 @@ class KeyManager {
...
+      for (final sessionEntry in roomEntry.value.sessions.entries) {
+        final sessionId = sessionEntry.key;
+        final session = sessionEntry.value;
+        final sessionData = session.sessionData;
+        Map<String, Object?>? decrypted;
+        try {
+          decrypted = json.decode(
+            decryption.decrypt(
+              vod.PkMessage(
+                base64decodeUnpadded(sessionData['ciphertext'] as String),
```
coder-with-a-bushido: we could use `PkMessage.fromBase64` and avoid decoding here

lib/encryption/key_manager.dart
```diff
@@ -669,54 +657,57 @@ class KeyManager {
...
-              senderClaimedKeys:
-                  decrypted.tryGetMap<String, String>('sender_claimed_keys') ??
-                      <String, String>{},
-              uploaded: true,
-            );
-          }
+            ),
+          );
+        } catch (e, s) {
+          Logs().e('[LibOlm] Error decrypting room key', e, s);
```
coder-with-a-bushido: `[LibOlm]` should be changed to `[Vodozemac]`

lib/encryption/key_manager.dart
```diff
@@ -1266,30 +1249,27 @@ RoomKeys generateUploadKeysImplementation(GenerateUploadKeysArgs args) {
...
       roomKeyBackup.sessions[sess.sessionId] = KeyBackupData(
-        firstMessageIndex: sess.inboundGroupSession!.first_known_index(),
+        firstMessageIndex: sess.inboundGroupSession!.firstKnownIndex,
         forwardedCount: sess.forwardingCurve25519KeyChain.length,
         isVerified: dbSession.verified, //device?.verified ?? false,
         sessionData: {
-          'ephemeral': encrypted.ephemeral,
-          'ciphertext': encrypted.ciphertext,
-          'mac': encrypted.mac,
+          'ephemeral': encrypted.ephemeralKey.toBase64(),
```
coder-with-a-bushido: we can use `toBase64` method here to get a tuple and then use it

lib/encryption/olm_manager.dart
```diff
@@ -301,17 +297,14 @@ class OlmManager {
...
         for (final otk in signedOneTimeKeys.values) {
           // Keys can only be removed by creating a session...
-          final session = olm.Session();
-          try {
-            final String identity =
-                json.decode(olmAccount.identity_keys())['curve25519'];
-            final key = otk.tryGet<String>('key');
-            if (key != null) {
-              session.create_outbound(_olmAccount!, identity, key);
-              olmAccount.remove_one_time_keys(session);
```
coder-with-a-bushido, line 311: umm, did I miss something in writing the bindings? Is there any equivalent of this in vodozemac or does it happen when you mark the keys as published?

lib/encryption/olm_manager.dart
```diff
@@ -477,26 +458,27 @@ class OlmManager {
...
-        _olmAccount!.remove_one_time_keys(newSession);
-        await encryption.olmDatabase?.updateClientKeys(pickledOlmAccount!);
+        final result = _olmAccount!.createInboundSession(
+          theirIdentityKey: vod.Curve25519PublicKey.fromBase64(senderKey),
+          preKeyMessageBase64: body,
+        );
+        plaintext = result.plaintext;
+        final newSession = result.session;
 
-        plaintext = newSession.decrypt(type, body);
```
coder-with-a-bushido, line 486: umm, why do we not try to decrypt again now?

scripts/prepare_vodozemac.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+git clone https://github.com/famedly/dart-vodozemac.git
+mv ./dart-vodozemac/rust ./
+rm -rf dart-vodozemac
+cd ./rust
+cargo build
+cd ..
```
coder-with-a-bushido, line 9: should we think about hosting the build? I remember you saying about the US regulation but still I'm wondering

lib/encryption/olm_manager.dart
coder-with-a-bushido: I can also check this later

lib/encryption/utils/key_verification.dart
```diff
@@ -1451,31 +1448,31 @@ class _KeyVerificationMethodSas extends _KeyVerificationMethod {
...
     await request.send('m.key.verification.key', {
-      'key': sas!.get_pubkey(),
+      'key': sas!.publicKey,
     });
   }
 
   void _handleKey(Map<String, dynamic> payload) {
     theirPublicKey = payload['key'];
-    sas!.set_their_key(payload['key']);
+    establishedSas = sas!.establishSasSecret(payload['key']);
```
coder-with-a-bushido: I think we should check if the Sas obj is disposed before calling this here. I'll create a method for checking

coder-with-a-bushido: Or if it's very sure that it will called only once per `Sas` object, then the code flow should be simpler

lib/encryption/utils/olm_session.dart
```diff
@@ -40,21 +43,24 @@ class OlmSession {
...
-    session = olm.Session();
     try {
-      session!.unpickle(key, dbEntry['pickle']);
+      try {
+        session = vod.Session.fromPickleEncrypted(
+          pickleKey: key.toPickleKey(),
+          pickle: dbEntry['pickle'],
+        );
+      } catch (_) {
+        session = vod.Session.fromOlmPickleEncrypted(
```
coder-with-a-bushido, line 54: We should probably have an info log here

lib/encryption/utils/outbound_group_session.dart
```diff
@@ -54,19 +55,24 @@ class OutboundGroupSession {
...
-      creationTime =
-          DateTime.fromMillisecondsSinceEpoch(dbEntry['creation_time']);
+      outboundGroupSession = vod.GroupSession.fromPickleEncrypted(
+        pickleKey: key.toPickleKey(),
+        pickle: dbEntry['pickle'],
+      );
     } catch (e, s) {
-      dispose();
-      Logs().e('[LibOlm] Unable to unpickle outboundGroupSession', e, s);
+      try {
```
coder-with-a-bushido, line 68: we should have an info log here as well

lib/encryption/utils/olm_session.dart
coder-with-a-bushido: and should we have another try-catch block for `fromOlmPickleEncrypted` like we have for outbound_group_session?

lib/encryption/utils/session_key.dart
```diff
@@ -99,15 +101,20 @@ class SessionKey {
...
     try {
-      inboundGroupSession!.unpickle(key, dbEntry.pickle);
+      inboundGroupSession = vod.InboundGroupSession.fromPickleEncrypted(
+        pickle: dbEntry.pickle,
+        pickleKey: key.toPickleKey(),
+      );
     } catch (e, s) {
-      dispose();
-      Logs().e('[LibOlm] Unable to unpickle inboundGroupSession', e, s);
+      try {
```
coder-with-a-bushido, line 109: an info log here

lib/src/utils/native_implementations.dart
```diff
@@ -133,17 +133,24 @@ class NativeImplementationsDummy extends NativeImplementations {
...
+  NativeImplementationsIsolate(
+    this.compute, {
+    /// To generate upload keys, vodozemac needs to be initialized in the isolate.
+    this.vodozemacInit,
+  });
 
   /// creates a [NativeImplementationsIsolate] based on a [ComputeRunner] as
   // ignore: deprecated_member_use_from_same_package
   /// known from [Client.runInBackground]
   factory NativeImplementationsIsolate.fromRunInBackground(
```
coder-with-a-bushido: I think there's no usage for this factory method anywhere else now. Should we remove this?

lib/src/utils/matrix_file.dart
```diff
@@ -145,10 +140,6 @@ class MatrixImageFile extends MatrixFile {
     @Deprecated('Use [nativeImplementations] instead') ComputeRunner? compute,
```
coder-with-a-bushido: should this `compute` be removed?

lib/src/utils/matrix_file.dart
```diff
@@ -199,10 +190,6 @@ class MatrixImageFile extends MatrixFile {
...
```
coder-with-a-bushido: should this `compute` be removed as well?

lib/src/utils/native_implementations.dart
coder-with-a-bushido: If we remove this, we could also remove `computeCallbackFromRunInBackground` from `compute_callback.dart` file

lib/src/utils/native_implementations.dart
```diff
@@ -133,17 +133,24 @@ class NativeImplementationsDummy extends NativeImplementations {
 class NativeImplementationsIsolate extends NativeImplementations {
   /// pass by Flutter's compute function here
   final ComputeCallback compute;
+  final Future<void> Function()? vodozemacInit;
```
coder-with-a-bushido, line 136: do we want to make this just a future that we can wait on?

coder-with-a-bushido: This way we can just store `vod.init()` in a variable after the function call and then use it here

coder-with-a-bushido (changes requested):
Looks really good except for some nitpicks I've added in comments :)

README.md
```diff
@@ -4,15 +4,17 @@ Matrix (matrix.org) SDK written in dart.
...
 
 Additionally, OpenSSL (libcrypto) must be provided on native platforms for E2EE.
 
 For flutter apps you can easily import it with the [flutter_olm](https://pub.dev/packages/flutter_olm) and the [flutter_openssl_crypto](https://pub.dev/packages/flutter_openssl_crypto) packages.
 
 ```sh
 flutter pub add matrix
-flutter pub add flutter_olm
+
+# Optional: For end to end encryption:
```
coder-with-a-bushido, line 16: I don't think this is optional. Won't not adding this also mean that things like SAS don't work?

README.md
krille-chan: Fixed

doc/get-started.md
krille-chan: Fixed

lib/encryption/key_manager.dart
krille-chan: Nice catch! Done

lib/encryption/key_manager.dart
krille-chan: Nice catch. Thank you. Done

doc/web.md
krille-chan: Done :)

lib/encryption/key_manager.dart
krille-chan: Fixed

lib/encryption/olm_manager.dart
krille-chan: You are right. If I see it correctly this is a rare case where we need this method: https://docs.rs/vodozemac/latest/vodozemac/olm/struct.Account.html#method.remove_one_time_key

The edge case here is: We failed to upload One Time Keys. So we need to invalidate all of them. In libolm you can do this but need to add the session object, that's why we craeted a new outbound_session without really using it.

In Vodozemac I just skipped it without fully understanding it. Sorry for that. Can you add the remove_one_time_key method from vodozemac to the bindings?

lib/encryption/olm_manager.dart
krille-chan: the decryption is done when in the `createInboundSession()` method already now in Vodozemac. It returns a tuple with the session and the plaintext

scripts/prepare_vodozemac.sh
krille-chan: I'd still prefer to not do it. Maybe we could cache it somehow? Any ideas?

lib/encryption/utils/olm_session.dart
krille-chan: Added an info log. Another try catch would be overhead IMO. Because fromOlmPickleEncrypted() would already crash and be catched by the underlying try block

lib/encryption/utils/session_key.dart
krille-chan: Done

lib/src/utils/native_implementations.dart
krille-chan: Nice catch. Removed

lib/src/utils/matrix_file.dart
krille-chan: Removed

lib/src/utils/native_implementations.dart
krille-chan: No, this is actually important to be executed again. I tried it out. You need to call `vod.init();` again in an isolate every time you spawn the isolate

README.md
krille-chan: The whole app can work without any end to end encryption. Then of course SAS also don't work. FYI: There is a version of FluffyChat out there which has end to end encryption completely disabled for multiple reasons

lib/encryption/utils/outbound_group_session.dart
krille-chan: Done

lib/encryption/utils/key_verification.dart
krille-chan: Done :) can you check?

lib/encryption/olm_manager.dart
krille-chan: Added

doc/get-started.md
```diff
@@ -8,9 +8,6 @@ In your `pubspec.yaml` file add the following dependencies:
   matrix: <latest-version>
   # If you plan to use the SDK in a Flutter application on IO:
   sqflite: <latest-version>
-  # For end to end encryption:
```
coder-with-a-bushido, line 11: I think we should have this and add a link to the end-to-end encryption doc

scripts/prepare_vodozemac.sh
coder-with-a-bushido: Can we host it on some EU server instead of in GitHub?

coder-with-a-bushido (approved):
LGTM except those 2 comments :)

doc/get-started.md
krille-chan: done :)

scripts/prepare_vodozemac.sh
krille-chan: No need for this. Our vodozemac repository only contains bindings and none of the prohibited encryption algorithms. We only import them there but they need to be built by the consumer anyway

PR #2104, opened 2025-06-06
refactor: Disable benchmarks by global boolean

krille-chan:
Benchmarks are flooting the
logs. We should disable them by default and allow
having low debug level
at the same time.

### DIFF
--- a/lib/src/utils/run_benchmarked.dart
+++ b/lib/src/utils/run_benchmarked.dart
@@ -18,6 +18,8 @@
 
 import 'package:matrix/matrix.dart';
 
+var kEnableMatrixSdkBenchmarks = false;
+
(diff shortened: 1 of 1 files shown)


lib/src/utils/run_benchmarked.dart
```diff
@@ -18,6 +18,8 @@
 
 import 'package:matrix/matrix.dart';
 
+var kEnableMatrixSdkBenchmarks = false;
```
yash-garg, line 21:
```suggestion
const kEnableMatrixSdkBenchmarks = false;
```
are we changing this anywhere else? if not, we can make this a `const`.

krille-chan: Yes this is intended to be switched one for those who are interested in the benchmarks

PR #2105, opened 2025-06-08
refactor: Follow up new database make possible to reuse client again

krille-chan:


lib/src/client.dart
```diff
@@ -2202,10 +2202,8 @@ class Client extends MatrixApi {
       _backgroundSync = true;
     } catch (e, s) {
       Logs().e('Unable to clear database', e, s);
-    } finally {
       await database.delete();
```
td-famedly, line 2205: why are we not deleting the database anymore? this would be a breaking change. Also do we not need to dispose now?

td-famedly (changes requested):

lib/src/client.dart
krille-chan: Because we only have one database now, not a databaseBuilder anymore. Before this was just creating a new database, now we reuse the same database, which is not a problem at all as sqflite and indexeddb are made for this

td-famedly: Hm in the app we encrypt the database and we want to change the key for every login. Have you tested doing this? Does reusing the database work with a different password?

krille-chan: it reuses the key now after logout/login. However all data is wiped before. So why should this be a problem? 🤔

krille-chan: However, if we want, we could just destroy the client on logout and instantly create a new one with a new database by calling `client.database.delete();` and `client.dispose();`

krille-chan: Nevermind, maybe we can keep it as it is and just create a new Client. I'm doing this now in FluffyChat. Might be a cleaner solution anyway

PR #2112, opened 2025-06-18
refactor: Handling soft logout via a timer

krille-chan:


lib/src/client.dart
```diff
@@ -2275,30 +2278,68 @@ class Client extends MatrixApi {
...
-        await onSoftLogout(this);
-        onLoginStateChanged.add(LoginState.loggedIn);
-      } catch (e, s) {
-        Logs().w('Unable to refresh session after soft logout', e, s);
-        await logout();
-        rethrow;
+
+      async.Result? softLogoutResult;
+
+      while (softLogoutResult?.isValue != true) {
```
td-famedly, line 2284: wouldn't https://github.com/famedly/matrix-dart-sdk/pull/2033 with a debounce be better?

PR #2114, opened 2025-07-10
fix: Can not logout and login again with same Client object

krille-chan:


mErrenst (approved):

test/client_test.dart
```diff
@@ -116,6 +117,22 @@ void main() {
       expect(client.isLogged(), true);
 
       await client.logout();
+
```
td-famedly, line 120: the relogging in works before the fix, I'm not sure this test makes sense? :thinking:

krille-chan: It does not work before the fix

lib/src/client.dart
```diff
@@ -2194,14 +2204,14 @@ class Client extends MatrixApi {
     }
     try {
       await abortSync();
-      await database.clear();
+      if (!_disposed) await database.clear();
       await legacyDatabase?.clear();
       _backgroundSync = true;
     } catch (e, s) {
```
td-famedly, line 2210: can we make this a breaking change then? we are significantly changing the api in my opinion

krille-chan: Added

td-famedly (approved):
client devs are required to delete the database on their own now. We only clear the database

PR #2118, opened 2025-07-17
Krille/matrix security predisclosure

krille-chan:
Related to https://matrix.org/blog/2025/07/security-predisclosure/


lib/src/room.dart
```diff
@@ -1930,13 +1930,32 @@ class Room {
...
   /// If a user_id is in the users list, then that user_id has the associated
   /// power level. Otherwise they have the default level users_default.
   /// If users_default is not supplied, it is assumed to be 0. If the room
   /// contains no m.room.power_levels event, the room’s creator has a power
   /// level of 100, and all other users have a power level of 0.
+  /// For room version 12 and above the room creator always has maximum
+  /// power level.
   int getPowerLevelByUserId(String userId) {
+    // Room creator has maximum power level:
+    if (userId == creatorUserId &&
```
td-famedly: probably want to do the same for `additional_creators`?

lib/src/room.dart
```diff
@@ -1930,13 +1930,32 @@ class Room {
     }
   }
 
+  /// Returns the room version if specified in the `m.room.create` state event.
```
td-famedly, line 1933: > Finally, clients which use power_level_content_override when creating rooms MUST NOT assign a power level to the room creator, otherwise the /createRoom request will fail.

we should manually remove that incase we get that in our `createRoom`

lib/src/room.dart
krille-chan: What is `additional_creators`? I find nothing in the matrix spec about it. However the current changes only mention the single creator

lib/src/room.dart
krille-chan: Not sure... I would leave this up to the client developer as it is actually the server's side to make sure that this does not happen. So the server should prevent the request anyway

lib/src/room.dart
td-famedly: > One of the other changes coming in v12 is that room creators will be privileged over other users in the room. Therefore clients which restrict user behaviour based on power level will need to be updated to be aware that room creators effectively have infinite power level. Creators are not listed in the users block of the m.room.power_levels event, and are instead defined as the sender field of the m.room.create event, or entries in a new optional additional_creators array field in the content of the create event. Full details will be released in the MSCs when embargo lifts.

https://matrix.org/blog/2025/07/security-predisclosure/

lib/src/room.dart
td-famedly: ye the server prevents it, but this is just a nice thing to do so all the clients don't have to make sure of it, I don't see why not?

krille-chan: It adds a lot of complexity for not much benefit. It would crash anyway as the server would responde with an error message. There are already a lot of other ways you can send requests with the sdk which the server would not allow and we can never catch all of them. IMO it would make sense if this depends on user input but as this is just something an app usually hard codes, this has not a big benefit for the complexity it adds here

td-famedly: I don't think it is complicated to implement at all and just a nice to have thing. But feel free to resolve the comment :)

lib/src/room.dart
krille-chan: ah got it. Thanks. Have updated it

PR #2120, opened 2025-08-05
feat: Use small versions of bullet point characters

Kelrap:


krille-chan:
@Kelrap Can you also adjust the tests in the text_to_html_test.dart file? If not we can also help you with it. Just need write access to your repo.

Kelrap:
@krille-chan The tests should be updated now. I also invited you as a collaborator, if I missed anything.

PR #2123, opened 2025-08-08
fix: fix the issue that user avatar can not be loaded on windows

### LINKED ISSUE #2113: [Bug on Windows] Avatars cannot be loaded properly on Windows.
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### In which Project did the bug appear?

_No response_

### If you selected \"Other\" as Project, please enter in which project the bug occurred.

_No response_

### On which platform did the bug appear?

Chrome

### SDK Version

1.0.1

### Describe the problem caused by this bug

I saw this code in database_file_storage_io.dart:
```
File _getFileFromMxc(Uri mxcUri) => File(
        '${Directory.fromUri(fileStorageLocation!).path}/${mxcUri.toString().split('/').last}',
      );
```

On windows, it returns path like: 
`C:\Users\xxx\AppData\Local\Temp\/filename`

This produces PathNotFoundException when loading user avatar.

### Steps To Reproduce

_No response_

### Screenshots or Logs

_No response_

### Security related

_No response_

huluwa-dev:
Fixes #2113

### DIFF
--- a/lib/src/database/database_file_storage_io.dart
+++ b/lib/src/database/database_file_storage_io.dart
@@ -1,6 +1,8 @@
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:path/path.dart';
+
 import 'package:matrix/matrix.dart';
 
 mixin DatabaseFileStorage {
@@ -9,9 +11,19 @@ mixin DatabaseFileStorage {
   late final Uri? fileStorageLocation;
   late final Duration? deleteFilesAfterDuration;
 
-  File _getFileFromMxc(Uri mxcUri) => File(
-        '${Directory.fromUri(fileStorageLocation!).path}/${mxcUri.toString().split('/').last}',
-      );
+  /// Map an MXC URI to a local File path
+  File _getFileFromMxc(Uri mxcUri) {
+    // Replace all special characters with underscores to avoid PathNotFoundException on Windows.
+    final host = mxcUri.host.replaceAll('.', '_');
+    final path = mxcUri.pathSegments.join('_');
+    final query = mxcUri.queryParameters.entries
+        .map((entry) => '${entry.key}${entry.value}')
+        .join('_');
+    final fileName = '${host}_${path}_$query';
+    return File(
+      join(Directory.fromUri(fileStorageLocation!).path, fileName),
+    );
+  }
 
   Future<void> storeFile(Uri mxcUri, Uint8List bytes, int time) async {
     final fileStorageLocation = this.fileStorageLocation;
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -24,6 +24,7 @@ dependencies:
   js: ^0.6.3
   markdown: ^7.1.1
   mime: ">=1.0.0 <3.0.0"
+  path: ^1.9.1
   random_string: ^2.3.1
   sdp_transform: ^0.3.2
   slugify: ^2.0.0
@@ -39,6 +40,5 @@ dev_dependencies:
   file: ">=6.1.1 <8.0.0"
   import_sorter: ^4.6.0
   lints: ^5.0.0
-  path: ^1.9.1
   sqflite_common_ffi: ^2.3.4+4 # sqflite_common_ffi aggressively requires newer dart versions
   test: ^1.25.13


lib/src/database/database_file_storage_io.dart
```diff
@@ -1,17 +1,31 @@
...
   late final Uri? fileStorageLocation;
   late final Duration? deleteFilesAfterDuration;
 
-  File _getFileFromMxc(Uri mxcUri) => File(
-        '${Directory.fromUri(fileStorageLocation!).path}/${mxcUri.toString().split('/').last}',
-      );
+  /// Map an MXC URI to a local File path
+  File _getFileFromMxc(Uri mxcUri) {
+    // Encode MXC to a filesystem-safe file name
+    final fileName = base64Url.encode(
```
td-famedly: just using the mxc hash should be enough and fs safe at the same time, any reason to encode it?

huluwa-dev: The uri may contains some characters that will make the file path invalid on Windows.

huluwa-dev: Hi @td-famedly , you are right, a hash is enough, I updated the PR. Thanks.

td-famedly: no as in just the mxcUri, I still you're using a hash, I do not think that is needed https://spec.matrix.org/v1.15/appendices/#opaque-identifiers

huluwa-dev: @td-famedly Without a hash, the file path on Windows might look like this:
`C:\Users\xxxx\AppData\Local\Temp\BiqFAimAAEoceEPuwBqeIykD?width=44&height=44&method=scale&animated=false`


And the app will throw an error:
`[Matrix] Unable to download mxc image - PathNotFoundException: Cannot open file, path = 'C:\Users\72379\AppData\Local\Temp\zrySvzNhuvwoDafHLRtOZaDl?width=44&height=44&method=scale&animated=false' (OS Error: The filename, directory name, or volume label syntax is incorrect.)`

So I believe we should encode or hash it.

td-famedly: Ah then just the mxc itself should be enough? you should just strip any query parameters after `zrySvzNhuvwoDafHLRtOZaDl`

krille-chan: > Ah then just the mxc itself should be enough? you should just strip any query parameters after `zrySvzNhuvwoDafHLRtOZaDl`

We might want to store different versions of the same mxc depending on the width, height, method or animated parameters so I would say this is a valid concern

krille-chan: Maybe an easier idea to prevent using hash algorithm which can be costly in the sum: Just do something like this:

```dart
  File _getFileFromMxc(Uri mxcUri) {
    final fileName =
        '${mxcUri.host.replaceAll('.', '_')}_${mxcUri.pathSegments.join('_')}_${mxcUri.queryParameters.entries.map((entry) => '${entry.key}${entry.value}').join('_')}';

    return File(
      path.join(Directory.fromUri(fileStorageLocation!).path, fileName),
    );
  }
```

This would create a mxc url of `mxc://matrix.org/abcd1234?width=40&height=40` into `matrix_org_abcd1234_width40_height40` which is human readable and should be faster and less complex than creating a hash every time

td-famedly: LG!

huluwa-dev: @td-famedly @krille-chan Thanks, updated.

lib/src/database/database_file_storage_io.dart
```diff
@@ -2,16 +2,27 @@ import 'dart:io';
...
   late final Duration? deleteFilesAfterDuration;
 
-  File _getFileFromMxc(Uri mxcUri) => File(
-        '${Directory.fromUri(fileStorageLocation!).path}/${mxcUri.toString().split('/').last}',
-      );
+  /// Map an MXC URI to a local File path
+  File _getFileFromMxc(Uri mxcUri) {
+    // Replace all special characters with underscores to avoid PathNotFoundException on Windows.
+    final host = mxcUri.host.replaceAll('.', '_');
+    final Path = mxcUri.pathSegments.join('_');
```
krille-chan: Papercut: Variable names should start with a lowercase character

huluwa-dev: @krille-chan Thanks, updated.

krille-chan:
@huluwa-dev thanks, one last step: Can you squash your commits to one signed commit?

huluwa-dev:
> @huluwa-dev thanks, one last step: Can you squash your commits to one signed commit?

Updated.

td-famedly (approved):
lg!

krille-chan:
@huluwa-dev sorry for the hassle but now the analyzer fails because the imports are not sorted. You can fix this by executing `dart run import_sorter:main --no-comments` once

huluwa-dev:
> @huluwa-dev sorry for the hassle but now the analyzer fails because the imports are not sorted. You can fix this by executing `dart run import_sorter:main --no-comments` once

Updated.

krille-chan:
> > @huluwa-dev sorry for the hassle but now the analyzer fails because the imports are not sorted. You can fix this by executing `dart run import_sorter:main --no-comments` once
> 
> Updated.

Well and if you now could squash you commits again as we have the policy to have a clean commit history on the main branch...

huluwa-dev:
> > > @huluwa-dev sorry for the hassle but now the analyzer fails because the imports are not sorted. You can fix this by executing `dart run import_sorter:main --no-comments` once
> > 
> > 
> > Updated.
> 
> Well and if you now could squash you commits again as we have the policy to have a clean commit history on the main branch...

Thanks, I believe it is alright now

PR #2127, opened 2025-08-25
fix: allow setting  keyRingSize for livekit calls (BREAKING CHANGE)

td-famedly:


td-famedly:
overwritten by https://github.com/famedly/matrix-dart-sdk/pull/2130

PR #2129, opened 2025-08-28
feat: add `eventId` parameter to call memberships

Yash-Garg:


lib/src/voip/models/call_membership.dart
```diff
@@ -32,6 +36,7 @@ class CallMembership {
   final String? scope;
   final CallBackend backend;
   final String deviceId;
+  final String? eventId;
```
td-famedly, line 65: why is it not required?

yash-garg: was following [CallMembership.ts](https://github.com/matrix-org/matrix-js-sdk/blob/2f1d654f14be8dd03896e9e76f12017b6f9eec1c/src/matrixrtc/CallMembership.ts#L130-L132), it has `undefined | string` type, so I kept it nullable.

PR #2134, opened 2025-09-02
feat: Implement most of MSC3911

nico-famedly:
This implements uploading restricted media (but not the create media async uploads, since they are not used by the SDK currently).

Additionally this implements using restricted media for setting the room avatar and sending media message.

Currently missing are:
- Attaching media on reactions and editing messages with inline media.

When forwarding events, these should be passed through `Event.copyMediaInContent`. Otherwise no changes to clients should be needed.


td-famedly:
would be nice to have this implemented, but I will close this for now, please reopen it if it's still needed

PR #2135, opened 2025-09-04
add to space - m.space.parent error handling

chagai95:
Adding a log line instead of throwing an error when trying to add a space parent, since adding the child was successful that should be enough


td-famedly:
I will close this for now, please reopen it if it's still needed

chagai95:
image


It's still an issue, can we merge it?

PR #2138, opened 2025-09-10
feat: add support for emoji reactions in calls

coder-with-a-bushido:
Required for https://github.com/famedly/product-management/issues/3265

### DIFF
--- a/lib/matrix_api_lite/model/event_types.dart
+++ b/lib/matrix_api_lite/model/event_types.dart
@@ -114,4 +114,5 @@ abstract class EventTypes {
   static const String GroupCallMemberReplaces = '$GroupCallMember.replaces';
   static const String GroupCallMemberAssertedIdentity =
       '$GroupCallMember.asserted_identity';
+  static const GroupCallMemberReaction = 'com.famedly.call.reaction';
 }
--- a/lib/src/voip/group_call_session.dart
+++ b/lib/src/voip/group_call_session.dart
@@ -57,6 +57,10 @@ class GroupCallSession {
   final CachedStreamController<GroupCallStateChange> onGroupCallEvent =
       CachedStreamController();
 
+  /// The stream is used to signal matrixrtc events for group calls
+  /// Currently, this is done for
+  /// - participants change
+  /// - group call emoji reaction
   final CachedStreamController<MatrixRTCCallEvent> matrixRTCEventStream =
       CachedStreamController();
 
--- a/lib/src/voip/models/matrixrtc_call_event.dart
+++ b/lib/src/voip/models/matrixrtc_call_event.dart
@@ -5,16 +5,40 @@ import 'package:matrix/matrix.dart';
 /// often.
 sealed class MatrixRTCCallEvent {}
 
+/// Event type for participants change
 sealed class ParticipantsChangeEvent implements MatrixRTCCallEvent {}
 
 final class ParticipantsJoinEvent implements ParticipantsChangeEvent {
+  /// The participants who joined the call
   final List<CallParticipant> participants;
 
   ParticipantsJoinEvent({required this.participants});
 }
 
 final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
+  /// The participants who left the call
   final List<CallParticipant> participants;
 
   ParticipantsLeftEvent({required this.participants});
+}
+
+/// Event type for group call emoji reaction update
+final class GroupCallReactionAddedEvent implements MatrixRTCCallEvent {
+  /// The participant who sent the reaction
+  final CallParticipant participant;
+
+  /// The emoji character
+  final String emoji;
+
+  /// Words describing the emoji
+  final String? emojiName;
+
+  final bool isEphemeral;
+
+  GroupCallReactionAddedEvent({
+    required this.participant,
+    required this.emoji,
+    required this.emojiName,
+    required this.isEphemeral,
+  });
 }
--- a/lib/src/voip/voip.dart
+++ b/lib/src/voip/voip.dart
@@ -192,8 +192,24 @@ class VoIP {
     if (event is Event) {
       room = event.room;
 
-      /// this can also be sent in p2p calls when they want to call a specific device
-      remoteDeviceId = event.content.tryGet<String>('invitee_device_id');
+      if (event.type == EventTypes.GroupCallMemberReaction) {
+        final isEphemeral = event.content.tryGet<bool>('is_ephemeral')!;
+        if (isEphemeral &&
+            event.originServerTs.isBefore(
+              DateTime.now()
+                  .subtract(CallConstants.groupCallEphemeralReactionTimeout),
+            )) {
+          Logs().d(
+            '[VOIP] Ignoring ephemeral group call emoji reaction event of type ${event.type} because it is older than ${CallConstants.groupCallEphemeralReactionTimeout}',
+          );
+          return;
+        }
+
+        remoteDeviceId = event.content.tryGet<String>('device_id');
+      } else {
+        /// this can also be sent in p2p calls when they want to call a specific device
+        remoteDeviceId = event.content.tryGet<String>('invitee_device_id');
+      }
     } else if (event is ToDeviceEvent) {
       final roomId = event.content.tryGet<String>('room_id');
       final confId = event.content.tryGet<String>('conf_id');
@@ -211,7 +227,10 @@ class VoIP {
         return;
       }
 
-      if (!event.type.startsWith(EventTypes.GroupCallMemberEncryptionKeys)) {
+      if (!{
+        EventTypes.GroupCallMemberEncryptionKeys,
+        EventTypes.GroupCallMemberEncryptionKeysRequest,
+      }.contains(event.type)) {
         // livekit calls have their own session deduplication logic so ignore sessionId deduplication for them
         final destSessionId = event.content.tryGet<String>('dest_session_id');
         if (destSessionId != currentSessionId) {
@@ -233,23 +252,32 @@ class VoIP {
       return;
     }
 
-    final content = event.content;
-
     if (room == null) {
       Logs().w(
         '[VOIP] _handleCallEvent call event does not contain a room_id, ignoring',
       );
       return;
-    } else if (client.userID != null &&
+    }
+
+    final content = event.content;
+
+    if (client.userID != null &&
         client.deviceID != null &&
         remoteUserId == client.userID &&
         remoteDeviceId == client.deviceID) {
-      Logs().v(
-        'Ignoring call event ${event.type} for room ${room.id} from our own device',
-      );
-      return;
-    } else if (!event.type
-        .startsWith(EventTypes.GroupCallMemberEncryptionKeys)) {
+      // We don't want to ignore group call reactions from our own device because
+      // we want to show them on the UI
+      if (!{EventTypes.GroupCallMemberReaction}.contains(event.type)) {
+        Logs().v(
+          'Ignoring call event ${event.type} for room ${room.id} from our own device',
+        );
+        return;
+      }
+    } else if (!{
+      EventTypes.GroupCallMemberEncryptionKeys,
+      EventTypes.GroupCallMemberEncryptionKeysRequest,
+      EventTypes.GroupCallMemberReaction,
+    }.contains(event.type)) {
       // skip webrtc event checks on encryption_keys
       final callId = content['call_id'] as String?;
       final partyId = content['party_id'] as String?;
@@ -267,8 +295,7 @@ class VoIP {
           );
           return;
         } else if (call != null) {
-          // multiple checks to make sure the events sent are from the the
-          // expected party
+          // multiple checks to make sure the events sent are from the expected party
           if (call.room.id != room.id) {
             Logs().w(
               'Ignoring call event ${event.type} for room ${room.id} claiming to be for call in room ${call.room.id}',
@@ -358,6 +385,9 @@ class VoIP {
           remoteDeviceId!,
           content,
         );
+        break;
+      case EventTypes.GroupCallMemberReaction:
+        await onGroupCallReactionReceived(room, event);
         break;
     }
   }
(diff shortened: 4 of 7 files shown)


lib/msc_extensions/msc_4140_delayed_events/models.dart
```diff
@@ -9,8 +9,9 @@ class ScheduledDelayedEventsResponse {
 
   factory ScheduledDelayedEventsResponse.fromJson(Map<String, dynamic> json) {
     final list = json['delayed_events'] ?? json['scheduled'] as List;
-    final List<ScheduledDelayedEvent> scheduledEvents =
-        list.map((e) => ScheduledDelayedEvent.fromJson(e)).toList();
+    final List<ScheduledDelayedEvent> scheduledEvents = list
```
td-famedly: not needed anymore

lib/src/voip/models/group_call_emoji_reaction.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:matrix/matrix.dart';
+
+sealed class GroupCallEmojiReactionUpdate {
```
td-famedly: is this something which should use MatrixRTCCallEvent?

lib/src/voip/voip.dart
```diff
@@ -721,6 +736,140 @@ class VoIP {
...
+
+    if (eventId == null || eventDeviceId == null) {
+      Logs().w('No event ID or device ID found in reaction content');
+      return;
+    }
+
+    // Get call memberships for the sender
+    final memberships =
+        room.getCallMembershipsForUser(event.senderId, eventDeviceId, this);
+    final membership = memberships.firstWhereOrNull((m) => m.callId == callId);
```
td-famedly: also should check application and scope for sanity

lib/src/voip/voip.dart
```diff
@@ -721,6 +736,140 @@ class VoIP {
...
+
+    if (membership == null) {
+      Logs().w(
+        'No matching membership found for reaction from ${event.senderId}',
+      );
+      return;
+    }
+
+    if (membership.eventId != eventId) {
+      /// If the membership event id doesn't match the event id, we should check
```
td-famedly, line 809: I would rather copy over all the reactions to the new event, should be much more simpler

td-famedly: Atleast for raise hand feature that's what we might go ahead with

lib/src/voip/voip.dart
```diff
@@ -721,6 +736,140 @@ class VoIP {
...
+      emoji: eventEmoji,
+      participant: participant,
+    );
+
+    groupCallEmojiReactionsMap[reaction.participant.id] = eventEmoji;
+    onGroupCallEmojiReaction.add(reaction);
+
+    _groupCallEmojiReactionsRemoverTimer[reaction.participant.id]?.cancel();
+    _groupCallEmojiReactionsRemoverTimer[reaction.participant.id] =
+        Timer(const Duration(milliseconds: 7000), () {
```
td-famedly: should go in CallTimeouts

lib/src/voip/voip.dart
coder-with-a-bushido: > I would rather copy over all the reactions to the new event, should be much more simpler

That was the approach I tried first and I personally think this right here is the simpler approach for 3 reasons:

1. Even if we're sending an event with `m.replace`, we need to do it for the past `n` number of events within a timeframe unlike the hand raise feature which is only about the latest event. 
2. Since reactions are stateless but time-sensitive, the only thing we need to care about is to show the reaction as soon as it is published. It's totally okay if a new member joining the call is not able to see the reaction published a few seconds back because the reaction is usually sent as a reply to something someone said in the call.
3. And finally this is an edge case that happens rarely due to a race condition (a new membership event is set almost exactly at the same time a reaction is sent). We don't want to send multiple edit events just for this.

lib/src/voip/models/group_call_emoji_reaction.dart
coder-with-a-bushido: done

lib/src/voip/voip.dart
coder-with-a-bushido: done

lib/src/voip/voip.dart
coder-with-a-bushido: done

lib/src/voip/voip.dart
```diff
@@ -194,6 +194,16 @@ class VoIP {
 
       /// this can also be sent in p2p calls when they want to call a specific device
       remoteDeviceId = event.content.tryGet<String>('invitee_device_id');
+
+      if (event.type == EventTypes.GroupCallEmojiReaction &&
+          event.originServerTs.isBefore(
+            DateTime.now().subtract(timeouts!.groupCallEmojiReactionLifetime),
+          )) {
+        Logs().w(
```
td-famedly: warning is possibly too high for this, I would say debug level

lib/src/voip/voip.dart
```diff
@@ -240,59 +250,63 @@ class VoIP {
...
-      );
-      return;
-    } else if (!event.type
-        .startsWith(EventTypes.GroupCallMemberEncryptionKeys)) {
-      // skip webrtc event checks on encryption_keys
-      final callId = content['call_id'] as String?;
-      final partyId = content['party_id'] as String?;
-      if (callId == null && event.type.startsWith('m.call')) {
-        Logs().w('Ignoring call event ${event.type} because call_id was null');
+    } else if (!{EventTypes.GroupCallEmojiReaction}.contains(event.type)) {
```
td-famedly: can you add a comment here specifying why we skip the own device events for reaction here? just to make it a bit more readable

lib/src/voip/voip.dart
```diff
@@ -721,6 +738,167 @@ class VoIP {
...
+        Logs().w(
+          '[onGroupCallEmojiReactionReceived] No matching event for event id $eventId found for reaction from ${event.senderId}',
+        );
+        return;
+      }
+
+      final famedlyCallMemberEvent =
+          FamedlyCallMemberEvent.fromJson(referencedEvent, this);
+      final referencedMembership =
+          famedlyCallMemberEvent.memberships.firstWhereOrNull(
```
td-famedly, line 824: should also match scope and application for redundancy

lib/src/voip/voip.dart
```diff
@@ -721,6 +738,167 @@ class VoIP {
...
+        '[onGroupCallEmojiReactionReceived] No matching membership found for reaction from ${event.senderId}',
+      );
+      return;
+    }
+
+    if (membership.eventId != eventId) {
+      /// If the membership event id doesn't match the event id, we should check
+      /// if an old membership event with the same event id exists which is
+      /// replaced by the new one
+      final referencedEvent = await room.getEventById(eventId);
```
td-famedly, line 812: why not get the event using the eventId in the first place? i.e why check the latest member event first?

td-famedly: Second question: what if I get a delayed reaction event and the user had already left the call? wouldn't this add a reaction event to the stream anyway? do we need to make sure that the referenced eventid originTs > the current membership originTs?

lib/src/voip/voip.dart
```diff
@@ -721,6 +738,167 @@ class VoIP {
...
+      userId: event.senderId,
+      deviceId: eventDeviceId,
+    );
+    final reaction = GroupCallEmojiReactionAddedEvent(
+      emoji: eventEmoji,
+      emojiName: eventEmojiName,
+      participant: participant,
+    );
+
+    groupCall.groupCallEmojiReactionsMap[reaction.participant.id] = eventEmoji;
```
td-famedly: why do you need a map and remove timer for this? can we not just react on the GroupCallEmojiReactionAddedEvent event in the client?

lib/src/voip/voip.dart
```diff
@@ -721,6 +738,167 @@ class VoIP {
...
+    );
+  }
+
+  Future<void> sendGroupCallEmojiReaction(
+    String groupCallId,
+    Room room,
+    String emoji,
+    String emojiName,
+  ) async {
+    Logs().i('Emoji reaction selected: $emoji');
```
td-famedly: debug level probably

lib/src/voip/voip.dart
```diff
@@ -721,6 +738,167 @@ class VoIP {
...
+    final memberships =
+        room.getCallMembershipsForUser(client.userID!, client.deviceID!, this);
+    final membership =
+        memberships.firstWhereOrNull((m) => m.callId == groupCallId);
```
td-famedly: needs more matchers, device, application, scope

lib/src/voip/voip.dart
coder-with-a-bushido: done + refactored that part a little more for clarity

lib/src/voip/voip.dart
coder-with-a-bushido: > why not get the event using the eventId in the first place? i.e why check the latest member event first?

We check the latest member first and if the member event is null or expired, we just return so the user who left the call will be ignored.

> Second question: what if I get a delayed reaction event and the user had already left the call? wouldn't this add a reaction event to the stream anyway? do we need to make sure that the referenced eventid originTs > the current membership originTs?

Answered with the first question.

lib/src/voip/voip.dart
coder-with-a-bushido: The `groupCallEmojiReactionsMap` is there for convenience so it's easy to show on UI instead of purely relying on the stream. Do you want me to remove it?

td-famedly: just relying on the stream should be cleaner an d easier IMO, you just listen on matrixRTCEventStream and then switch on the event type

coder-with-a-bushido: I already use that for the floating emoji animation. The map is used for the emoji shown in the participant tile.

lib/matrix_api_lite/model/event_types.dart
```diff
@@ -114,4 +114,5 @@ abstract class EventTypes {
   static const String GroupCallMemberReplaces = '$GroupCallMember.replaces';
   static const String GroupCallMemberAssertedIdentity =
       '$GroupCallMember.asserted_identity';
+  static const GroupCallEmojiReaction = 'com.famedly.call.reaction';
```
td-famedly: probably makes sense to keep it aligned with the rest so $GroupCallMember.reaction

lib/src/voip/group_call_session.dart
```diff
@@ -57,9 +57,19 @@ class GroupCallSession {
...
 
+  /// The stream is used to signal matrixrtc events for group calls
+  /// Currently, this is done for
+  /// - participants change
+  /// - group call emoji reaction
   final CachedStreamController<MatrixRTCCallEvent> matrixRTCEventStream =
       CachedStreamController();
 
+  /// Map of group call emoji reactions by participant id
+  final Map<String, String> groupCallEmojiReactionsMap = {};
```
td-famedly: these can be removed because it's easy to just listen to the streams and show reactions if you get events on a stream, no need to worry about clearing queues

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -5,16 +5,45 @@ import 'package:matrix/matrix.dart';
...
-sealed class ParticipantsChangeEvent implements MatrixRTCCallEvent {}
-
-final class ParticipantsJoinEvent implements ParticipantsChangeEvent {
+/// Event type for participants change
+sealed class ParticipantsChangeEvent implements MatrixRTCCallEvent {
+  /// The participants who joined or left the call
   final List<CallParticipant> participants;
 
-  ParticipantsJoinEvent({required this.participants});
+  ParticipantsChangeEvent({required this.participants});
```
td-famedly: because ParticipantsChangeEvent is sealed and you can't instantiate it doesn't make sense to have parameters in it IMO, we should keep them in join and leave classes

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -5,16 +5,45 @@ import 'package:matrix/matrix.dart';
...
+final class ParticipantsLeftEvent extends ParticipantsChangeEvent {
+  ParticipantsLeftEvent({required super.participants});
+}
+
+/// Event type for group call emoji reaction update
+sealed class GroupCallEmojiReactionEvent implements MatrixRTCCallEvent {
+  GroupCallEmojiReactionEvent({required this.participant});
+
+  /// The participant who sent the reaction
+  final CallParticipant participant;
```
td-famedly, line 28: same over here, cannot instantiate `GroupCallEmojiReactionEvent`

lib/src/voip/models/matrixrtc_call_event.dart
td-famedly: This allows me to trigger the event without any parameters, but I guess you could always have a empty list as well so it's a matter of taste?

lib/src/voip/voip.dart
td-famedly: I am thinking about a more general implementation of reactions which can be used for raising hands as well. Here's how I think it should look like

- We send reactions with the custom event type in the room
- We mark reactions as ephemeral in the content body and hand raises as permanent
- On member update refreshes we copy over any permanent reactions from the current membership event (usually this will only be the raise hands reaction)
- if you get a reaction on a not latest member event you ignore it, because it should have been copied over on the update anyway so it should arrive in a few seconds if it's a permanent type of reaction. 
- We listen on the rtcEvents stream for reactions, if it's a ephemeral reaction we show it as a reaction and then forget about it, if it's a perm reaction the only thing we have to do is copy it over to a new membership event on refresh.


Open to opinions if these should be 2 different events or the same event


(Alternative to marking eph/perm would be using delayed events to delete a reaction after 3 seconds or so, but that would junk up the timeline a lot so it should be better to just say so in the content body)

cc: @Yash-Garg

coder-with-a-bushido: Agree with other things but I think they should be 2 different events as one is a reaction from the user for something in the call and the other one is to show "intent to speak". Even element is doing the same -> https://github.com/element-hq/element-call/pull/2708

Also, I don't want to ignore the reaction if the event id doesn't match with latest membership event as this could be race condition rarely. We do enough checks before we reach that point to check older membership events in this logic, so it should be fine the way it is now as that code won't be executed needlessly. Wdyt @td-famedly ?

td-famedly: > Agree with other things but I think they should be 2 different events as one is a reaction from the user for something in the call and the other one is to show "intent to speak". Even element is doing the same -> https://github.com/element-hq/element-call/pull/2708

I would like to reduce on the code duplication, both events are nearly identical except for the permanent/ephemeral flag. Is there any other reason apart from element doing it that we should go with 2 different event types?

> Also, I don't want to ignore the reaction if the event id doesn't match with latest membership event as this could be race condition rarely. We do enough checks before we reach that point to check older membership events in this logic, so it should be fine the way it is now as that code won't be executed needlessly. Wdyt @td-famedly ?

I am fine with this if we can use it for the raise hands feature as well. So we first check if the event is on the current state, if not we check if there is a non expired membership, we then check the referenced event id, if the content is permanent we show the raise hand, if it's ephemeral but the sent time was less than 2 seconds ago or so we show the reaction, we ignore all other ephemeral reactions because those could be very old. Does that make sense? @coder-with-a-bushido

yash-garg: > I would like to reduce on the code duplication, both events are nearly identical except for the permanent/ephemeral flag. Is there any other reason apart from element doing it that we should go with 2 different event types?

I like this approach as well, there aren't much differences - one common event can be used for both instead of separating it out.

coder-with-a-bushido: @td-famedly Would an extra field in the event content work? Do you have any other specific detail on how to implement it?

Also, would 2 seconds be enough for the ephemeral event age check?

td-famedly: @coder-with-a-bushido just the extra field should be enough IMO and 2 seconds for ephemeral makes sense :D

coder-with-a-bushido: Done!

@Yash-Garg and I discussed about this and we decided to hold on merging this until the hand raise PR is also rebased on these changes.

lib/matrix_api_lite/model/event_types.dart
```diff
@@ -114,4 +114,5 @@ abstract class EventTypes {
   static const String GroupCallMemberReplaces = '$GroupCallMember.replaces';
   static const String GroupCallMemberAssertedIdentity =
       '$GroupCallMember.asserted_identity';
+  static const GroupCallMemberReaction = 'com.famedly.call.reaction';
```
td-famedly, line 117: could be `com.famedly.call.member.reaction` for consistency

lib/src/voip/voip.dart
```diff
@@ -192,8 +192,24 @@ class VoIP {
     if (event is Event) {
       room = event.room;
 
-      /// this can also be sent in p2p calls when they want to call a specific device
-      remoteDeviceId = event.content.tryGet<String>('invitee_device_id');
+      if (event.type == EventTypes.GroupCallMemberReaction) {
+        final isEphemeral = event.content.tryGet<bool>('is_ephemeral')!;
+        if (isEphemeral &&
```
td-famedly, line 197: wouldn't this miss permanent reactions like hand raises? or is that handled in the other PR then?

lib/src/voip/voip.dart
```diff
@@ -721,6 +751,167 @@ class VoIP {
...
+        '[onGroupCallReactionReceived] No matching membership found or found expired for reaction from ${event.senderId}',
+      );
+      return;
+    }
+
+    if (membership.eventId != eventId) {
+      /// If the membership event id doesn't match the event id, we should check
+      /// if an old membership event with the same event id exists which is
+      /// replaced by the new one
+      final referencedEvent = await room.getEventById(eventId);
```
td-famedly, line 812: we talked about removing this because it might be more error prone than expected. for permanent reactions we will copy over the reactions and for eph ones we just ignore them if not on the latest member event. 

The current member updates are an hour or so so you the race condition should be really rare given that you use delayed events and actually increase the member event update interval

PR #2144, opened 2025-09-13
feat: implement reaction events for calls

yash-garg:


lib/src/voip/group_call_session.dart
```diff
@@ -272,4 +274,204 @@ class GroupCallSession {
...
+    required String key,
+    String? deviceId,
+  }) async {
+    final memberships = room.getCallMembershipsForUser(
+      userId,
+      deviceId ?? '',
+      voip,
+    );
+
+    final membership = memberships.firstOrNull;
```
td-famedly, line 295: needs a application and scope match for sanity

lib/src/voip/group_call_session.dart
```diff
@@ -272,4 +274,204 @@ class GroupCallSession {
...
+  /// Handle incoming reaction events from Matrix
+  Future<void> onReactionReceived(
+    BasicEventWithSender event,
+    String reactionKey,
+    String eventId,
+  ) async {
+    final participant = CallParticipant(
+      voip,
+      userId: event.senderId,
+      deviceId: null, // We might not have device ID from the reaction
```
td-famedly, line 435: the other reactions PR makes a custom event for this, I think we can go with that

PR #2148, opened 2025-09-18
refactor: BREAKING make onGroupCallEvent return actual events type

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/1950


lib/src/voip/models/group_call_state.dart
```diff
@@ -0,0 +1,71 @@
+import 'package:matrix/src/voip/utils/types.dart';
+
+sealed class GroupCallEvent {}
```
td-famedly: I think they should all extend the matrtcRTC stream

coder-with-a-bushido: Can it also just use that matrixrtc stream so we can remove onGroupEvent and onGroupStateChange?

td-famedly: sounds good to me!

coder-with-a-bushido: DONE!!!

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -133,7 +133,7 @@ class MeshBackend extends CallBackend {
   Future<void> _addCall(GroupCallSession groupCall, CallSession call) async {
     _callSessions.add(call);
     _initCall(groupCall, call);
-    groupCall.onGroupCallEvent.add(GroupCallStateChange.callsChanged);
+    groupCall.matrixRTCEventStream.add(GroupCallAddedEvent());
```
td-famedly: This is easy to have backwards compatibility with to make lives of client devs a bit easier, can we keep both of them with a notice?

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -375,7 +375,7 @@ class MeshBackend extends CallBackend {
 
     if (nextActiveSpeaker != null && _activeSpeaker != nextActiveSpeaker) {
       _activeSpeaker = nextActiveSpeaker;
-      groupCall.onGroupCallEvent.add(GroupCallStateChange.activeSpeakerChanged);
+      groupCall.matrixRTCEventStream.add(GroupCallActiveSpeakerChanged());
```
td-famedly: this could send data along as in who the active speaker now is

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -196,7 +196,7 @@ class MeshBackend extends CallBackend {
 
     _callSessions.removeWhere((element) => call.callId == element.callId);
 
-    groupCall.onGroupCallEvent.add(GroupCallStateChange.callsChanged);
+    groupCall.matrixRTCEventStream.add(GroupCallRemovedEvent());
```
td-famedly: this should give indication of what group call was removed, same with added and replaced

lib/src/voip/backend/mesh_backend.dart
```diff
@@ -401,8 +401,8 @@ class MeshBackend extends CallBackend {
   ) {
     _screenshareStreams.add(stream);
     onStreamAdd.add(stream);
-    groupCall.onGroupCallEvent
-        .add(GroupCallStateChange.screenshareStreamsChanged);
+    groupCall.matrixRTCEventStream
+        .add(GroupCallScreenShareStreamsChanged(GroupCallStreamsChange.added));
```
td-famedly: should be easy to also add info like it was the screen sharing stream that was added/removed

lib/src/voip/backend/mesh_backend.dart
coder-with-a-bushido: that would mean we keep GroupCallStateChange enum, do you want that?

coder-with-a-bushido: done

td-famedly (changes requested):
ah @coder-with-a-bushido can you also add some tests please?

PR #2149, opened 2025-09-18
feat: Add onProgress for upload and download methods

krille-chan:
Closes https://github.com/famedly/matrix-dart-sdk/issues/290

Tested with FluffyChat. I was able to display progress of uploading and downloading files.


lib/src/client.dart
```diff
@@ -1523,6 +1524,7 @@ class Client extends MatrixApi {
     Uint8List file, {
     String? filename,
     String? contentType,
+    void Function(int)? onProgress,
```
coder-with-a-bushido, line 1530: can we also add a comment for the `onProgress` parameter above?

coder-with-a-bushido: A small example in the documentation would be really great too!

krille-chan: done :)

PR #2150, opened 2025-09-19
feat: Leave DM rooms and invite when ignoring a user

krille-chan:


lib/src/client.dart
```diff
@@ -3771,10 +3771,30 @@ class Client extends MatrixApi {
 
   /// Ignore another user. This will clear the local cached messages to
   /// hide all previous messages from this user.
-  Future<void> ignoreUser(String userId) async {
+  Future<void> ignoreUser(
+    String userId, {
+    /// Wether to also decline all invites and leave DM rooms with this user.
```
coder-with-a-bushido:
```suggestion
    /// Whether to also decline all invites and leave DM rooms with this user.
```

krille-chan: Thanks :)

PR #2152, opened 2025-09-19
feat: implement reaction events for calls

yash-garg:


lib/src/voip/voip.dart
```diff
@@ -192,8 +192,24 @@ class VoIP {
     if (event is Event) {
       room = event.room;
 
-      /// this can also be sent in p2p calls when they want to call a specific device
-      remoteDeviceId = event.content.tryGet<String>('invitee_device_id');
+      if (event.type == EventTypes.GroupCallMemberReaction ||
+          event.type == EventTypes.GroupCallMemberReactionRemoved) {
+        final isEphemeral = event.content.tryGet<bool>('is_ephemeral')!;
+        if (isEphemeral &&
```
td-famedly, line 197: won't this skip permanent reactions then?

yash-garg: nope, those are marked as `false` right? It does work fine as of now 🤔

td-famedly: ah it's an early return not the other way around ok!

lib/src/voip/voip.dart
```diff
@@ -721,6 +756,130 @@ class VoIP {
...
+    );
+
+    if (membership == null || membership.isExpired) {
+      Logs().w(
+        '[VOIP] _handleReactionEvent: No matching membership found or found expired for reaction from ${event.senderId}',
+      );
+      return;
+    }
+
+    if (membership.eventId != eventId) {
```
td-famedly: this was proposed to be removed, we only want to check the reactions on the latest events

lib/src/voip/group_call_session.dart
```diff
@@ -151,6 +153,29 @@ class GroupCallSession {
...
+    // Get current member event ID to preserve permanent reactions
+    final currentMemberships = room.getCallMembershipsForUser(
+      client.userID!,
+      client.deviceID!,
+      voip,
+    );
+
+    final currentMembership = currentMemberships.firstWhereOrNull(
+      (m) =>
+          m.callId == groupCallId &&
```
td-famedly, line 174: doesn't this also need a deviceId check? 🤔

td-famedly: ah currentmembership is for the deviceId ignore

td-famedly: actually do not ignore, because without 3757 it would still return all deviceIds. So please also add a deviceId check here and everywhere else where you do these kinds of checks

lib/src/voip/group_call_session.dart
```diff
@@ -169,6 +194,13 @@ class GroupCallSession {
       ),
     );
 
+    // Copy permanent reactions to the new member event
+    if (permanentReactions.isNotEmpty) {
+      // Add a small delay to ensure the new member event is processed
+      await Future.delayed(const Duration(milliseconds: 100));
```
td-famedly: do you need to wait for the timeline to have the event here? if you get a eventId back, the server should have it and you should be able to just send reactions to that eventId

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +303,269 @@ class GroupCallSession {
...
+    final reactionEvent = await _buildReactionEvent(
+      emoji: emoji,
+      name: name,
+      isEphemeral: false,
+    );
+
+    if (reactionEvent == null) return null;
+
+    // Send reaction removal as unencrypted event to avoid decryption issues
+    final txid = client.generateUniqueTransactionId();
```
td-famedly: why is this a special event and not just a redaction? 🤔  Does this actually redact anything?

lib/src/voip/group_call_session.dart
```diff
@@ -151,6 +153,30 @@ class GroupCallSession {
...
+          m.deviceId == client.deviceID! &&
+          m.application == application &&
+          m.scope == scope &&
+          m.roomId == room.id,
+    );
+
+    // Store permanent reactions from the current member event if it exists
+    List<MatrixEvent> permanentReactions = [];
+    if (currentMembership?.eventId != null) {
+      permanentReactions = await _getPermanentReactionsForEvent(
```
td-famedly, line 186: You also need to make sure that the currentMembership is not expired

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+    required String emoji,
+    required String name,
+    bool isEphemeral = true,
+  }) async {
+    Logs().d('Group call reaction selected: $emoji');
+    final memberships =
+        room.getCallMembershipsForUser(client.userID!, client.deviceID!, voip);
+    final membership = memberships.firstWhereOrNull(
+      (m) =>
+          m.callId == groupCallId &&
```
td-famedly, line 343: needs deviceId and roomId checks

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
       onGroupCallEvent.add(GroupCallStateChange.participantsChanged);
     }
   }
+
+  Future<Map<String, dynamic>?> _buildReactionEvent({
```
td-famedly: Does this need to be async?

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+  /// [emoji] - The reaction emoji to filter by (e.g., '🖐️')
+  ///
+  /// Returns a list of [MatrixEvent] objects representing the reactions
+  Future<List<MatrixEvent>> getAllReactions({required String emoji}) async {
+    final reactions = <MatrixEvent>[];
+
+    for (final participant in participants) {
+      final memberships = room.getCallMembershipsForUser(
+        participant.userId,
+        participant.deviceId ?? '',
```
td-famedly: if the participant deviceId is null just ignore that participant, makes no sense for group calls if that's the case

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+
+      for (final membership in membershipsForCurrentGroupCall) {
+        if (membership.eventId == null) {
+          Logs().w(
+            'Cannot find membership event for ${participant.userId}',
+          );
+          continue;
+        }
+
+        final events = await client.getRelatingEventsWithRelTypeAndEventType(
```
td-famedly: so if there are 40 participants, this would do 40 network requests, I don't think that's a good idea. We should use the timeline to get these events first and then backfill once we hit the end of the timeline.

Can you check how element is using this API?

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+  ) async {
+    // Get the new member event ID
+    final updatedMemberships = room.getCallMembershipsForUser(
+      client.userID!,
+      client.deviceID!,
+      voip,
+    );
+    final updatedMembership = updatedMemberships.firstWhereOrNull(
+      (m) =>
+          m.callId == groupCallId &&
```
td-famedly: deviceId check

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+
+        if (reactionKey == null) {
+          Logs().w(
+            '[VOIP] Skipping permanent reaction copy: missing reaction key',
+          );
+          continue;
+        }
+
+        // Build new reaction event with updated event ID
+        final newReactionEvent = {
```
td-famedly: This should probably have it's own model. Typing it out everytime is not a good idea

lib/src/voip/voip.dart
```diff
@@ -721,6 +753,135 @@ class VoIP {
...
+    final participant = CallParticipant(
+      this,
+      userId: event.senderId,
+      deviceId: deviceId,
+    );
+
+    final reaction = ReactionAddedEvent(
+      participant: participant,
+      reactionKey: reactionKey,
+      reactionName: reactionName,
```
td-famedly: what is the reaction name used for? Do we need this field at all?

lib/src/voip/voip.dart
```diff
@@ -721,6 +753,135 @@ class VoIP {
...
+      (groupCall) =>
+          groupCall.room.id == room.id &&
+          groupCall.state == GroupCallState.entered,
+    );
+
+    for (final groupCall in activeGroupCalls) {
+      final participant = CallParticipant(
+        this,
+        userId: event.senderId,
+        deviceId: null,
```
td-famedly: the content should get you the deviceId right?

yash-garg: No, the redaction event doesn't send the device id - we don't use a custom event for this.

td-famedly: Hm probably not. Maybe abuse the reason key, pass in the deviceId there, check if the event was actually a com.famedly.member.reaction and then form the participant?

td-famedly: alternatively check how element does it

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +302,253 @@ class GroupCallSession {
...
+      reactionEvent,
+    );
+  }
+
+  /// Remove a reaction event from the group call
+  ///
+  /// [eventId] - The event ID of the reaction to remove
+  ///
+  /// Returns the event ID of the removed reaction event
+  Future<String?> removeReactionEvent({
```
td-famedly: also add a check that you are indeed removing a com.famedly.call.member.reaction event here for sanity

lib/src/voip/voip.dart
yash-garg: @coder-with-a-bushido added it, but I don't see usage of it anywhere - I'll remove it.

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -18,3 +18,73 @@ final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
```
td-famedly, line 18: the event_id in callmembership shouldn't be in the toJson function, this is causing it to be pushed as null in the room state always, can you remove this please?

lib/src/voip/group_call_session.dart
```diff
@@ -271,4 +304,247 @@ class GroupCallSession {
...
+      );
+    }
+
+    return permanentReactions;
+  }
+
+  /// Copy permanent reactions to the new member event
+  ///
+  /// [permanentReactions] - List of permanent reaction events to copy
+  Future<void> _copyPermanentReactionsToNewEvent(
```
td-famedly, line 493: The copy logic is not working because the new reaction event still relates to the old member event, you should update the relation on copy.

This also causes another bug where if you join with a new client, old reactions from other participants are not visible

lib/src/voip/voip.dart
```diff
@@ -359,6 +385,12 @@ class VoIP {
           content,
         );
         break;
+      case EventTypes.GroupCallMemberReaction:
+        await _handleReactionEvent(room, event);
+        break;
+      case EventTypes.Redaction:
+        await _handleRedactionEvent(room, event);
```
td-famedly, line 395: how do you stop other redactions from entering this function? 🤔  maybe we need to mention in the reason field that it was actually a call reaction redaction?

td-famedly (commented):
probably also want to remove the permanent reactions on leave call? or removeMemberEvent?

lib/matrix_api_lite/generated/api.dart
```diff
@@ -4581,6 +4582,7 @@ class Api {
     request.bodyBytes = utf8.encode(
       jsonEncode({
         if (reason != null) 'reason': reason,
+        if (deviceId != null) 'device_id': deviceId,
```
td-famedly: no this is the generated code you shouldn't modify it, I would suggest overriding it.

But you can also try just using the contentKeyWhiteList on the client side and see if that works

lib/src/room.dart
```diff
@@ -2385,6 +2385,7 @@ class Room {
   /// Redacts this event. Throws `ErrorResponse` on error.
   Future<String?> redactEvent(
     String eventId, {
+    String? deviceId,
```
coder-with-a-bushido: how are we using this deviceId?

lib/src/client.dart
```diff
@@ -4062,6 +4062,59 @@ class Client extends MatrixApi {
...
+  /// [txnId] The [transaction ID](https://spec.matrix.org/unstable/client-server-api/#transaction-identifiers) for this event. Clients should generate a
+  /// unique ID; it will be used by the server to ensure idempotency of requests.
+  ///
+  /// [reason] The reason for the event being redacted.
+  ///
+  /// [metadata] is a map which will be expanded and sent along the reason field
+  ///
+  /// returns `event_id`:
+  /// A unique identifier for the event.
+  Future<String?> redactEventWithMetadata(
```
coder-with-a-bushido, line 4068: may i ask how this method is different from the redact method we already have in the sdk?

coder-with-a-bushido: should we not just add the metadata field to it then?

coder-with-a-bushido: I'm sorry if I'm missing something but I just don't see the metadata field in the client server spec. I thought we are doing a custom implementation but I don't see the metadata used anywhere else in this PR

coder-with-a-bushido: my bad, it's inside the content!

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -18,3 +18,73 @@ final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
 
   ParticipantsLeftEvent({required this.participants});
 }
+
+sealed class ReactionEvent implements MatrixRTCCallEvent {}
```
coder-with-a-bushido: just a suggestion but can we have it called `CallReactionEvent` instead for clarity?

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -18,3 +18,73 @@ final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
...
+
+final class ReactionAddedEvent implements ReactionEvent {
```
coder-with-a-bushido: same here ?

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -18,3 +18,73 @@ final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
...
+  ReactionAddedEvent({
+    required this.participant,
+    required this.reactionKey,
+    required this.membershipEventId,
+    required this.reactionEventId,
+    required this.isEphemeral,
+  });
+}
+
+final class ReactionRemovedEvent implements ReactionEvent {
```
coder-with-a-bushido: and here

lib/src/voip/models/matrixrtc_call_event.dart
```diff
@@ -18,3 +18,73 @@ final class ParticipantsLeftEvent implements ParticipantsChangeEvent {
...
+  final CallParticipant participant;
+  final String redactedEventId;
+
+  ReactionRemovedEvent({
+    required this.participant,
+    required this.redactedEventId,
+  });
+}
+
+final class ReactionPayload {
```
coder-with-a-bushido: and here

coder-with-a-bushido (approved):
LGTM

lib/src/voip/models/matrixrtc_call_event.dart
td-famedly: isn't exposed so we don't need to

coder-with-a-bushido:
Can we clean up the commits too before merging?

Issue #1991, opened 2025-01-01
Compatibility Issue with WebAssembly (WASM)


tusharbhambere:
### Checklist

- [x] I could not find a solution in the documentation, the existing issues or discussions.
- [ ] I already asked for help in the chat

### On which platform did the bug appear?

Chrome

### SDK Version

matrix: ^0.36.0

### Describe the problem caused by this bug

When building a Flutter web application with the --wasm flag, I encountered a compatibility issue involving the matrix-dart-sdk. The error occurs during the build process and appears to stem from the SDK relying on the dart:html package, which is incompatible with WASM builds.

```
main.dart => package:qviple_user_desktop => package:matrix => dart:html  

```
This indicates that the matrix-dart-sdk has a hard dependency on dart:html, which prevents successful compilation when targeting WebAssembly.

Supporting WASM in the matrix-dart-sdk would unlock significant potential for Flutter web applications targeting modern web standards and optimized performance.

### Steps To Reproduce

1.Create a Flutter web project with a dependency on matrix-dart-sdk.
2.Run the following command:
flutter build web --wasm

PR #2153, opened 2025-09-24
refactor: migrate to web and js_interop pkgs

coder-with-a-bushido:
Closes https://github.com/famedly/matrix-dart-sdk/issues/1991


krille-chan (approved):
Works in FluffyChat

PR #2154, opened 2025-09-24
refactor: Support matrix spec 1.16

krille-chan:


analysis_options.yaml
```diff
@@ -47,4 +47,5 @@ analyzer:
     todo: ignore
   exclude:
     - example/main.dart
+    - lib/matrix_api_lite/generated/**
```
td-famedly: these can usually be fixed, any reason why they are excluded now?

td-famedly (dismissed):

analysis_options.yaml
krille-chan: It now clashes with missing trailing commas which is not fixed by a dart format lib

krille-chan: Okay I think I fixed it. However the clean up script isn't working good anyway.

PR #2155, opened 2025-09-28
refactor: Upgrade to vodozemac cryptoutils

krille-chan:


lib/encryption/ssss.dart
```diff
@@ -242,7 +246,7 @@ class SSSS {
             info: content.passphrase!,
           ),
         ),
-      ).timeout(Duration(seconds: 10));
+      ).timeout(Duration(minutes: 2));
```
coder-with-a-bushido: do we really need 2 minutes here? If yes, is there such a big performance difference that we should look into?

krille-chan: I'll check

krille-chan: Reverted this

PR #2169, opened 2025-10-19
fix: Set join rules with knowk_restricted and multiple allow condition room ids

krille-chan:


lib/src/room.dart
```diff
@@ -2437,18 +2437,30 @@ class Room {
...
       EventTypes.RoomJoinRules,
       '',
       {
-        'join_rule': joinRules.toString().replaceAll('JoinRules.', ''),
-        if (allowConditionRoomId != null)
-          'allow': [
-            {'room_id': allowConditionRoomId, 'type': 'm.room_membership'},
-          ],
+        'join_rule': joinRules.text,
+        if (allowConditionRoomIds != null)
```
coder-with-a-bushido: can we also check if the list is empty here?

krille-chan: Done :)

PR #2171, opened 2025-10-20
refactor: Simpler update user device keys

krille-chan:
fixes https://github.com/famedly/product-management/issues/3471

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2235,8 +2235,8 @@ class Client extends MatrixApi {
       await dispose();
     }
 
-    _id = accessToken = _syncFilterId =
-        homeserver = _userID = _deviceID = _deviceName = _prevBatch = null;
+    _id = accessToken = _syncFilterId = homeserver =
+        _userID = _deviceID = _deviceName = _prevBatch = _trackedUserIds = null;
     _rooms = [];
     _eventsPendingDecryption.clear();
     await encryption?.dispose();
@@ -2547,6 +2547,7 @@ class Client extends MatrixApi {
       for (final userId in deviceLists.left ?? []) {
         if (_userDeviceKeys.containsKey(userId)) {
           _userDeviceKeys.remove(userId);
+          _trackedUserIds?.remove(userId);
         }
       }
     }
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -3058,10 +3069,20 @@ class Client extends MatrixApi {
...
+        if (stateKey != null &&
             (!room.partial || importantStateEvents.contains(event.type))) {
           room.setState(event);
+
+          if (room.encrypted &&
+              event.type == EventTypes.RoomMember &&
+              {'join', 'invite'}
+                  .contains(event.content.tryGet<String>('membership'))) {
+            // New members should be added to the tracked user IDs for encryption:
+            _trackedUserIds?.add(stateKey);
```
techno-disaster, line 3082: do we not have to update that on leave?

lib/src/client.dart
```diff
@@ -3213,13 +3232,15 @@ class Client extends MatrixApi {
...
+  Set<String>? _trackedUserIds;
 
   Future<void> updateUserDeviceKeys({Set<String>? additionalUsers}) async {
     try {
       final database = this.database;
       if (!isLogged()) return;
       final dbActions = <Future<dynamic> Function()>[];
-      final trackedUserIds = await _getUserIdsInEncryptedRooms();
+      final trackedUserIds =
+          _trackedUserIds ??= await _getUserIdsInEncryptedRooms();
```
techno-disaster, line 3248: looks like it will use the original function on every init then, do we not want to persist this across restarts?

lib/src/client.dart
krille-chan: no because the user could be in a different room still. We update it in `Client._handleDeviceListsEvents()` because there we get notified from sync if we no longer share an encrypted room with this user.

lib/src/client.dart
krille-chan: Not yet. I intended to make the change as less intrusive as possible so we still use this function which works, but use it much less often. A next refactoring step could be to not update user device keys in sync at all but only if we need them like when sending an encrypted event. But that would be a very huge refactoring. I intended to just fix the gigantic performance leak for now

lib/src/client.dart
```diff
@@ -3213,13 +3232,15 @@ class Client extends MatrixApi {
   }
 
   final Map<String, DateTime> _keyQueryFailures = {};
+  Set<String>? _trackedUserIds;
```
nico-famedly, line 3240: As someone, who hasn't touched this code in quite a while, I think it would be good to add more explanations to the code. For example what this variable is for. I don't think it is quite clear without context, what "tracking" means here and might confuse someone who doesn't have the full context. It would also help me to understand, what the intent of this tracking is, if it was described somewhere instead of defined implicitly in the code. :)

lib/src/client.dart
nico-famedly: Would it be a huge refactor? Imo you would just need to make https://github.com/famedly/matrix-dart-sdk/blob/96f1f056e981ba464e8977b6f26cbe0e2f3495d6/lib/src/room.dart#L2550 be aware of outdated device keys (i.e. by deleting them for the user). Then it should basically magically work. In sync you could then just do a proactive fetch of the device keys, but tbh, it isn't really necessary.
The biggest issue might be telling a potential UI, that it needs to call the `getUserDeviceKeys` function again.

That would be an actual simplification of the code imo. As it stands, I think the title for this PR is wrong, but it does work, so it is probably good to merge as is as well.

nico-famedly, line 3248: Is this potentially racy between a user signing out and this field being set? In theory the logout could be quicker, so the field could be set even after logout, which might lead to weird crypto issues?

krille-chan: Maybe not a huge refactor in code size but I'm just scared to change the behavior here 😅 Right now the change is not that intrusive and I think the chance it breaks something is very small. I'm using it as a daily driver on my phone and in web for over a week now and had not a single problem

lib/src/client.dart
krille-chan: Good catch. I will add some dartdoc documentation :)

lib/src/client.dart
krille-chan: Good point. We could in theory get logged out while _getUserIdsInEncryptedRooms() started. That would be very unlikely but not impossible. Right after the call we do another ´isLogged()` check. I added, that in this case we set `_trackedUserIds = null;` so we should be safe now.

lib/src/client.dart
td-famedly: oki

td-famedly (commented):
can this be tested in unit tests easily?

PR #2175, opened 2025-10-31
fix: Do not retry profile request if error is any server message

krille-chan:
We can assume that the server does not
return the profile for a reason. However
if we have a retry after ms field we can
unblock the retry after waiting.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1064,11 +1064,20 @@ class Client extends MatrixApi {
           super.getUserProfile(userId).timeout(timeout));
     } catch (e) {
       Logs().d('Unable to fetch profile from server', e);
-      if (cachedProfile == null) rethrow;
-      return cachedProfile;
-    } finally {
-      unawaited(_userProfileRequests.remove(userId));
+      if (cachedProfile != null) return cachedProfile;
+      if (e is MatrixException) {
+        final retryAfterMs = e.retryAfterMs;
+        if (retryAfterMs != null) {
+          await Future.delayed(Duration(milliseconds: retryAfterMs));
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -1064,11 +1064,20 @@ class Client extends MatrixApi {
...
     } catch (e) {
       Logs().d('Unable to fetch profile from server', e);
-      if (cachedProfile == null) rethrow;
-      return cachedProfile;
-    } finally {
-      unawaited(_userProfileRequests.remove(userId));
+      if (cachedProfile != null) return cachedProfile;
+      if (e is MatrixException) {
+        final retryAfterMs = e.retryAfterMs;
+        if (retryAfterMs != null) {
```
techno-disaster, line 1070: Do we not want to have a global retry mechanism for all kinds of matrix api requests then? :3

krille-chan: nah, this depends on the specific use case I would say

td-famedly: Is there a specific reason why we do the retry like this for profile requests and not for other requests?

krille-chan: because we cache it. Right now we retry it automatically for any error so it makes sense to interrupt here when we run intro a rate limit

td-famedly: hmm oki

td-famedly (commented):
I think we can easily have a test for this, do you want to add it now?

td-famedly:
@krille-chan can you give this some love?

krille-chan:
Doesn't happen for me since a while so I think we can close this

PR #2179, opened 2025-11-05
fix: Replies in threads

krille-chan:
This fixes the situation that an event can be
a reply and in a thread. Before we have seen reply
as an relationshipType but this does conflict with
the concept of threads, where an event can be
of relationship type "thread" but also be a reply
with being a fallback or not.

## Some thoughts:

We have `Event.getReplyEvent()` and `Event.relationshipType`. And there are three different cases:

1. Normal replies
2. Thread events with reply fallback (`is_falling_back = true`)
3. Reply in a thread (`is_falling_back = false`)

##### Expected return for `Event.relationshipType`:
1. `RelationshipType.reply`
2. `RelationshipType.thread`
3. What should this be? Actually it is both. So maybe we should remove `RelationshipType.reply` and replace it with a `Event.isReply`

We should remove the `RelationshipType.reply` which is actually not a relationshiptype according to the matrix spec. Instead we should add a new getter `Event.inReplyToEventId({bool includingFallback = true})` where we can also say if we want to include fallbacks or not.

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -32,7 +32,6 @@ import 'package:matrix/src/utils/markdown.dart';
 import 'package:matrix/src/utils/multipart_request_progress.dart';
 
 abstract class RelationshipTypes {
-  static const String reply = 'm.in_reply_to';
   static const String edit = 'm.replace';
   static const String reaction = 'm.annotation';
   static const String reference = 'm.reference';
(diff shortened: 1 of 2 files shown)


lib/src/event.dart
```diff
@@ -1027,6 +1009,11 @@ class Event extends MatrixEvent {
     if (mRelatesTo == null) {
       return null;
     }
+    if (mRelatesTo.containsKey('m.in_reply_to') &&
```
td-famedly: this looks like duplicated code from one line below, can we somehow merge these 2 cases?

lib/src/event.dart
```diff
@@ -32,7 +32,6 @@ import 'package:matrix/src/utils/markdown.dart';
 import 'package:matrix/src/utils/multipart_request_progress.dart';
 
 abstract class RelationshipTypes {
-  static const String reply = 'm.in_reply_to';
```
td-famedly, line 35: so a breaking change?

lib/src/event.dart
```diff
@@ -1021,30 +1002,30 @@ class Event extends MatrixEvent {
...
-            ?.tryGet<String>('event_id');
+  /// Get the relationship type of an event. `null` if there is none.
+  String? get relationshipType => content
+      .tryGetMap<String, Object?>('m.relates_to')
+      ?.tryGet<String>('rel_type');
+
+  /// Get the event ID that this relationship will reference and `null` if there
+  /// is none. This could for example be the thread root, the original event for
+  /// an edit or the event, this is an reaction for. For replies please use
+  /// `Event.inReplyTo` instead!
```
td-famedly: should this point to inReplyToEventId?

td-famedly (commented):
lg for the most part! will take a quick look tmrw moin again :D

lib/src/event.dart
krille-chan: Yes, also mentioned in the commit message. however we already have a lot breaking changes on main anyway

lib/src/event.dart
krille-chan: Good catch. Changed :)

td-famedly (approved):

PR #2180, opened 2025-11-08
feat: New searchEvents method in room

krille-chan:
So this tries a new approach for searching events in encrypted rooms. Searching events in encrypted rooms is hard. We actually need to download the whole timeline. Instead of a stream we now try a simpler API where we just have a Future which returns the found events, plus a next batch and a DateTime of until we have investigated the room timeline. This leads to a user experience like this:

https://github.com/user-attachments/assets/1c378969-06be-4544-b0fb-389012ca1db0

This fixes that:

- [x] Sometimes events are duplicated in the search
- [x] Events remain in memory and after a longer search the app gets laggy
- [x] When we perform a longer search without a result we still get user feedback

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -2728,6 +2728,100 @@ class Room {
     }
   }
 
+  /// Performs a search for `searchTerm` or a more complex `searchFunc()`.
+  /// Be aware that the entire room history (limited by `limit`) will be
+  /// downloaded, decrypted and then checked which needs a lot of time. If
+  /// there are events left in the timeline you get a `nextBatch` back which
+  /// you can use for the next iteration.
+  Future<
+      ({
+        List<Event> events,
+        String? nextBatch,
+        DateTime? searchedUntil,
+      })> searchEvents({
+    String? searchTerm,
+    bool Function(Event)? searchFunc,
+    String? nextBatch,
+
+    /// The amount of events requested from the server in one call. Also used
+    /// to chunk database requests.
+    int limit = 1000,
+
+    /// Other event types are not requested from the server.
+    Set<String> includeEventTypes = const {
+      EventTypes.Message,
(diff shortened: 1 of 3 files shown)


lib/src/room.dart
```diff
@@ -2718,6 +2718,79 @@ class Room {
...
+    assert(searchTerm != null || searchFunc != null);
+
+    searchFunc ??= (event) => event.body.toLowerCase().contains(
+          searchTerm!.toLowerCase(),
+        );
+    final foundEvents = <Event>[];
+
+    // We first check out what we have in database:
+    if (nextBatch == null) {
+      final databaseEvents = await client.database.getEventList(this, start: 0);
```
coder-with-a-bushido: I think it would also be nice to use `limit` here and paginate the DB too so as to save memory?

krille-chan: I've added pagination there as well but in a way that we load everything on the first check. I guess doing this takes less time than a network call

lib/src/room.dart
```diff
@@ -2718,6 +2718,91 @@ class Room {
...
+    if (nextBatch == null) {
+      List<Event> databaseEvents;
+      var start = 0;
+      var timelineComplete = false;
+      do {
+        databaseEvents = await client.database
+            .getEventList(this, start: start, limit: limit);
+        start += limit;
+        foundEvents.addAll(databaseEvents.where(searchFunc));
+        if (databaseEvents.lastOrNull?.type == EventTypes.RoomCreate) {
```
krille-chan, line 2771: I'm going to double check again if the pagination here works!

krille-chan: Okay it works in FluffyChat. I'd now implement it in Famedly in a draft PR and if there I don't run into a blocker, I'd merge this here

PR #2182, opened 2025-11-09
feat: Allow init with access token

krille-chan:
Makes it possible to just init
with an access token and a
homeserver uri and the init()
method would just call /whoami
to fetch the missing data.
Makes it easier to login with
Matrix native OIDC

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1982,9 +1982,9 @@ class Client extends MatrixApi {
   ///
   /// Sends [LoginState.loggedIn] to [onLoginStateChanged].
   ///
-  /// If one of [newToken], [newUserID], [newDeviceID], [newDeviceName] is set then
-  /// all of them must be set! If you don't set them, this method will try to
-  /// get them from the database.
+  /// If one of [newToken] is set, but one of [newUserID], [newDeviceID] is
+  /// null, then this method calls `/whoami` to fetch user ID and device ID
+  /// and rethrows if this request fails.
   ///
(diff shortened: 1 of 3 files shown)


lib/src/client.dart
```diff
@@ -2101,6 +2094,12 @@ class Client extends MatrixApi {
         return;
       }
 
+      if (accessToken != null && (userID == null || deviceID == null)) {
```
coder-with-a-bushido, line 2105: should we throw a `ClientInitPreconditionError` when `accessToken` is null while the `userID` and `deviceID` are null too?

lib/src/client.dart
```diff
@@ -2101,6 +2094,12 @@ class Client extends MatrixApi {
...
+        final userInfo = await getTokenOwner();
+        _userID = userID = userInfo.userId;
+        _deviceID = userInfo.deviceId;
```
coder-with-a-bushido, line 2108: should we also set `deviceID` here like we set the `userID` too in the above line?

lib/src/client.dart
coder-with-a-bushido, line 1978: can we also change the comment to reflect the same?

lib/src/client.dart
krille-chan: That just means that we are not logged in 🤔

lib/src/client.dart
krille-chan: Here `deviceID` is just a getter pointing to `_deviceID` and is also used below to store it in database

lib/src/client.dart
krille-chan: Thanks, good catch :)

PR #2184, opened 2025-11-09
chore: Cache auth metadata response in client

krille-chan:
This is helpful for oidc and
a requirement for implementing
soft logout with oidc.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -132,7 +132,7 @@ class Client extends MatrixApi {
   @override
   set homeserver(Uri? homeserver) {
     if (this.homeserver != null && homeserver?.host != this.homeserver?.host) {
-      _wellKnown = null;
+      _wellKnown = _getAuthMetadataResponseCache = null;
     }
     super.homeserver = homeserver;
   }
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -733,6 +733,12 @@ class Client extends MatrixApi {
     return response;
   }
 
+  GetAuthMetadataResponse? _getAuthMetadataResponseCache;
+
+  @override
+  Future<GetAuthMetadataResponse> getAuthMetadata() async =>
+      _getAuthMetadataResponseCache ?? await super.getAuthMetadata();
```
coder-with-a-bushido: I see this getting reset in the code (homeserver setter) but where exactly is this set in the first place?

lib/src/client.dart
```diff
@@ -132,7 +132,7 @@ class Client extends MatrixApi {
   @override
   set homeserver(Uri? homeserver) {
     if (this.homeserver != null && homeserver?.host != this.homeserver?.host) {
-      _wellKnown = null;
+      _wellKnown = _getAuthMetadataResponseCache = null;
```
coder-with-a-bushido, line 135: What if someone logs out and logs in again with same homeserver but different account? Shouldn't we clear the cache in that case as well?

lib/src/client.dart
coder-with-a-bushido: maybe you wanted to write `??=` instead?

Yash-Garg (approved):

lib/src/client.dart
krille-chan: ah right, thank you so much

lib/src/client.dart
krille-chan: this should be static data at server level so it should always return the same json, similar to the well known

PR #2189, opened 2025-11-17
refactor: CI to integrate job without split of main and app

krille-chan:


scripts/integration-server-synapse.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
-docker run -d --rm --name synapse --tmpfs /data \
-    --volume="$(pwd)/test_driver/synapse/data/homeserver.yaml":/data/homeserver.yaml:rw \
-    --volume="$(pwd)/test_driver/synapse/data/localhost.log.config":/data/localhost.log.config:rw \
+chmod -R 777 test_driver/synapse/data
```
td-famedly: I don't like this, can we turn this to something sensible like 600 or 660?

.github/workflows/app.yml
```diff
@@ -102,17 +125,6 @@ jobs:
           path: merged/
           retention-days: 1
 
-  review_app_coverage:
```
td-famedly, line 105: hm do we no longer need this?

scripts/integration-server-synapse.sh
krille-chan: I can try

.github/workflows/app.yml
krille-chan: We upload coverage to codecov now anyway so I thought we really don't need it do we?

scripts/integration-server-synapse.sh
krille-chan: No, does not work: https://github.com/famedly/matrix-dart-sdk/actions/runs/19461448687/job/55686386853?pr=2189

td-famedly: chown -R user:group would be better then

.github/workflows/app.yml
td-famedly: ah fair enough!

scripts/integration-server-synapse.sh
krille-chan: Okay it works :) had to figure out the userid and used this instead together with sudo in ci but now it works

td-famedly (approved):
lg

PR #2191, opened 2025-11-20
feat: Add room upgrade methods to Room

krille-chan:


td-famedly:
will close until reopened

PR #2193, opened 2025-11-26
feat: add a flag to use MSC3757

yash-garg:


lib/src/voip/voip.dart
```diff
@@ -26,6 +26,10 @@ class VoIP {
   /// everytime this changed would be a pain
   final bool enableSFUE2EEKeyRatcheting;
 
+  /// whether to allow MSC3757 group calls (per-device state keys) even if the
+  /// room version doesn't officially support it.
+  final bool allowRoomMsc3757Calls;
```
td-famedly: should be forceMSC3757

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -488,10 +488,9 @@ class LiveKitBackend extends CallBackend {
       Logs().i(
         '[VOIP E2EE] onCallEncryptionKeyRequest: checkPartcipantStatusAndRequestKey returned false, therefore retrying by getting state from server and rebuilding participant list for sanity',
       );
-      final stateKey =
-          (groupCall.room.roomVersion?.contains('msc3757') ?? false)
-              ? '${userId}_$deviceId'
-              : userId;
+      final roomHasMsc3757 = groupCall.voip.forceMSC3757 ||
```
td-famedly: s/roomHasMsc3757/useMSC3757 probably

lib/src/voip/voip.dart
```diff
@@ -26,6 +26,10 @@ class VoIP {
   /// everytime this changed would be a pain
   final bool enableSFUE2EEKeyRatcheting;
 
+  /// whether to allow MSC3757 group calls (per-device state keys) even if the
+  /// room version doesn't officially support it.
```
td-famedly, line 30: please add a note 

If the room is not actually created for msc3757, state keys that start with your userId but are appended with other strings are not protected and anyone can change them

PR #2196, opened 2025-12-08
feat: MSC4075 - MatrixRTC Notification Event (call ringing)

coder-with-a-bushido:
Needed for https://github.com/famedly/product-management/issues/1005 and https://github.com/famedly/product-management/issues/3270

### DIFF
--- a/lib/msc_extensions/msc_4075_rtc_notification_events/models.dart
+++ b/lib/msc_extensions/msc_4075_rtc_notification_events/models.dart
@@ -1,0 +1,77 @@
+/* MIT License
+* 
+* Copyright (C) 2019, 2020, 2021, 2022 Famedly GmbH
+* 
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+* 
+* The above copyright notice and this permission notice shall be included in all
+* copies or substantial portions of the Software.
+* 
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*/
+
+/// MSC4075: RTC notification events
+/// https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixrtc-call-ringing/proposals/4075-rtc-notification-event.md
+library;
+
+/// Event type constant for RTC notifications
+class RTCNotificationEventTypes {
+  static const String rtcNotification = 'org.matrix.msc4075.rtc.notification';
+}
+
+/// Content for an RTC notification event
+class RTCNotificationContent {
+  /// The unique ID of the call being notified about
+  final String callId;
+
+  /// The room ID where the call is happening
+  final String roomId;
+
+  /// The type of call: 'voice' or 'video'
+  final String callType;
+
+  /// The type of notification: 'ring' (full ringing UI), 'notify' (silent notification)
+  final String notifyType;
+
+  /// When this notification was sent (milliseconds since epoch)
+  final int timestamp;
+
+  RTCNotificationContent({
+    required this.callId,
+    required this.roomId,
+    required this.callType,
+    required this.notifyType,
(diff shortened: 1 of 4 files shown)


lib/msc_extensions/msc_4075_rtc_notification_events/models.dart
```diff
@@ -0,0 +1,77 @@
...
+    required this.timestamp,
+  });
+
+  factory RTCNotificationContent.fromJson(Map<String, dynamic> json) {
+    return RTCNotificationContent(
+      callId: json['call_id'] as String,
+      roomId: json['room_id'] as String,
+      callType: json['call_type'] as String,
+      notifyType: json['notify_type'] as String,
+      timestamp: json['timestamp'] as int,
```
krille-chan, line 64: This does not seem to match with the content description in the MSC. There is no call_id, no room_id etc.
`notify_type` is `notification_type`.  `timestamp` is `sender_ts`. Also `room_id` doesn't make sense here as it is part of the Event object around the content.

```json
{
  "type":"m.rtc.notification", // org.matrix.msc4075.rtc.notification
  "content": {
    "sender_ts": 1752583130365,
    "lifetime": 30000,
    "m.mentions": {"user_ids": [], "room": true | false},
    "m.relates_to": {"rel_type":"m.reference", "event_id":"$rtc_member_event_id"},
    "notification_type": "ring | notification",
  }
}
```

lib/msc_extensions/msc_4075_rtc_notification_events/msc_4075_rtc_notification_events.dart
```diff
@@ -0,0 +1,121 @@
...
+  /// `callId` - The call ID being notified about
+  /// `callType` - The type of call ('voice' or 'video')
+  /// `notifyType` - The notification type ('ring' or 'notify'), defaults to 'ring'
+  Future<String?> sendRTCNotificationEvent({
+    required String roomId,
+    required String callId,
+    required String callType,
+    String notifyType = 'ring',
+  }) async {
+    final room = getRoomById(roomId);
```
krille-chan, line 30: Shouldn't we just make this an extension on `Room` instead of an extension on `Client`? Would be much easier and more intuitive.

lib/msc_extensions/msc_4075_rtc_notification_events/msc_4075_rtc_notification_events.dart
```diff
@@ -0,0 +1,121 @@
...
+library;
+
+import 'package:matrix/matrix.dart';
+import 'package:matrix/msc_extensions/msc_4075_rtc_notification_events/models.dart';
+import 'package:matrix/src/utils/cached_stream_controller.dart';
+
+extension RTCNotificationHandler on Client {
+  /// Stream for incoming RTC notification events
+  static final CachedStreamController<RTCNotificationContent>
+      _onRTCNotification = CachedStreamController();
```
krille-chan, line 12: This has several problems. What if we have multiple Clients like in a multi-Client environment? And also this is a global state which we should definitely avoid.

lib/msc_extensions/msc_4075_rtc_notification_events/msc_4075_rtc_notification_events.dart
```diff
@@ -0,0 +1,121 @@
...
+  /// RTC notifications in your application.
+  ///
+  /// Example:
+  /// ```dart
+  /// client.setupRTCNotificationListener();
+  /// client.onRTCNotification.listen((notification) {
+  ///   print('Incoming call: ${notification.callType} in room ${notification.roomId}');
+  /// });
+  /// ```
+  void setupRTCNotificationListener() {
```
krille-chan, line 64: I think what you actually want to achieve here is: Get a stream of RTC notifications, right? You can just do this like this:

```dart
Stream<RTCNotificationContent> get onRTCNotification => onTimelineEvent.stream.where((event) => event.type == RTCNotificationEventTypes.rtcNotification).map((event) => RTCNotificationContent.fromJson(event.content));
```

And you have what you want without any additional state which you need to "set up".

lib/msc_extensions/msc_4075_rtc_notification_events/msc_4075_rtc_notification_events.dart
```diff
@@ -0,0 +1,121 @@
...
+  /// notification events are received. The rule will make the events trigger
+  /// notifications with a "ring" sound, similar to m.call.invite events.
+  ///
+  /// [ruleId] The ID for this push rule (default: '.m.rule.rtc_notification')
+  ///
+  /// Example:
+  /// ```dart
+  /// await client.enableRTCNotificationPushRule();
+  /// ```
+  Future<void> enableRTCNotificationPushRule({
```
krille-chan, line 89: You can `enable` and `disable` existing push rules also you can `create` and `delete` push rules. This seems to `create` one but the title says `enable`. I guess I know what you want to achieve but I would vote for not creating this method at all. Having it in the scope of the Client API might only produce confusion. What happens if the push rule is already existing?
We have the API already to manage push rules. I would just write in the Readme what the SDK Consumer needs to do.

lib/msc_extensions/msc_4075_rtc_notification_events/msc_4075_rtc_notification_events.dart
```diff
@@ -0,0 +1,121 @@
...
+    );
+    await setPushRuleEnabled(PushRuleKind.underride, ruleId, true);
+  }
+
+  /// Disable push notifications for RTC notification events
+  ///
+  /// Disables the push rule for RTC notification events.
+  ///
+  /// [ruleId] The ID of the push rule to disable (default: '.m.rule.rtc_notification')
+  Future<void> disableRTCNotificationPushRule({
```
krille-chan, line 116: Same here

coder-with-a-bushido:
Closing this in favour of https://github.com/famedly/matrix-dart-sdk/pull/2219

PR #2198, opened 2025-12-09
feat: add primitive models and handler for Simplified Sliding Sync

TheOneWithTheBraid:
**WIP : No need to review so far.**

This PR is part of a larger attempt to get [Simplified Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md) to work with the Matrix Dart SDK.

### DIFF
--- a/lib/msc_extensions/msc_4186_simplified_sliding_sync/models/maybe.dart
+++ b/lib/msc_extensions/msc_4186_simplified_sliding_sync/models/maybe.dart
@@ -1,0 +1,21 @@
+/// Used to depict the difference between null and undefined in JavaScript
+abstract class Maybe<T> {}
+
+class Some<T> extends Maybe<T> {
+  Some(this.data);
+
+  final T data;
+
+  @override
+  int get hashCode => data.hashCode;
+
+  @override
+  String toString() => data.toString();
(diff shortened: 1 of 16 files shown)


lib/msc_extensions/msc_4186_simplified_sliding_sync/models/maybe.dart
```diff
@@ -0,0 +1,21 @@
+/// Used to depict the difference between null and undefined in JavaScript
+abstract class Maybe<T> {}
```
krille-chan, line 2: Interesting concept. What do you think about we make this a `sealed` class? Then we could even use `switch` like:

```dart
extension on Map<String,Object?> {
  Maybe<T> getMaybe<T>(String key) {
    if (!containsKey(key)) return Undefined<T>();
    return Some<T>(this[key] as T?);
  }
}


final maybe = {'value': 5}.getMaybe<int>('value');
switch (maybe) {
  case Some: print('This is something'); break;
  case Undefined: print('This is undefined'); break;
}
```

td-famedly:
will close until reopened

PR #2201, opened 2025-12-15
fix: skip unnecessary 404 errors in profile fetching for member displayname change events in timeline

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/3561

### DIFF
--- a/lib/src/utils/event_localizations.dart
+++ b/lib/src/utils/event_localizations.dart
@@ -158,15 +158,18 @@ abstract class EventLocalizations {
       }
     },
     EventTypes.RoomMember: (event, i18n, body) {
-      final targetName = event.stateKeyUser?.calcDisplayname(i18n: i18n) ?? '';
-      final senderName =
+      // Use lazy functions to avoid triggering profile fetches for cases
+      // like displayname changes where we don't need the sender/target name.
+      String targetName() =>
+          event.stateKeyUser?.calcDisplayname(i18n: i18n) ?? '';
+      String senderName() =>
           event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n);
       final userIsTarget = event.stateKey == event.room.client.userID;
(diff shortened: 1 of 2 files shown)


lib/src/utils/event_localizations.dart
```diff
@@ -158,6 +158,16 @@ abstract class EventLocalizations {
       }
     },
     EventTypes.RoomMember: (event, i18n, body) {
+      // Handle displayname changes early to avoid triggering profile fetches for
+      // historical events where the user may no longer exist (causing 404 errors).
+      if (event.roomMemberChangeType == RoomMemberChangeType.displayname) {
```
krille-chan: Looks a little bit hacky. Doesn't it make more sense to just move the senderName creation more below to where it is needed?

krille-chan: Or you change `final senderName` to not be a String but a function which returns a String and then only call the function where you need it like `i18n.unbannedUser(senderName(), targetName);`

test/event_test.dart
```diff
@@ -1221,6 +1221,79 @@ void main() async {
       await matrix.dispose(closeDatabase: true);
     });
 
+    test('calcLocalizedBody displayname change should not trigger user lookup',
```
krille-chan: I don't think this test is necessary. We should not test that something does not exist. Making sure that something does not have a side effect would more look like that something is a black box and could do anything. Should more be handled by a elegant code design instead of a test. Otherwise this would impliciet that we need every single method in the SDK to test that it does not 10 different possible side effects

coder-with-a-bushido: I wrote a test to check it instead of manually and then I thought it'll make codecov happy somehow, so I just included it haha

coder-with-a-bushido: Removed it now after testing again!

PR #2203, opened 2025-12-16
refactor: (BREAKING) Actually wipe ssss and cross signing when calling bootstrap wipe methods

krille-chan:
Closes https://github.com/famedly/product-management/issues/3521

Before we have just done nothing
when calling wipeCrossSigning() or
wipeSsss() in bootstrap and just
skipped to setup new. While
wiping usually means we want to re-init something, this
could lead to an undesired
state. Maybe the
consumer wants to just wipe
everything but start from scratch
later. I also think this could
lead to the case that some
bootstrap setups are
corrupted as we leave the
old recovery keys in
account data forever and
never wipe them.

The new behavior sends an
empty json object instead which
is the same behavior, Element
seem to do.

### DIFF
--- a/lib/encryption/cross_signing.dart
+++ b/lib/encryption/cross_signing.dart
@@ -123,6 +123,19 @@ class CrossSigning {
                 key.identifier != client.deviceID,
       );
 
+  Future<void> wipe() async {
+    await client.accountDataLoading;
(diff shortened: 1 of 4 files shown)


krille-chan:
Instead I'd like to completely refactor the Bootstrap class

PR #2207, opened 2025-12-18
feat: implement retry encryption key request mechanism for livekit calls

coder-with-a-bushido:
Closes https://github.com/famedly/product-management/issues/2097
Closes https://github.com/famedly/product-management/issues/2078

### DIFF
--- a/lib/src/voip/backend/livekit_backend.dart
+++ b/lib/src/voip/backend/livekit_backend.dart
@@ -24,6 +24,12 @@ class LiveKitBackend extends CallBackend {
   /// participant:keyIndex:keyBin
   final Map<CallParticipant, Map<int, Uint8List>> _encryptionKeysMap = {};
 
+  /// Tracks pending encryption key request retries per participant
+  final Map<CallParticipant, Timer> _requestEncryptionKeyPending = {};
+
+  /// Retry interval for key requests
+  static const Duration _keyRequestRetryInterval = Duration(seconds: 2);
+
   final List<Future> _setNewKeyTimeouts = [];
 
   int _indexCounter = 0;
@@ -79,10 +85,7 @@ class LiveKitBackend extends CallBackend {
         '_makeNewSenderKey using previous key because last created at ${_lastNewKeyTime.toString()}',
       );
       // still a fairly new key, just send that
-      await _sendEncryptionKeysEvent(
-        groupCall,
-        _latestLocalKeyIndex,
-      );
+      await _sendEncryptionKeysEvent(groupCall, _latestLocalKeyIndex);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/models/retry_event_model.dart
```diff
@@ -0,0 +1,14 @@
+import 'dart:async';
+
+/// Retries the `retryFunction` after a set `timeInterval` until `dispose` is called
+class RetryEventModel {
```
krille-chan: I don't really get why we need this. Can't we just use `Timer` directly? I guess you want the following:

1. Maintain a list of delayed retries
2. Check if a retry is already pending
3. Cancel a retry

You can all of this do with a Timer:

```dart
_requestEncryptionKeyPending[callParticipant] = Timer(const Duration(...), retryFunction);

_requestEncryptionKeyPending[callParticipant]?.cancel();
_requestEncryptionKeyPending.remove(callParticipant);
```

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -24,6 +25,18 @@ class LiveKitBackend extends CallBackend {
   /// participant:keyIndex:keyBin
   final Map<CallParticipant, Map<int, Uint8List>> _encryptionKeysMap = {};
 
+  /// Tracks pending encryption key request retries per participant
+  final Map<CallParticipant, RetryEventModel> _requestEncryptionKeyPending = {};
+
+  /// Tracks retry count per participant to prevent infinite retries
+  final Map<CallParticipant, int> _keyRequestRetryCount = {};
```
krille-chan: Do we really need to store it as a state? I'd like to avoid any kind of state and just make it part of the retry-method which just stops after 5 attempts?

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -403,8 +429,14 @@ class LiveKitBackend extends CallBackend {
...
     final callId = keyContent.callId;
-    final p =
-        CallParticipant(groupCall.voip, userId: userId, deviceId: deviceId);
+    final p = CallParticipant(
+      groupCall.voip,
+      userId: userId,
+      deviceId: deviceId,
+    );
+
+    // Cancel any pending retry for this participant since we received keys
```
td-famedly, line 438: How does this fix the issue mentioned in https://github.com/famedly/product-management/issues/2078? They might send us keys for index N+1 but never start using it, we should only cancel requests when we get a answer for the requested index

lib/src/voip/backend/livekit_backend.dart
```diff
@@ -387,6 +390,29 @@ class LiveKitBackend extends CallBackend {
...
+          retryCount++;
+          if (retryCount >= 5) {
+            Logs().w(
+              '[VOIP E2EE] Max retries (5) reached for ${rp.id}, giving up key request',
+            );
+            timer.cancel();
+            _requestEncryptionKeyPending.remove(rp);
+            return;
+          }
+          unawaited(requestEncrytionKey(groupCall, [rp]));
```
td-famedly, line 412: any reason this is unawaited?

td-famedly (changes requested):

lib/src/voip/backend/livekit_backend.dart
coder-with-a-bushido: My bad, I misread the issue! thank you

coder-with-a-bushido: Will work on it now

PR #2212, opened 2025-12-29
refactor: Store read receipts in own box instead of room account data

krille-chan:
Closes https://github.com/famedly/product-management/issues/423


coder-with-a-bushido (approved):
LGTM :)

PR #2214, opened 2025-12-31
feat: Simplified bootstrap with crypto identity extension

krille-chan:
closes https://github.com/famedly/product-management/issues/3493

I think it makes sense to still keep the Bootstrap class but offer a simplified API on top of it. The Bootstrap object with it's state makes sure that we do never run into a broken state (in theory) and some consumers might still want to use it.

### DIFF
--- a/doc/end-to-end-encryption.md
+++ b/doc/end-to-end-encryption.md
@@ -38,3 +38,48 @@ final client = Client('Matrix Client',
     // ...
 );
 ```
+
+### Setup your crypto identity
+
+To use **Secure Storage and Sharing**, **Cross Signing** and the **Online Key Backup**,
+you should set up your crypto identity. The crypto identity is defined as the
+combined feature of those three features. First you should check if it is already
+set up for this account:
+
+```dart
+final state = await client.getCryptoIdentityState();
+if (state.initialized) {
+    print('Your crypto identity is initialized. You can either restore or wipe it.');
+}
+if (state.connected) {
+    print('Your crypto identity is initialized and you are connected. You can now only wipe it to reset your passphrase or recovery key!');
+}
+```
+
+If `initialized` is `false` you need to initialize your crypto identity first:
+
+```dart
+final recoveryKey = await client.initCryptoIdentity();
+```
+
+You can also set a custom passphrase:
+
+```dart
+final passphrase = await client.initCryptoIdentity('SuperSecurePassphrase154%');
+```
+
+To then reconnect on a new device you can restore your crypto identity:
+
+```dart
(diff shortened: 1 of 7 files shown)


lib/encryption/utils/crypto_setup_extension.dart
```diff
@@ -0,0 +1,194 @@
+import 'dart:async';
+
+import 'package:matrix/encryption/utils/bootstrap.dart';
+import 'package:matrix/matrix.dart';
+
+extension CryptoSetupExtension on Client {
+  /// Returns the current state of the crypto identity. If initialized is
+  /// true, then `Client.restoreCryptoIdentity()` can be used to connect. When
+  /// initialized is false, then `Client.initCryptoIdentity()` must
+  /// be executed first.
```
td-famedly: Can you add a comment which says what initiated and connected are?

test_driver/dendrite/data/dendrite.yaml
```diff
@@ -185,7 +185,7 @@ client_api:
   # will be released after the cooloff time in milliseconds. Server administrators
   # and appservice users are exempt from rate limiting by default.
   rate_limiting:
-    enabled: true
+    enabled: false
```
yash-garg, line 188: is this expected?

doc/end-to-end-encryption.md
```diff
@@ -37,4 +37,47 @@ final client = Client('Matrix Client',
...
+await client.restoreCryptoIdentity(passphraseOrRecoveryKey);
+```
+
+If you have lost your passphrase or recovery key, you can wipe your crypto
+identity and get a new key with `client.initCryptoIdentity()` at any time.
+
+> hint: An alternative to `client.restoreCryptoIdentity()` can be that you use
+> **key verification** to connect with another session which is already connected.
+> The Client would then request all necessary secrets of your crypto identity
+> automatically via **to-device-messaging**.
```
yash-garg:
```suggestion
> [!TIP]
> An alternative to `client.restoreCryptoIdentity()` can be that you use
> **key verification** to connect with another session which is already connected.
>
> The Client would then request all necessary secrets of your crypto identity
> automatically via **to-device-messaging**.
```

lib/encryption/utils/crypto_setup_extension.dart
krille-chan: Done :)

test_driver/dendrite/data/dendrite.yaml
krille-chan: Yes :) It is because we logout and login once more in the e2ee tests and this triggers rate limit in dendrite otherwise

PR #2215, opened 2026-01-02
refactor: (BREAKING) Event info map

krille-chan:
closes https://github.com/famedly/product-management/issues/2210

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -519,13 +519,13 @@ class Event extends MatrixEvent {
   }
 
   /// Gets the info map of file events, or a blank map if none present
-  Map get infoMap =>
+  Map<String, Object?> get infoMap =>
       content.tryGetMap<String, Object?>('info') ?? <String, Object?>{};
 
   /// Gets the thumbnail info map of file events, or a blank map if nonepresent
-  Map get thumbnailInfoMap => infoMap['thumbnail_info'] is Map
-      ? infoMap['thumbnail_info']
-      : <String, dynamic>{};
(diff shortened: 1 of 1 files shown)


lib/src/event.dart
```diff
@@ -794,16 +809,24 @@ class Event extends MatrixEvent {
 
     // Decrypt the file
     if (isEncrypted) {
-      final fileMap =
-          getThumbnail ? infoMap['thumbnail_file'] : content['file'];
-      if (!fileMap['key']['key_ops'].contains('decrypt')) {
+      final fileMap = getThumbnail
+          ? infoMap.tryGetMap<String, Object?>('thumbnail_file')
```
coder-with-a-bushido, line 813: we can use `thumbnailInfoMap` here?

lib/src/event.dart
```diff
@@ -794,16 +809,24 @@ class Event extends MatrixEvent {
...
+          : content.tryGetMap<String, Object?>('file');
+      if (fileMap == null) throw ('No encrypted file info');
```
coder-with-a-bushido: throw ('No encrypted file info found')

krille-chan: Fixed :)

lib/src/event.dart
krille-chan: No, because thumbnailInfoMap gives us 'thumbnail_info' and not ' thumbnail_file' which is a different json object for encryption purposes

coder-with-a-bushido: sorry, my bad!

PR #2216, opened 2026-01-02
refactor: (BREAKING) Use unified requestAndCache method

krille-chan:
This caches multiple static
content requests in the database
with a unified and simple way
which reduces in memory
state.

closes https://github.com/famedly/product-management/issues/2358

This simplifies some things a lot but also introduces an inconveniency: You can no longer access the wellKnown cache in and synchronious way by `client.wellKnown` but have to always await client.getWellKnown(). IMO this is definitely worth it to get rid of the state and win simplicity and also have these static content stored in the exactly same manner:

- Well-Known
- Auth Metadata
- Versions
- Client Config


lib/src/utils/request_and_cache.dart
```diff
@@ -0,0 +1,36 @@
...
+      await database.cacheCustomObject(cacheKey, toJson(content));
+      return content;
+    } catch (error, stackTrace) {
+      Logs().w(
+        'Unable to update cache for $cacheKey',
+        error,
+        stackTrace,
+      );
+
+      if (cachedResponse != null) {
```
coder-with-a-bushido: is it really okay to return the user an expired cached data when a new request fails? I personally think we should just rethrow directly

lib/src/client.dart
```diff
@@ -625,7 +628,7 @@ class Client extends MatrixApi {
     _wellKnown = wellKnown;
     await database.storeWellKnown(wellKnown);
     return wellKnown;
-  }
+  }*/
```
coder-with-a-bushido: can we remove this commented section?

lib/src/client.dart
```diff
@@ -1304,23 +1313,26 @@ class Client extends MatrixApi {
...
+        super.getVersions,
+        fromJson: GetVersionsResponse.fromJson,
+        toJson: (response) => response.toJson(),
+        cacheKey: 'get_versions',
+        cacheLifetime: cacheLifetime,
+      );
 
   Future<bool> authenticatedMediaSupported() async {
-    return (await versionsResponse).versions.any(
+    return (await getVersions()).versions.any(
```
coder-with-a-bushido, line 1317: we could have a variable for `await getVersions()` so that it's not called twice

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1773,22 +1773,26 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   }
 
   @override
-  Future<void> storeWellKnown(DiscoveryInformation? discoveryInformation) {
-    if (discoveryInformation == null) {
-      return _clientBox.delete('discovery_information');
-    }
+  Future<void> cacheCustomObject(String cacheKey, Map<String, Object?> object) {
```
coder-with-a-bushido, line 1798: we need to make sure that all these values are also cleared on `clearCache()` method in the same class. Maybe add some tests too?

lib/src/utils/request_and_cache.dart
krille-chan: I would say it is better for the UX. We are doing this in a lot of other places already. And without WellKnown information for example most features of our app are not functional. And as these things very rarely change at all (maybe once a year or so or maybe even never?) having an "expired" result shouldn't be that bad.

lib/src/client.dart
krille-chan: ah good catch!

lib/src/utils/request_and_cache.dart
coder-with-a-bushido: What if there's a config for blocking calls in well known and its changed on server side. Let's say the app fails to fetch this due to some internet issue and gives the old cached data. Later let's say the internet recovers but the app gets well known only on app startup so it has old data. This will make the user able to make calls even if it's changed on server side.

krille-chan: Yes but that is impossible to solve anyway, right? If the well known changes and the app can not download it, then the app should not be unusable. Changing something on server side should always make the customer expect that this can take time until it affects everyone

coder-with-a-bushido: Hmm, but I still think if we throw the error at least we can expect it to be handled in the app with whatever logic they want it to be. Maybe they can show a "no internet" sign and then retry instead of using an invalidated well-known without even knowing there was an error fetching it?

What if the endpoint stopped working 6 months back and no client complained since everything was working fine? I think this could only do bad things here personally.

coder-with-a-bushido: Maybe we could just also have another way to get the invalidated cache, which can be used when this method fails (if the implementation wants it). We could also write it into the dartdoc comment for the method

krille-chan: > What if the endpoint stopped working 6 months back and no client complained since everything was working fine? I think this could only do bad things here personally.

Before we also never showed an error and I really don't think this would be any helpful. We **need** any well known for the app to be functional. If well known stops working for 6 months then makign the whole server unusable is not a good idea IMO.

However making this configurable sounds like a good idea. I'll add a `throwOnUpdateFailure` flag

lib/src/client.dart
krille-chan: If it is called once then it will fetch the next from the cache which is already in memory of the database cache. I'd prefer to avoid an additional state here and make the sdk in the long term more database focused

lib/src/database/matrix_sdk_database.dart
krille-chan: I added a new method `clearCustomCacheObjects()` which is called in `clearCache()` and I've added a test for it :)

lib/src/utils/request_and_cache.dart
coder-with-a-bushido: Sounds good to me :D

PR #2223, opened 2026-01-07
refactor: Add name property to ssss key and pass through keyId for opening

krille-chan:
Makes it possible to set a `name` for
the SSSS key which is specified in the
spec but has never been implemented
here.

Also makes it possible to
set the key ID when using an existing
SSSS in bootstrap. by default it still
uses the default key ID so this is
not a breaking change.

### DIFF
--- a/lib/encryption/utils/bootstrap.dart
+++ b/lib/encryption/utils/bootstrap.dart
@@ -204,13 +204,13 @@ class Bootstrap {
     }
   }
 
-  void useExistingSsss(bool use) {
+  void useExistingSsss(bool use, {String? keyIdentifier}) {
     if (state != BootstrapState.askUseExistingSsss) {
       throw BootstrapBadStateException('Wrong State');
     }
     if (use) {
       try {
-        newSsssKey = encryption.ssss.open(encryption.ssss.defaultKeyId);
+        newSsssKey = encryption.ssss.open(keyIdentifier);
         state = BootstrapState.openExistingSsss;
       } catch (e, s) {
         Logs().e('[Bootstrapping] Error open SSSS', e, s);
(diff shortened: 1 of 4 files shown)


lib/encryption/utils/bootstrap.dart
```diff
@@ -204,13 +204,13 @@ class Bootstrap {
     }
   }
 
-  void useExistingSsss(bool use) {
+  void useExistingSsss(bool use, {String? keyId}) {
```
td-famedly: This is called a `identifier` in other places, maybe we want to stay consistent?

lib/encryption/utils/bootstrap.dart
```diff
@@ -258,14 +258,14 @@ class Bootstrap {
     state = BootstrapState.askNewSsss;
   }
 
-  Future<void> newSsss([String? passphrase]) async {
+  Future<void> newSsss([String? passphrase, String? name]) async {
```
td-famedly, line 261: And then this name is set as the defaultKeyId and we don't have to worry about other places like wipessss correct?

lib/encryption/utils/bootstrap.dart
krille-chan: Let's say `keyIdentifier` as a compromise because that what it's it, ok?

lib/encryption/utils/bootstrap.dart
krille-chan: it is just a String added to the content of the account data type. It does not change any other behavior, just for future purpuses to identify the right key

td-famedly: yep but wiping for example still uses the default key id, if that's supposed to be handled in a different PR feel free to resolve and merge

PR #2243, opened 2026-02-04
refactor: Always try deserialize error as matrixexception

krille-chan:
Actually only error codes below 500 should be Matrix Exception but in reality Synapse sends them formatted as MatrixException sometimes also with 500 or 502 error code.

closes https://github.com/famedly/product-management/issues/3712


td-famedly (approved):
lg

PR #2259, opened 2026-02-20
build: Add dependabot config

krille-chan:


.github/dependabot.yml
```diff
@@ -1,11 +1,24 @@
...
+    open-pull-requests-limit: 5
+    versioning-strategy: increase-if-necessary
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "*"
+    commit-message:
+      prefix: "build: "
```
td-famedly, line 23: not dependabot? not sure if build makes the most sense here

krille-chan: To use conventional commits

td-famedly: then maybe ci makes more sense? because it was created by the ci? but idm really

krille-chan: build comes from the conventional commits extension from angular and is documented here:
https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

> build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)

td-famedly: sure lg!

PR #2272, opened 2026-02-25
fix: expiresIn from OIDC is in seconds not milliseconds

krille-chan:
This fixes that the SDK assumed that expiresIn is in milliseconds like with the legacy refresh token endpoint from matrix. In fact it makes it more robust by making the expiresIn field in the response model a Duration.


rmoradi88 (approved):
LGTM

PR #2273, opened 2026-02-25
fix: Retry refresh token on io exception

krille-chan:
This also uses a default for soft logout if we get soft logged out but the SDK consumer hasn't set the soft logout callback. In case of Matrix native OIDC this makes sense now as with oidc you can not set anymore if you want to use refresh tokens or not, you just get forced to use it and the SDK now instantly logs you out if you don't set it.

Tested it on FluffyChat and works fine
there.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -2372,11 +2372,7 @@ class Client extends MatrixApi {
   Future<void>? _handleSoftLogoutFuture;
 
   Future<void> _handleSoftLogout() async {
-    final onSoftLogout = this.onSoftLogout;
-    if (onSoftLogout == null) {
-      await logout();
-      return;
-    }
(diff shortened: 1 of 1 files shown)


lib/src/client.dart
```diff
@@ -287,24 +288,36 @@ class Client extends MatrixApi {
...
+        // We are using Matrix Native OIDC so we fetch the refresh endpoint first:
+        final String oidcClientId =>
+          await oidcRefresh(oidcClientId, refreshToken),
+      };
+    } on http.ClientException catch (_) {
+      Logs().w(
+        'Connection problem while trying to refresh access token. Try again in ${retryIn.inSeconds} seconds...',
+      );
+      await Future.delayed(retryIn);
+      return refreshAccessToken(
```
td-famedly: we shouldn't have loops, maybe terminate it after a while?

PR #2277, opened 2026-02-27
refactor: (BREAKING) add type for power level as extension type over int

krille-chan:


lib/src/room.dart
```diff
@@ -2160,8 +2159,8 @@ class Room {
   bool get canBan {
     if (membership != Membership.join) return false;
     return (getState(EventTypes.RoomPowerLevels)?.content.tryGet<int>('ban') ??
-            50) <=
-        ownPowerLevel;
+            PowerLevel.defaultModeratorLevel) <=
```
td-famedly, line 2185: hm that looks incorrect actually, shouldn't this be state default first? then fallback to 50?

lib/src/room.dart
```diff
@@ -2242,39 +2241,43 @@ class Room {
...
   }
 
   /// The level required to kick a user.
   bool get canKick {
     if (membership != Membership.join) return false;
-    return (getState(EventTypes.RoomPowerLevels)?.content.tryGet<int>('kick') ??
-            50) <=
+    return PowerLevel(
+          getState(EventTypes.RoomPowerLevels)?.content.tryGet<int>('kick') ??
+              PowerLevel.defaultModeratorLevel,
```
td-famedly, line 2279: same here?

td-famedly (commented):
not exactly related to the PR but maybe a diff bug. Also not completely sure about having a type for this. Is there a particular reason we are going for types over magic numbers? (just trying to avoid the breaking change)

lib/src/room.dart
krille-chan: You can read it here: https://spec.matrix.org/v1.17/client-server-api/#mroompower_levels

> ban 	integer 	The level required to ban a user. Defaults to 50 if unspecified.

So this is not a state event you send (while it manipulates a state event later) but banning is an action on it's onw. It's either what is in the power level state or 50 and does **not** fallback to state default as far as I understood the spec

lib/src/room.dart
krille-chan: Same here

lib/src/room.dart
td-famedly: ah good catch yes

td-famedly (approved):
lg

PR #2280, opened 2026-02-28
refactor: Use RoomAlias type for aliases

krille-chan:


td-famedly:
will close until reopened

PR #2294, opened 2026-03-09
feat: Store sending files in file cache for 30 days

krille-chan:
This makes it possible to retry sending a file event even after the app got restarted. It replaces the current in-memory-cache for sending files with using the existing store file methods in the database.

I've tested it with FluffyChat where it works like a charm :)

Tested on:

- [x] FluffyChat Android
- [x] FluffyChat macOS
- [x] FluffyChat Web (still only stores in memory but now with better error message)

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -438,7 +438,76 @@ class Event extends MatrixEvent {
     room.client.onCancelSendEvent.add(eventId);
   }
 
+  Future<MatrixFile?> _getCachedFile({bool getThumbnail = false}) async {
+    if (transactionId == null) return null;
+
+    final filename = content.tryGet<String>('filename')!;
+
+    if (getThumbnail) {
+      final thumbnailBytes = await room.client.database.getFile(
+        Uri(
+          scheme: 'cache',
+          host: 'thumbnail',
+          path: transactionId,
+        ),
+      );
+      if (thumbnailBytes != null) {
+        return MatrixImageFile(
+          bytes: thumbnailBytes,
+          name: filename,
+          mimeType: thumbnailMimetype,
+          width: thumbnailInfoMap.tryGet<int>('w'),
+          height: thumbnailInfoMap.tryGet<int>('h'),
+        );
+      }
+    }
+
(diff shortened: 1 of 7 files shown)


lib/src/event.dart
```diff
@@ -438,7 +438,76 @@ class Event extends MatrixEvent {
...
+          name: filename,
+          mimeType: thumbnailMimetype,
+          width: thumbnailInfoMap.tryGet<int>('w'),
+          height: thumbnailInfoMap.tryGet<int>('h'),
+        );
+      }
+    }
+
+    final fileBytes = await room.client.database.getFile(
+      Uri(
```
td-famedly, line 466: can you add docs for the getFile and storeFile method because these are public and people might want to figure out the scheme without looking into the code

lib/src/room.dart
```diff
@@ -1063,8 +1068,14 @@ class Room {
       threadLastEventId: threadLastEventId,
       displayPendingEvent: displayPendingEvent,
     );
-    sendingFilePlaceholders.remove(txid);
-    sendingFileThumbnails.remove(txid);
+    await client.database.deleteFile(
```
td-famedly, line 1071: do we not want to delete the cached file after sending successfully?

lib/src/event.dart
krille-chan: I've added dart docs which explain the Uri format :)

lib/src/room.dart
krille-chan: This does delete it

td-famedly: right, no idea what this comment was then

test/room_test.dart
```diff
@@ -1383,18 +1385,31 @@ void main() {
...
+      } catch (_) {}
+
+      final storedFile = await room.client.database.getFile(
+        Uri(
+          scheme: 'cache',
+          host: 'file',
+          path: txnid,
+        ),
+      );
+      expect(storedFile, testFile.bytes);
```
td-famedly: actually can you also add a successful file send and then make sure the cache is also deleted?

PR #2296, opened 2026-03-10
feat: add fetchPollResponses method for handling fragmented timelines

kamor-famedly:
Add fetchPollResponses method to handle polls in fragmented timelines.

### DIFF
--- a/lib/msc_extensions/msc_3381_polls/poll_event_extension.dart
+++ b/lib/msc_extensions/msc_3381_polls/poll_event_extension.dart
@@ -48,6 +48,18 @@ extension PollEventExtension on Event {
     );
   }
 
+  /// Fetches poll response events from the server for fragmented timelines
+  /// where responses may not be in the current timeline chunk.
+  /// After fetching, the existing sync [getPollResponses] method will
+  /// return the correct data from [timeline.aggregatedEvents].
+  Future<void> fetchPollResponses(Timeline timeline) async {
+    assert(type == PollEventContent.startType);
+    await timeline.fetchAggregatedEvents(
+      eventId,
+      RelationshipTypes.reference,
+    );
+  }
+
   Event? _getEndPollEvent(Timeline timeline) {
     assert(type == PollEventContent.startType);
     final aggregatedEvents =
--- a/lib/src/timeline.dart
+++ b/lib/src/timeline.dart
@@ -56,6 +56,13 @@ class Timeline {
 
   final Map<String, Event> _eventCache = {};
 
+  /// Tracks in-flight and completed relation fetches. A present key means
+  /// the fetch is either in progress or already done — concurrent callers
+  /// await the same future, and later callers get an already-completed
+  /// future (no-op). The key is removed on failure so the fetch can be
+  /// retried.
+  final Map<String, Future<void>> _relationRequests = {};
+
   TimelineChunk chunk;
 
   /// Searches for the event in this timeline. If not
@@ -315,6 +322,15 @@ class Timeline {
       }
     }
 
+    await Future.wait(
+      newEvents.where((event) => event.type == PollEventContent.startType).map(
+            (event) => fetchAggregatedEvents(
+              event.eventId,
+              RelationshipTypes.reference,
+            ),
+          ),
+    );
+
     if (onUpdate != null) {
       onUpdate!();
     }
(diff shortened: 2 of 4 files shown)


lib/msc_extensions/msc_3381_polls/poll_event_extension.dart
```diff
@@ -48,6 +48,18 @@ extension PollEventExtension on Event {
     );
   }
 
+  /// Fetches poll response events from the server for fragmented timelines
+  /// where responses may not be in the current timeline chunk.
+  /// After fetching, the existing sync [getPollResponses] method will
+  /// return the correct data from [timeline.aggregatedEvents].
+  Future<void> fetchPollResponses(Timeline timeline) async {
```
krille-chan, line 55: Do we maybe want to call this automatically inside of the timeline object instead? As part of the methods:

`Timeline.requestFuture()` and `Room.getTimeline()` if `eventContext != null`

lib/src/timeline.dart
```diff
@@ -525,6 +532,107 @@ class Timeline {
...
+    String? eventType,
+  }) async {
+    final key = eventType != null
+        ? '$eventId:$relType:$eventType'
+        : '$eventId:$relType';
+
+    if (_fetchedRelations.contains(key)) return;
+
+    // Coalesce concurrent requests for the same key
+    if (_inFlightRelationFetches.containsKey(key)) {
```
td-famedly: maybe instead of your own deduplication you can use https://pub.dev/documentation/async/latest/async/AsyncCache-class.html? the duration can be infinite ofc. You can ever store them in the db (not along the timeline ofc, so that it's faster the next time you go to a pinned msg and you can support offline mode)

td-famedly: maybe the offline mode is out of scope for this bug though. feel free to move it to a new issue

krille-chan:
> @td-famedly & @krille-chan
> 
> I’ve made the proper changes based on your reviews. Let me know if you have any further notes.

Import sorter is complaining 🤔

lib/src/timeline.dart
```diff
@@ -54,6 +55,14 @@ class Timeline {
 
   final Map<String, Event> _eventCache = {};
 
+  /// Tracks which relation fetches have already been completed to avoid
+  /// redundant API calls. Key format: "$eventId:$relType" or
+  /// "$eventId:$relType:$eventType".
+  final Set<String> _fetchedRelations = {};
+
+  /// Deduplicates concurrent in-flight relation fetches.
+  final Map<String, AsyncCache<void>> _inflightRelationRequests = {};
```
krille-chan: If we just do not remove the key here, can't this also cover the use case of `_fetchedRelations` so that we can get rid of `_fetchedRelations`?

lib/src/timeline.dart
```diff
@@ -313,6 +320,12 @@ class Timeline {
       }
     }
 
+    for (final event in newEvents) {
+      if (event.type == PollEventContent.startType) {
+        await fetchAggregatedEvents(event.eventId, RelationshipTypes.reference);
```
td-famedly: hm this might break offline support now because we do a network request here. Can you add a try catch please?

kamor-famedly: fetchAggregatedEvents already catches and logs the failure, without throwing.
We should simply unawait here so we don't block timeline loading.

lib/src/timeline.dart
```diff
@@ -313,6 +320,14 @@ class Timeline {
       }
     }
 
+    for (final event in newEvents) {
+      if (event.type == PollEventContent.startType) {
+        unawaited(
```
td-famedly: actually getRoomEvents already does a network request so it's fine to await and block it here

td-famedly: If we do not block it, the network request will finish at a future point but then onUpdate of the timeline will not be called

kamor-famedly: @td-famedly 
I realized my initial loop was sequential...

Is this the intended concurrent approach?


```dart
    await Future.wait(
      newEvents.where((event) => event.type == PollEventContent.startType).map(
            (event) => fetchAggregatedEvents(
              event.eventId,
              RelationshipTypes.reference,
            ),
          ),
    );

    if (onUpdate != null) {
      onUpdate!();
    }
```

td-famedly (approved):
lg

PR #2301, opened 2026-03-17
fix: shrinkImage fail blocks sendFileEvent without ux feedback

### LINKED ISSUE #2724: Can’t send photos from the default camera app
### Bug Description

I just tried to send a photo for the first time to show something to an contact but for some reason I could not send it, in fact, no matter which room on which homeserver I’m in it all give me the same error over and over again, it seems that it’s only happens from the photos taken from the camera app, normal images doesn’t seem to happen.

### Steps to Reproduce

Take any picture from the default camera app
Go to any homeserver/room and try to send the photo

### Expected Behavior

It’s supposed to send the photo like any image 

### App Version

2.4.0

### Platform

iOS (iPhone)

### Additional Platform Information

iPhone 11, iOS 16.6.1

### Additional Context

https://files.catbox.moe/d6ncsr.mp4
I had to upload the video in another platform because for some reason GitHub wasn’t uploading the video file. This doesn’t mean there’s something wrong in the iOS system here because I can send everything in others chat apps, FluffyChat seems to be having this issue.

But as you can see this is what happening to me.

krille-chan:
We could either say that in this edge case we block the sending of the file event or we continue with the original file. I'd vote for try to continue with the original file because otherwise when using for example an unsupported image format it gets impossible to send this image at all.

Fixes https://github.com/krille-chan/fluffychat/issues/2724

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -922,22 +922,26 @@ class Room {
       syncUpdate.rooms!.join!.values.first.timeline!.events!.first
               .unsigned![fileSendingStatusKey] =
           FileSendingStatus.generatingThumbnail.name;
-      thumbnail ??= await file.generateThumbnail(
-        nativeImplementations: client.nativeImplementations,
-        customImageResizer: client.customImageResizer,
-      );
-      if (shrinkImageMaxDimension != null) {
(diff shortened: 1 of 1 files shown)


td-famedly (approved):
lg

PR #2303, opened 2026-03-23
fix: null check operator in SSSS accountData parsing

coder-with-a-bushido:
Required for https://github.com/famedly/product-management/issues/3718

### DIFF
--- a/lib/encryption/ssss.dart
+++ b/lib/encryption/ssss.dart
@@ -789,9 +789,10 @@ class OpenSSSS {
     }
     await ssss.store(type, secret, keyId, privateKey, add: add);
     while (!ssss.client.accountData.containsKey(type) ||
-        !(ssss.client.accountData[type]!.content
-            .tryGetMap<String, Object?>('encrypted')!
-            .containsKey(keyId)) ||
+        (ssss.client.accountData[type]!.content
+                .tryGetMap<String, Object?>('encrypted')
+                ?.containsKey(keyId) !=
+            true) ||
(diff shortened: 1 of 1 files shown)


lib/encryption/ssss.dart
```diff
@@ -64,14 +64,39 @@ class SSSS {
   final CachedStreamController<String> onSecretStored =
       CachedStreamController();
 
-  SSSS(this.encryption);
+  String? _latestLocallyCreatedKeyId;
+  String? _previousDefaultKeyId;
+
+  /// Callback for when the default key has been changed.
+  void Function()? onDefaultKeyChanged;
```
krille-chan: Do we really need this?

lib/encryption/ssss.dart
```diff
@@ -64,14 +64,39 @@ class SSSS {
...
     _cache.clear();
   }
 
+  Future<void> handleAccountDataUpdate(String type) async {
+    if (type == EventTypes.SecretStorageDefaultKey) {
+      final currentDefault = defaultKeyId;
+      if (currentDefault != _previousDefaultKeyId) {
+        _previousDefaultKeyId = currentDefault;
+        if (currentDefault != _latestLocallyCreatedKeyId) {
+          await clearCache();
```
krille-chan: We should not clear the cache when the default key changes. It does not always mean that the Secrets also got wiped. Maybe they just got re encrypted but are still the same.

mErrenst (approved):

PR #2305, opened 2026-03-24
chore: add engineering standards (dependabot)

nikzen:
Automated rollout from [engineering-standards](https://github.com/famedly/engineering-standards).

**Base (always included):**
- `.github/dependabot.yml` – automated dependency updates (Dependabot)
- Ecosystems: github-actions, dart
- Grouped PRs: major updates separate from minor+patch
- Cooldown: 14 days

Detected scope: **dart**


.github/dependabot.yml
```diff
@@ -1,24 +1,41 @@
...
+    groups:
+      actions:
+        patterns: ["*"]
     commit-message:
-      prefix: "build: "
-      include: "scope"
-    assignees:
-      - "famedly/frontend"
-    open-pull-requests-limit: 5
-    versioning-strategy: increase-if-necessary
```
krille-chan, line 15: We must not remove the versioning-strategy for dart packages as we want to update the range for major versions only if necessary to let the consumer app in the end decide about the packages. So for packages we must use a different versioning-strategy than for apps in the dart ecosystem

td-famedly:
will wait for a dev to reopen

PR #2307, opened 2026-03-30
feat: custom rooms list sorter

coder-with-a-bushido:
Needed for https://github.com/famedly/product-management/issues/3582


krille-chan (approved):

krille-chan (changes requested):
Can you ake customRoomSorter private then? It should only be changed through the method now

krille-chan (approved):

PR #2308, opened 2026-03-31
fix(voip): allow updating `application` and stop timers for cancelled delayed events

yash-garg:


td-famedly (changes requested):
this will break compatibility with old calls now. You should still send application in the member event, but just remove the checks for it.

td-famedly:
please remove

`application == other.application &&`

from the callmem object as well

td-famedly (approved):
lg nice!

PR #2311, opened 2026-04-01
Fix flaky tests in CI + failing tests + issue on CI passing with failed tests

coder-with-a-bushido:
Another attempt since it cannot be reproduced locally

Also fixes CI green even when tests fail like in https://github.com/famedly/matrix-dart-sdk/actions/runs/23843600999/job/69505225066?pr=2307

image
image

### DIFF
--- a/test/device_keys_list_test.dart
+++ b/test/device_keys_list_test.dart
@@ -44,29 +44,49 @@ void main() async {
     });
 
     test('fromJson', () async {
+      const userId = '@alice:example.com';
+      const deviceId = 'JLAFKJWSCS';
+      final account = vod.Account();
+      final identityKeys = account.identityKeys;
+      final curve25519Key = identityKeys.curve25519.toBase64();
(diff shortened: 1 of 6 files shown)


test/device_keys_list_test.dart
```diff
@@ -44,29 +44,49 @@ void main() async {
...
-                'dSO80A01XiigH3uBiDVx/EjzaoycHcjq9lfQX0uWsqxl2giMIiSPR8a4d291W1ihKJL/a+myXS367WT6NAIcBA',
+          userId: {
+            'ed25519:$deviceId': signature,
           },
         },
         'unsigned': {'device_display_name': "Alice's mobile phone"},
       };
 
       final key = DeviceKeys.fromJson(rawJson, client);
-      // NOTE(Nico): this actually doesn't do anything, because the device signature is invalid...
```
td-famedly, line 69: is this note important?

td-famedly (approved):

PR #2317, opened 2026-04-20
feat: update ssss to centeralise key logic

ab-famedly:
Added reusable public migration APIs in SDK SSSS:

- analyzeEncryptedSecrets()
- orderedCandidateKeyIds(...)
- migrateSecretsToKey(...)

Also added private SDK helper:

- _tryOpenAndUnlockKey(...)

These now contain the logic to centralised (secret discovery, key iteration, migration, cache warm-up via maybeCacheAll()).

## Summary by CodeRabbit

* **New Features**
  * Improved detection of recovery keys vs passphrases and smarter unlock ordering.
  * Expanded secret-storage management: resolve named keys, analyze/enumerate encrypted secrets, rank candidate keys, and migrate secrets between keys with optional cleanup.

* **Bug Fixes**
  * More reliable secret-storage key validation and account-data sync alignment to avoid local/state drift during key operations.
  * Improved handling of cross-signing during reconnects.

* **Tests**
  * Added tests for recovery-key detection, migration scenarios, and malformed/invalid encrypted entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/famedly/matrix-dart-sdk/pull/2317?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/encryption/ssss.dart
+++ b/lib/encryption/ssss.dart
@@ -152,6 +152,12 @@ class SSSS {
     );
   }
 
+  /// Whether [input] looks like an Olm/SSSS recovery key (not a passphrase).
+  static bool looksLikeRecoveryKey(String input) {
+    final cleaned = input.replaceAll(RegExp(r'\s+'), '');
+    return cleaned.length == 48 && cleaned.startsWith('Es');
+  }
+
   static String encodeRecoveryKey(Uint8List recoveryKey) {
     final keyToEncode = <int>[...olmRecoveryKeyPrefix, ...recoveryKey];
     final parity = keyToEncode.fold<int>(0, (a, b) => a ^ b);
@@ -197,11 +203,55 @@ class SSSS {
       .key;
 
   Future<void> setDefaultKeyId(String keyId) async {
-    await client.setAccountData(
-      client.userID!,
+    await _setAccountDataAndWaitForSync(
       EventTypes.SecretStorageDefaultKey,
       SecretStorageDefaultKeyContent(key: keyId).toJson(),
     );
+  }
+
+  /// PUTs account data, then waits until [/sync] has applied the same payload so
+  /// [Client.accountData] and the DB stay aligned (both are updated in
+  /// [Client]'s sync handler). [MatrixApi.setAccountData] alone does not touch
+  /// local state; mirroring the PUT locally would race concurrent sync updates.
+  Future<void> _setAccountDataAndWaitForSync(
+    String type,
+    Map<String, Object?> content,
+  ) async {
+    final expected = content.copy();
+    await client.setAccountData(client.userID!, type, content);
+    await _waitForAccountDataFromSync(type, expected);
+  }
+
+  Future<void> _waitForAccountDataFromSync(
+    String type,
+    Map<String, Object?> expectedContent,
+  ) async {
+    bool matchesExpected() {
+      final ev = client.accountData[type];
+      if (ev == null) return false;
+      return const DeepCollectionEquality().equals(ev.content, expectedContent);
+    }
+
+    if (matchesExpected()) return;
+
+    final completer = Completer<void>();
+    final subscription = client.onAccountData.stream.listen((event) {
+      if (event.type == type && matchesExpected()) {
+        if (!completer.isCompleted) completer.complete();
+      }
+    });
+    try {
+      if (matchesExpected()) return;
+      await completer.future.timeout(
+        const Duration(seconds: 60),
+        onTimeout: () => throw TimeoutException(
+          'Timed out waiting for account data "$type" from sync after '
+          'setAccountData.',
+        ),
+      );
+    } finally {
+      await subscription.cancel();
+    }
   }
 
   SecretStorageKeyContent? getKey(String keyId) {
@@ -257,16 +307,7 @@ class SSSS {
     final accountDataTypeKeyId = EventTypes.secretStorageKey(keyId);
     // noooow we set the account data
 
-    await client.setAccountData(
-      client.userID!,
-      accountDataTypeKeyId,
-      content.toJson(),
-    );
-
-    while (!client.accountData.containsKey(accountDataTypeKeyId)) {
-      Logs().v('Waiting accountData to have $accountDataTypeKeyId');
-      await client.oneShotSync();
-    }
+    await _setAccountDataAndWaitForSync(accountDataTypeKeyId, content.toJson());
 
     final key = open(keyId);
     await key.setPrivateKey(privateKey);
@@ -276,9 +317,13 @@ class SSSS {
   Future<bool> checkKey(Uint8List key, SecretStorageKeyContent info) async {
     if (info.algorithm == AlgorithmTypes.secretStorageV1AesHmcSha2) {
       if ((info.mac is String) && (info.iv is String)) {
-        final encrypted = await encryptAes(zeroStr, key, '', info.iv);
-        return info.mac!.replaceAll(RegExp(r'=+$'), '') ==
-            encrypted.mac.replaceAll(RegExp(r'=+$'), '');
+        return client.nativeImplementations.checkSecretStorageKey(
+          CheckSecretStorageKeyArgs(
+            key: key,
+            iv: info.iv!,
+            mac: info.mac!,
+          ),
+        );
       } else {
         // no real information about the key, assume it is valid
         return true;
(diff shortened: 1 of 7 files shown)


lib/encryption/ssss.dart
```diff
@@ -644,6 +644,127 @@ class SSSS {
     return null;
   }
 
+  /// Returns secret event types mapped to valid SSSS key ids.
+  ///
+  /// Only account data entries with a valid `encrypted` map shape are included.
+  Map<String, Set<String>> analyzeEncryptedSecrets() {
```
td-famedly, line 752: nice! @ab-famedly can we use them in the sdk as well? the deduplication only makes sense then really

ab-famedly: thanks @td-famedly i've updated bootstrap.dart

lib/encryption/ssss.dart
```diff
@@ -644,6 +644,143 @@ class SSSS {
...
+      if (key == null || !key.isUnlocked) continue;
+
+      for (final secretType in claimSecretsForKey(keyId)) {
+        final secret = await key.getStored(secretType);
+        await destinationKey.store(secretType, secret, add: true);
+        migratedSecretTypes.add(secretType);
+      }
+      if (remainingSecrets.isEmpty) break;
+    }
+
```
td-famedly, line 878: do we not call validateAndStripOtherKeys here? and then could this not be reused in newSsss function?

ab-famedly: good catch, thanks iv updated it!

lib/encryption/ssss.dart
```diff
@@ -644,6 +644,158 @@ class SSSS {
...
+    await destinationKey.maybeCacheAll();
```
td-famedly: Do we need this here? I think we cacheAll in validateAndStripMigratedSecrets anyway?

lib/encryption/ssss.dart
```diff
@@ -211,13 +211,26 @@ class SSSS {
...
       EventTypes.SecretStorageDefaultKey,
       SecretStorageDefaultKeyContent(key: keyId).toJson(),
     );
   }
 
+  /// PUTs account data and mirrors it into [Client.accountData] and the
+  /// local DB. [MatrixApi.setAccountData] does not update the in-memory map;
+  /// without this, SSSS reads stale `encrypted` key sets until sync arrives
+  /// (and verify/migration code can see superseded key ids).
+  Future<void> _pushAccountDataOptimistic(
```
td-famedly: This can result in databases having unmatching data because of race conditions, please wait for the data to come back from sync if you need to make sure the client has the data locally. 

i.e setAccountData then wait until client.accountData has the data (it will at some point because of the sync loop)

lib/encryption/ssss.dart
```diff
@@ -271,16 +284,7 @@ class SSSS {
     final accountDataTypeKeyId = EventTypes.secretStorageKey(keyId);
     // noooow we set the account data
 
-    await client.setAccountData(
-      client.userID!,
-      accountDataTypeKeyId,
-      content.toJson(),
-    );
-
-    while (!client.accountData.containsKey(accountDataTypeKeyId)) {
```
td-famedly, line 266: @ab-famedly so basically this part ^

lib/encryption/ssss.dart
```diff
@@ -428,17 +437,21 @@ class SSSS {
...
+    final otherKeys = Set<String>.from(
+      encryptedContent.keys.where(
+        (k) => isDefaultKey || defaultKeyId == null
+            ? k != keyId
+            : k != keyId && k != defaultKeyId,
+      ),
+    );
     encryptedContent.removeWhere((k, v) => otherKeys.contains(k));
-    // yes, we are paranoid...
+    content['encrypted'] = encryptedContent;
```
td-famedly, line 475: eh why are we doing this? (also the previous comment was helpful so someone does not remove the getStored check)

lib/encryption/ssss.dart
```diff
@@ -671,6 +726,25 @@ class SSSS {
     return secrets;
   }
 
+  /// Returns, per secret account-data type, key ids in the `encrypted` map
+  /// that are not in [allowedKeyIds].
+  ///
+  /// Used to verify that stripping or migration left only expected keys (for
+  /// example default recovery plus one passphrase key). Types with no
+  /// extraneous keys are omitted.
+  Map<String, Set<String>> extraneousEncryptedSecretKeyIds(
```
td-famedly: I think this function is not that helpful. We only need this after stripping the keys so we can just make sure our stripping code is correct. We do not need client implementations to do another check

lib/encryption/ssss.dart
```diff
@@ -758,15 +832,40 @@ class SSSS {
     return migratedSecretTypes;
   }
 
+  /// Secret account-data types that [validateAndStripMigratedSecrets] will
```
td-famedly: I think `migratedSecretTypes` in `validateAndStripMigratedSecrets` already has the required keys. More importantly instead of having a public function to validateAndStripMigratedSecrets, we should just have a flag inside the `migrateSecretsToKey` method and then do it depending on that flag.

td-famedly (changes requested):
another review, looks like some of the methods are not really needed

lib/encryption/ssss.dart
ab-famedly: We need either “assign the mutated copy back into content” or “mutate the nested map in place,” because tryGetMap copies, stripping only the copy would upload unstripped account data. I've added the comment back though.

td-famedly:
@coderabbitai full review

lib/encryption/utils/bootstrap.dart
```diff
@@ -253,32 +224,16 @@ class Bootstrap {
...
+        final existingOldKeys = oldSsssKeys!;
+        final primaryUnlockedKey =
+            existingOldKeys[encryption.ssss.defaultKeyId] ??
+                existingOldKeys.values.first;
+        await encryption.ssss.migrateSecretsToKey(
+          primaryUnlockedKey: primaryUnlockedKey,
+          destinationKey: newSsssKey!,
+          candidateOldKeys: existingOldKeys,
+          stripKeys: true,
+        );
```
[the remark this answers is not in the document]
ab-famedly: resolved in commit 84b65d5

lib/encryption/ssss.dart
```diff
@@ -630,6 +684,235 @@ class SSSS {
...
+    if (keyIds.isEmpty) return {};
+
+    final migratedSecretTypes = <String>{};
+    Set<String> claimSecretsForKey(String keyId) {
+      final claimed = remainingSecrets.entries
+          .where((entry) => entry.value.contains(keyId))
+          .map((entry) => entry.key)
+          .toSet();
+      remainingSecrets.removeWhere((_, keys) => keys.contains(keyId));
+      return claimed;
```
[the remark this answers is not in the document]
ab-famedly: resolved in commit 84b65d5

lib/encryption/ssss.dart
```diff
@@ -630,6 +684,235 @@ class SSSS {
...
+  bool hasInvalidEncryptedEntries(String type) {
+    final encryptedContent = client.accountData[type]?.content
+        .tryGetMap<String, Object?>('encrypted');
+    if (encryptedContent == null) return false;
+
+    for (final keyEntry in encryptedContent.entries) {
+      final key = keyEntry.key;
+      final value = keyEntry.value;
+      if (value is! Map) continue;
+      if (!_isUsableEncryptedKeyEntry(key, value)) return true;
```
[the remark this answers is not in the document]
ab-famedly: resolved in commit 84b65d5

lib/encryption/ssss.dart
```diff
@@ -630,6 +684,235 @@ class SSSS {
...
+      }
+    }
+    if (candidates.isEmpty) return null;
+
+    final usedKeyIds =
+        analyzeEncryptedSecrets().values.expand((s) => s).toSet();
+    if (candidates.length == 1) {
+      final id = candidates.first;
+      return usedKeyIds.contains(id) ? id : null;
+    }
```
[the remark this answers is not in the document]
ab-famedly: resolved in commit 84b65d5

td-famedly:
@coderabbitai full review

td-famedly (approved):
lg

PR #2318, opened 2026-04-21
feat: gate LaTeX markdown parsing behind enableLatexMarkdown flag

kamor-famedly:
## Description
Messages containing `$` were incorrectly treated as LaTeX math delimiters and rendered as `<code>` elements with double-escaped HTML entities.

## Changes
- Added `enableLatexMarkdown` flag on `Client` (defaults to `true`) to control LaTeX parsing
- Gated `InlineLatexSyntax` and `BlockLatexSyntax` behind this flag
- Added regression test to ensure proper gating behavior

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -114,6 +114,8 @@ class Client extends MatrixApi {
 
   final bool convertLinebreaksInFormatting;
 
+  final bool enableLatexMarkdown;
+
   final Duration sendTimelineEventTimeout;
 
   /// The timeout until a typing indicator gets removed automatically.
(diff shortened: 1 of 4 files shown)


td-famedly (changes requested):
we should still have this in the sdk, but only remove it for tim clients, so this should somehow go either into the tim sdk or the app

kamor-famedly:
@td-famedly Which one is it going to be?

Would it be useful in anyway to have it on matrix sdk behind a flag?

td-famedly (approved):
lg

PR #2319, opened 2026-04-22
fix: Do not send dozens of key request to device messages on timeline…

krille-chan:
… export

This is super annoying. When we do a (legacy) search or a timeline export, we should not send dozens of key request to device messages to all other devices on a room export. It makes no sense. If the key is not in the key backup we don't have it and key requests shouldn't be sent automatically for an unknown amount of messages. `decryptRoomEvent()` already fetches the key from the key backup


td-famedly (approved):
oki

PR #2322, opened 2026-04-28
fix: preserve call rejection room previews

yash-garg:
Keep MatrixRTC member cleanup events from overwriting rejected-call previews so room lists show the semantic call outcome.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -3201,6 +3201,8 @@ class Client extends MatrixApi {
         // Is this event of an important type for the last event?
         if (!roomPreviewLastEvents.contains(event.type)) break;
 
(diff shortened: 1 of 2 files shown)


lib/src/client.dart
```diff
@@ -3213,6 +3215,34 @@ class Client extends MatrixApi {
     room.onUpdate.add(room.id);
   }
 
+  bool _shouldKeepCallRejectLastEvent(Event event, Event? lastEvent) {
```
td-famedly, line 3225: Can you add a comment on why we need to do this? apart from that lgtm!

td-famedly (approved):

lib/src/client.dart
yash-garg: done

td-famedly:
@yash-garg please fix the tests

PR #2328, opened 2026-05-05
Move heavy getUserDeviceKeys processing off the UI thread

Azare77:
This PR moves the expensive processing inside getUserDeviceKeys
to a background isolate to avoid blocking the main/UI thread.

Large device key datasets previously caused noticeable UI freezes
during startup and sync operations.

Changes:
- Offloaded heavy parsing/grouping work to Isolate.run
- Kept Matrix client object construction on the main isolate
- Reduced UI stutter during device key loading


td-famedly:
We plan to refactor this code very soon, so closing this PR until then

PR #2332, opened 2026-05-07
feat: add matrix content scanner support

kamor-famedly:
https://github.com/famedly/product-management/issues/3861
This PR implements support for routing media downloads through a Matrix content scanner proxy.

  When `contentScannerConfig` is set, MXC download and thumbnail URLs are resolved against the scanner endpoints. `downloadAndDecryptAttachment()` now also supports scanner-backed downloads for encrypted
  attachments by posting the encrypted file metadata to `download_encrypted`, then decrypting the returned bytes client-side as before.
  
  The implementation was tested end-to-end against our staging server using the local scanner proxy tool at `tool/content_scanner_proxy.dart`.

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -659,7 +659,7 @@ class Event extends MatrixEvent {
   /// Throws an exception if the scheme is not `mxc` or the homeserver is not
   /// set.
   ///
-  /// Important! To use this link you have to set a http header like this:
+  /// Scanner and authenticated media URLs may need an authorization header:
   /// `headers: {"authorization": "Bearer ${client.accessToken}"}`
   Future<Uri?> getAttachmentUri({
     bool getThumbnail = false,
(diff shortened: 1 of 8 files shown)


krille-chan (changes requested):
@kamor-famedly can you link an issue here?

test/tool/content_scanner_proxy.dart
```diff
@@ -0,0 +1,605 @@
+#!/usr/bin/env dart
+// ignore_for_file: avoid_print
```
krille-chan, line 6: Would it make sense to move this to /test/tool? So we don't introduce another root level directory

krille-chan (approved):

lib/src/event.dart
```diff
@@ -833,9 +841,7 @@ class Event extends MatrixEvent {
     getThumbnail = mxcUrl != attachmentMxcUrl;
     final isEncrypted =
         getThumbnail ? isThumbnailEncrypted : isAttachmentEncrypted;
-    if (isEncrypted && !room.client.encryptionEnabled) {
-      throw ('Encryption is not enabled in your Client.');
-    }
+    // File attachments use per-file keys, independent of room encryption state.
```
td-famedly: I do not see why this was removed

kamor-famedly: I treated file encryption as seperate from room/events encryption.
I think this is more appropriate

if (isEncrypted && !room.client.fileEncryptionEnabled) {
    throw ('Encryption is not enabled in your Client.');
  }

td-famedly (approved):
lg as well

PR #2338, opened 2026-05-12
fix: do not send notify rtc event in direct chats

td-famedly:


yash-garg (approved):

td-famedly:
dangling tag, but is probably fine

PR #2339, opened 2026-05-13
fix: Workaround that synapse's initial sync account data may not be up to date

krille-chan:
https://famedly.atlassian.net/browse/CS-1623


td-famedly (commented):
well that's a annoying bug but this makes sense to me. We could have a test for this. will approve once ready

krille-chan:
> well that's a annoying bug but this makes sense to me. We could have a test for this. will approve once ready

It is already covered by the existing tests as the account data is fetched every time before the app does an initial bootstrap.

td-famedly (approved):
oki

PR #2343, opened 2026-05-20
fix(CachedStreamController): set _lastError with error parameter instead of value property

Feichtmeier:
I noticed that the `addError` method of the `CachedStreamController` class sets the current `_value` as the `_lastError` instead of using the provider `error` parameter, I am not 100% sure if this is a desired fix but just in case here it is :) Thanks and best regards


td-famedly:
hmm I am almost scared to merge it because someone might depend on it 🤔  Maybe it makes sense to store the error in another variable now and expose that and give a migration warning?

PR #2357, opened 2026-06-01
refactor(BREAKING): Require client to be initialized for getEventByPushNotification()

krille-chan:
Before we have (in theory) supported that the client does not need to be initialized to call client.getEventByPushNotification() with the intention to allow multiple isolates. However this has never been used (or tested) in any of our clients (thanks to fcm_shared_isolate) and probably never worked. Also suggesting to not need to initialize might more harm than actually been useful. Initializing a client is now possible without waiting for all rooms or other data to be loaded so can be done in very short time.

The current implementation also had the problem that it only loads the token, not the refresh token or refresh token expires in datetime. Instead of duplicating all these logic or abstracting it, I'd find it more clean to remove this "feature" and require to initialize the client before, avoiding confusion, for the sake of simplicity and more intuitive behavior of the SDK.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -1855,10 +1855,8 @@ class Client extends MatrixApi {
   /// Fetches the corresponding Event object from a notification including a
   /// full Room object with the sender User object in it. Returns null if this
   /// push notification is not corresponding to an existing event.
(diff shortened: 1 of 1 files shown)


krille-chan:
Will reevalute this

PR #2360, opened 2026-06-02
fix: Soft logout with Matrix native OIDC does not send soft_logout flag

krille-chan:


lib/src/client.dart
```diff
@@ -2504,7 +2513,14 @@ class Client extends MatrixApi {
         ),
       );
       if (e.error == MatrixError.M_UNKNOWN_TOKEN) {
-        if (e.raw.tryGet<bool>('soft_logout') == true) {
+        // The server explicitely moved the session into soft logout state:
+        final isSoftLogoutState = e.raw.tryGet<bool>('soft_logout') == true;
+        // The access token is expired, so can assume from client side we
+        // are in soft logout state (Matrix Native OIDC):
+        final accessTokenExpired =
+            accessTokenExpiresAt?.isAfter(DateTime.now()) == true;
```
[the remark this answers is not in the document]
krille-chan: Very good catch!

lib/src/client.dart
```diff
@@ -2367,11 +2367,20 @@ class Client extends MatrixApi {
...
-      } catch (e, s) {
-        Logs().w('Unable to refresh session after soft logout', e, s);
-        // cannot logout without a token, so we just clear our database
+      } on MatrixException catch (e) {
+        Logs().w(
+          'Unable to refresh session after soft logout. Clearing session...',
+          e,
+        );
         await clear();
         rethrow;
```
td-famedly, line 2376: so now the rethrow will just be caught? 🤔

lib/src/client.dart
```diff
@@ -2494,13 +2503,16 @@ class Client extends MatrixApi {
...
-        if (e.raw.tryGet<bool>('soft_logout') == true) {
-          Logs().w(
-            'The user has been soft logged out! Calling client.onSoftLogout() if present.',
-          );
+        // The server explicitely moved the session into soft logout state:
+        final isSoftLogoutState = e.raw.tryGet<bool>('soft_logout') == true;
+
+        final hasRefreshToken = accessTokenExpiresAt != null;
+
+        if (isSoftLogoutState || hasRefreshToken) {
```
td-famedly, line 2511: is it possible that the server sends soft_logout true but we do not have a `accessTokenExpiresAt`? I do not see why we need both of these here

td-famedly (changes requested):

lib/src/client.dart
td-famedly: ah I read this wrong

lib/src/client.dart
```diff
@@ -2367,11 +2367,20 @@ class Client extends MatrixApi {
...
+        );
         await clear();
         rethrow;
+      } catch (e, s) {
+        Logs().e(
+          'Unable to refresh session after soft logout. Try again...',
+          e,
+          s,
+        );
+        return;
```
td-famedly, line 2383: any reason we do not rethrow here? This will be a breaking change because clients might be depending on this behaviour. Can you also log the error and the stacktrace here please

lib/src/client.dart
krille-chan: This is not and but or. So in matrix native oidc the server sends us that the token is no longer valid but doesn't send the soft_logout: true. We have to check by ourself. However if the access token is no longer valid, but we have a refresh token, then it doesn't harm to at least try once to refresh before clearing the session in any case

lib/src/client.dart
krille-chan: If we rethrow, then we don't try again but instantly log out. This error here happens when for example network error happens. So soft logout + no network -> instant session lost! We are right now workarounding this in the famedly app by catching it there but it makes more sense to have this behavior by default in the sdk

td-famedly: ok but then can you mark it as a breaking change because this changes the API

lib/src/client.dart
td-famedly: hm oki

td-famedly (approved):
breaking change but lg!

PR #2361, opened 2026-06-03
fix: Allow image resizer to return new mime type

krille-chan:
Our image resizer changes the type of any image to jpg but the mimetype is never returned by it. This change allows it to also return a new mimetype which then gets correctly added to the image metadata and sent over matrix.

closes https://github.com/famedly/product-management/issues/3935

### DIFF
--- a/lib/src/utils/matrix_file.dart
+++ b/lib/src/utils/matrix_file.dart
@@ -200,7 +200,7 @@ class MatrixImageFile extends MatrixFile {
     final thumbnailFile = MatrixImageFile(
       bytes: resizedData.bytes,
       name: name,
-      mimeType: mimeType,
+      mimeType: resizedData.mimeType,
       width: resizedData.width,
       height: resizedData.height,
       blurhash: resizedData.blurhash,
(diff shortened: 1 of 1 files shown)


lib/src/utils/matrix_file.dart
yurtemre7, line 306:
```suggestion
'height': height,
if (mimeType != null) 'mimeType': mimeType,
if (blurhash != null) 'blurhash': blurhash,
```

yurtemre7 (changes requested):
Please add the mimeType on the toJson so we don't lose it :)!
Rest looks good, will approve.

lib/src/utils/matrix_file.dart
```diff
@@ -274,6 +277,7 @@ class MatrixImageFileResizedResponse {
     required this.bytes,
     required this.width,
     required this.height,
+    required this.mimeType,
```
yurtemre7:
```suggestion
   this.mimeType,
```

Maybe also remove the required keyword as the parameter is nullable.

krille-chan: Done :)

lib/src/utils/matrix_file.dart
krille-chan: Hah! Good catch! :-)

yurtemre7 (approved):

PR #2372, opened 2026-06-17
fix: null check exception on room.ownPowerLevel

mErrenst:
This should fix FM-625 ( https://famedly-ip.sentry.io/issues/7433903576/?environment=production&environment=prod&environment=staging&project=1448986&query=is%3Aunresolved&referrer=issue-stream )


yurtemre7 (approved):
Good null check, thank you.

PR #2391, opened 2026-06-29
fix: selfSign() with passphrase fails sometimes when using existing ssss instead of just the passphrase

krille-chan:
fixes FM-683


mErrenst (approved):
LGTM

PR #2396, opened 2026-07-07
docs: add voice and video calling guide

### LINKED ISSUE #2329: Documentation + some questions
### Preflight Checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions

### Describe your problem

Hi,

I'm implementing a calling module around Matrix chats using this SDK + LiveKit, and I'd like to flag that the documentation around calls / MatrixRTC is currently very hard to use as an integrator. There are API references and scattered mentions, but no end-to-end guide that walks through a real client implementation, so I've been reverse-engineering from the source and the Element clients.

Rather than just complaining, here are the concrete questions I got stuck on — answers would both unblock me and, ideally, become the basis for a docs page:

**1. Call lifecycle**
- What is the recommended entry point to start a call against a 1:1 room? Is it through `MatrixRTCSession` / `GroupCallSession` / something else?
- What's the correct sequence: create session → join → publish tracks → ... ? 
- How do I cleanly tear a call down on hangup vs. on disconnect vs. on app kill?

**2. State tracking**
- Which streams or listeners expose call state transitions (ringing → connecting → connected → ended)?
- How do I observe remote participants joining/leaving, and their media state (mute/unmute, video on/off)?
- Is there a single object I should treat as the source of truth, or do I need to combine multiple?

**3. Incoming-call signalling**
- What event do I send to tell the callee "I'm calling you" so their client can show an incoming-call screen?
- What should the receiving client listen for to detect an incoming call?
- How is this expected to integrate with push notifications (VoIP push / FCM data messages) when the app is backgrounded or killed?

### Describe your ideal solution

A "Building calls with LiveKit" guide with a minimal, runnable 1:1 call example covering: starting a MatrixRTC call backed by LiveKit, signalling the callee over Matrix so an incoming-call UI can be shown before joining the LiveKit room, observing both Matrix call state and LiveKit room/participant events, publishing and muting media tracks, and cleanly hanging up — including how this is expected to work alongside push notifications when the app is backgrounded.

### Version

matrix: ^6.2.0

### Security requirements

_No response_

### Additional Context

livekit_client: ^2.4.5

saylee21:
Closes #2329.

Adds doc/calling.md — a minimal guide for 1:1 voice and video calls
using flutter_webrtc.

Signed-off-by: Saylee Bharsakle <sayleebharsakle@gmail.com>

### DIFF
--- a/doc/calling.md
+++ b/doc/calling.md
@@ -6,7 +6,10 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 
 # Voice and Video Calls
 
-This guide walks through adding 1:1 voice and video calls to a Flutter app using `flutter_webrtc`. For group calls and LiveKit integration, see the [VoIP module README](https://github.com/famedly/matrix-dart-sdk/tree/main/lib/src/voip).
+This guide walks through adding voice and video calls to a Flutter app using the MatrixRTC API.
+It covers both P2P mesh calls (via `MeshBackend`) and SFU-backed group calls (via `LiveKitBackend`).
+
+> **Note:** The legacy `WebRTCDelegate` / `CallSession` APIs are deprecated. New projects should use the MatrixRTC event stream described here.
 
 ## Add dependencies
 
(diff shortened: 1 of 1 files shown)


td-famedly (changes requested):
I would suggest adding the MatrixRTC version of this not the legacy calls which are deprecated

saylee21:
> I would suggest adding the MatrixRTC version of this not the legacy calls which are deprecated

Updated the guide to use MatrixRTC — GroupCallSession, matrixRTCEventStream, MeshBackend/LiveKitBackend, and MSC4075 notifications. Please take another look.

PR #2400, opened 2026-07-08
fix: clear LiveKit key debounce timestamp on dispose

marco-a-affinidi:
If a LiveKit backend is reused within `makeKeyOnJoinDelay` after dispose, `preShareKey()` may try to resend a recent key even though local key storage was already cleared. That null-key reuse path recurses through key generation and sending until it overflows the stack.

Reset `_lastNewKeyTime` on dispose so rejoining generates a fresh key.


td-famedly (approved):
good catch!

marco-a-affinidi:
> good catch!

Thanks @td-famedly, looks like auto-merge did not work, do I need to do anything on my side?

td-famedly:
@marco-a-affinidi looks like it needs a rebase so you are not outdated with main

marco-a-affinidi:
Thanks @td-famedly, it's now rebased

PR #2402, opened 2026-07-08
fix: do not assume room member count is 1 if absent

krille-chan:
Assuming this is somehow dirty and we should more transparent return null here instead of guessing. There are actually cases where synapse or another homeserver is not giving us the member count (yet).

This fixes that the sdk sometimes does not fetch keys for a newly created room.

fixes https://famedly.atlassian.net/browse/FM-736


ab-famedly (approved):
lgtm

td-famedly (approved):
oki

PR #2404, opened 2026-07-10
feat: (BREAKING) advanced TOFU

krille-chan:


krille-chan:
We should also sign it with a fourth cross signing key

PR #2405, opened 2026-07-10
fix: avoid concurrent modification when updating group calls

marco-a-affinidi:
The VoIP room-state listener iterated directly over `groupCalls.entries` while awaiting `onMemberStateChanged()`, which can mutate the same map during membership processing and trigger `Concurrent modification during iteration`, aborting updates for remaining calls. Iterate over a snapshot instead and add a regression test that reproduces the mutation and verifies the listener continues processing safely.


td-famedly (approved):
oki

PR #2407, opened 2026-07-10
fix: get user device keys for gappy syncs and partial rooms

td-famedly:


test/new_room_first_message_repro_test.dart
```diff
@@ -0,0 +1,707 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+// Reproduction tests for: "first message in a new room is undecryptable
+// (The sender has not sent us the session key.)"
+//
+// These tests emulate the sender side of the flow as faithfully as possible:
+// every simulated /sync is processed inside a database transaction and is
+// followed by `updateUserDeviceKeys()`, exactly like `Client._sync()` does.
+
+import 'dart:convert';
```
krille-chan, line 12: are we sure we need so many LoC for the test and this is even testing what we want? Looks like a total overkill. We should also keep tests not too big so that they are maintainable

analysis_options.yaml
```diff
@@ -8,6 +8,7 @@ linter:
   rules:
     # TODO: Enable these lints one after another and fix them
     non_constant_identifier_names: false
+    directives_ordering: true
```
krille-chan: Doesn't this come from our templates?

lib/src/client.dart
```diff
@@ -2617,6 +2617,13 @@ class Client extends MatrixApi {
         if (userKeys != null) {
           userKeys.outdated = true;
           await database.storeUserDeviceKeysInfo(userId, true);
+        } else {
+          // Per spec `changed` also includes users who *now* share an
+          // encrypted room with us. We might not know about them yet, for
+          // example when their membership event was truncated out of a
+          // limited (gappy) sync timeline. Track them so that the next
+          // `updateUserDeviceKeys()` fetches their device keys.
+          _trackedUserIds?.add(userId);
```
krille-chan, line 2626: Which ticket does this fix? I have applied a fix already for a similar thing and I don't think we need this here. In fact it might even harm

lib/src/client.dart
```diff
@@ -3134,17 +3141,21 @@ class Client extends MatrixApi {
...
+        if (stateKey != null) {
+          if (!room.partial || importantStateEvents.contains(event.type)) {
+            room.setState(event);
+          }
 
+          // New members must be added to the tracked user IDs for encryption
+          // even if the room is only partially loaded, as otherwise their
+          // device keys are never fetched and they never receive room keys:
           if (room.encrypted &&
+              room.membership == Membership.join &&
```
krille-chan, line 3153: I don't get this change and the comment above. We already have added user IDs for encrypted rooms even when the room was partially loaded. But now we add another condition that we have already joined the room, but why not track user IDs before anymore? So in fact with this change we track less user IDs than before

lib/src/room.dart
```diff
@@ -1388,7 +1389,12 @@ class Room {
           // to a direct chat and then User A leaves the chat before User B
           // joined.
           // See: https://github.com/element-hq/synapse/issues/1533
-          exception.error == MatrixError.M_UNKNOWN) {
+          e.error == MatrixError.M_UNKNOWN) {
+        Logs().w(
+          'Tried to join room which is unknown to server, will leave',
```
krille-chan: Unrelated?

krille-chan: also we should not print this to the console if we rethrow it anyway. Application must handle this

lib/src/room.dart
```diff
@@ -2590,13 +2596,30 @@ class Room {
...
+    // able to decrypt any message in this room.
+    final missingUserIds = users
+        .map((user) => user.id)
+        .where((userId) => client.userDeviceKeys[userId] == null)
+        .toSet();
+    if (missingUserIds.isNotEmpty) {
+      Logs().i(
+        '[E2EE] Fetching device keys of ${missingUserIds.length} untracked users in $id before encrypting',
+      );
+      await client.updateUserDeviceKeys(additionalUsers: missingUserIds);
```
krille-chan, line 2610: We are already doing this in the sync loop. Why do we need this here? This is duplicated logic and even slows down everything. We should do it correctly in one place and not do it multiple times partly in different places. If it is laggy somewhere we should just fix it

analysis_options.yaml
td-famedly: yes but I needed this because my linter won't tell me about failing analyzer and then I have to depend on the ci, do we have a solution for this?

test/new_room_first_message_repro_test.dart
td-famedly: I'll try to minimize this, I will ask the synapse team which of these cases are possible, but if they are all required this test file makes sense

lib/src/client.dart
td-famedly: So what I see is we only track partial room users when _trackUserIds is already null and then we defer to _getUserIdsInEncryptedRooms, otherwise we will never track partial rooms correct? _getUserIdsInEncryptedRooms is called in updateUserDeviceKeys which is called in innerSync but _handleSync is called before this which can call one of the functions which updates _trackUserIds and then we are stuck?

lib/src/room.dart
td-famedly: Need this for debugging and still think this is potentially useful

lib/src/client.dart
td-famedly: Regarding your second point, we do not need to track leave and invited rooms userids? why is this wrong?

lib/src/room.dart
td-famedly: This is again because we cannot be sure if _trackUserIds was null and that _getUserIdsInEncryptedRooms was called

td-famedly: If we merge the other fix, and synapse team confirms I will not miss member events from gappy syncs I can remove this part, but tbh I think this will be more helpful rather than harmful

analysis_options.yaml
krille-chan: my linter does and works fine with it. Clear cache or smth?

lib/src/room.dart
krille-chan: Why is it useful to log something which you rethrow anyway?

lib/src/client.dart
krille-chan: but this doesn't change anything about it as member events are always important state events

td-famedly: >  member events are always important state events

they are not

lib/src/client.dart
td-famedly: Discussed in PMs this is an actual fix. The ticket is the same, decrypting some messages is broken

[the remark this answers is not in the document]
td-famedly: If it's null updateUserDeviceKeys() gets triggered and therefore it's recomputed anyway so this is not a bug.

krille-chan (approved):

lib/src/client.dart
nico-famedly, line 3153: Will this not break getting keys for invites? Basically when you are invited to a room you also receive the room keys for the messages in it. You will need the device keys to decrypt those. Does that still work with that?

lib/src/client.dart
```diff
@@ -2617,6 +2617,13 @@ class Client extends MatrixApi {
         if (userKeys != null) {
           userKeys.outdated = true;
           await database.storeUserDeviceKeysInfo(userId, true);
+        } else {
+          // Per spec `changed` also includes users who *now* share an
+          // encrypted room with us. We might not know about them yet, for
+          // example when their membership event was truncated out of a
+          // limited (gappy) sync timeline. Track them so that the next
```
nico-famedly, line 2624: The state should still be in the state section of the sync response though.

td-famedly: we only listen to device_lists change here, the other way would be a much bigger refactor

lib/src/client.dart
td-famedly: _trackedUserIds is only used for outbound messages afaik

PR #2412, opened 2026-07-14
fix: make login types check optional

### LINKED ISSUE #3250: Native OIDC-only homeservers (Continuwuity v26.6.0 and above) cannot be logged into, even with enableMatrixNativeOIDC enabled
### Bug Description

When attempting to log into a matrix server using a native OIDC-only implementation - like a Continuwuity server running version 26.6.0 and above with OIDC configured - regardless of the state of the enableMatrixNativeOIDC toggle, attempting to log into the server will result with "User-interactive authentication is not available on this server". This is due to the fact that if OIDC is configured, Continuwuity stops serving UIAA endpoints that point to its authentication endpoints.

### Steps to Reproduce

1. Try logging into a Continuwuity v26.6.0+ Matrix homeserver (hopeserv.net is a test instance I can lend)
2. Observe the issue.

### Expected Behavior

The client checks for the Native OIDC endpoints and allows logging in even if the UIAA endpoints are missing.

### App Version

2.7.2

### Platform

Android (PlayStore)

### Additional Platform Information

POCO X7 PRO

### Additional Context

_No response_

krille-chan:
check for login types is soon legacy duo to matrix native oidc. Making this a requirement actually makes login to oidc only homeservers like newest version of conduwuit impossible.

closes https://github.com/krille-chan/fluffychat/issues/3250

closes https://famedly.atlassian.net/browse/FM-757

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -8,10 +8,11 @@ import 'dart:core';
 import 'dart:math';
 import 'dart:typed_data';
 
+import 'package:async/async.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:http/http.dart' as http;
 import 'package:matrix/encryption.dart';
-import 'package:matrix/matrix.dart';
+import 'package:matrix/matrix.dart' hide Result;
 import 'package:matrix/matrix_api_lite/generated/fixed_model.dart';
 import 'package:matrix/msc_extensions/msc_unpublished_custom_refresh_token_lifetime/msc_unpublished_custom_refresh_token_lifetime.dart';
 import 'package:matrix/src/models/timeline_chunk.dart';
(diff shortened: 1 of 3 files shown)


N00byKing:
This change alone was not enough for me to get it to work, I also needed #2411.
Was that not the case for you?

krille-chan:
> This change alone was not enough for me to get it to work, I also needed #2411. Was that not the case for you?

I haven't tried it on web so wasn't affected. I will review your PR :)

lib/src/client.dart
```diff
@@ -576,8 +577,15 @@ class Client extends MatrixApi {
         assert(false);
       }
 
-      final loginTypes = await getLoginFlows() ?? [];
-      if (!loginTypes.any((f) => supportedLoginTypes.contains(f.type))) {
+      final loginTypesResult = await Result.capture(getLoginFlows());
+      final loginTypes = loginTypesResult.asValue?.value ?? [];
+      final loginTypesError = loginTypesResult.asError?.error;
+      if (loginTypesError != null && loginTypesError is! MatrixException) {
+        Logs().w('Unable to fetch legacy login types', loginTypesError);
```
coder-with-a-bushido: Do we not want to throw error here?

lib/src/client.dart
```diff
@@ -576,8 +577,15 @@ class Client extends MatrixApi {
...
+      }
```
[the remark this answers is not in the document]
krille-chan: fixed

coder-with-a-bushido (approved):

lib/src/client.dart
krille-chan: We do in the newest version :)

PR #2413, opened 2026-07-14
fix: allow redact-power users to end polls (MSC3381)

Quantumheart:
## Summary

I am working on a Matrix client called [Kohera](https://github.com/Quantumheart/Kohera). I am currently, working through the [polling spec 3381](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3381-polls.md) and found that the power level redaction is not respected for moderators.

## Changes
* add !room.canRedact
* add test "endPoll: creator may end, non-creator without redact power may not, redact power may end"

### DIFF
--- a/lib/msc_extensions/msc_3381_polls/poll_event_extension.dart
+++ b/lib/msc_extensions/msc_3381_polls/poll_event_extension.dart
@@ -120,11 +120,23 @@ extension PollEventExtension on Event {
     );
   }
 
+  /// Ends this poll by sending an `m.poll.end` reference event into the
+  /// room.
+  ///
+  /// Per MSC3381, a poll may be ended by its creator or by a user with
+  /// permission to redact other users' messages in the room (the redaction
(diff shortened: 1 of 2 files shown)


krille-chan (approved):
From the MSC:

> If a m.poll.end event is received from someone other than the poll creator or user with permission to redact other's messages in the room, the event must be ignored by clients due to being invalid.

So you are right :) I also checked that we are already correctly parsing it manually, just this check was missing. Thank you very much

krille-chan:
@Quantumheart you need to sign your commit and make sure you accepted the CLA (the cla bot may be prompted once you signed your commit)

Quantumheart:
Resolved in 6c5ce3e — the denial-case `expectLater` is now awaited so the matcher completes before the redact-power injection. Test passes locally.

Quantumheart:
@krille-chan are you able to approve the workflow? Thanks in advance.

PR #2416, opened 2026-07-20
feat: add customVideoThumbnailGenerator callback to create video thumbnails on send

td-famedly:
### Description

Part of [FM-567](https://famedly.atlassian.net/browse/FM-567) (no video thumbnails when sending from web).

The SDK can not decode videos itself, so `m.video` messages are sent without a thumbnail unless the app passes one explicitly (which the Famedly app only does on mobile via `video_compress`). This adds a `customVideoThumbnailGenerator` callback to `Client`, analogous to `customImageResizer`:

- `Room.sendFileEvent` invokes the callback for a `MatrixVideoFile` when no explicit `thumbnail` is provided, uploads the generated thumbnail and writes `thumbnail_url`/`thumbnail_file` + `thumbnail_info` into the event.
- The response can also carry the video's own metadata, which is used to fill in missing `w`/`h`/`duration` in the video info.
- The thumbnail's blurhash is written to the video event's `info` (`xyz.amorgan.blurhash`), as already done for images.
- A failing generator is logged and the video is sent without a thumbnail, like the image path.

The app-side browser implementation is in famedly/app (uses an `HTMLVideoElement` + canvas).

### Verification

- New tests for the happy path (thumbnail + metadata + blurhash in the sent event) and for a throwing generator.
- Full test suite passes (54/54 files), `dart analyze` and `dart format` clean.

<div><a href="https://cursor.com/agents/bc-7e8c8b3b-c66f-4bc5-a031-13cc1f17d4a1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7e8c8b3b-c66f-4bc5-a031-13cc1f17d4a1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[FM-567]: https://famedly.atlassian.net/browse/FM-567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -111,6 +111,9 @@ class Client extends MatrixApi {
   )?
   customImageResizer;
 
+  Future<MatrixVideoThumbnailResponse?> Function(MatrixVideoThumbnailArguments)?
+  customVideoThumbnailGenerator;
+
   /// Optional matrix-content-scanner proxy configuration.
   ///
   /// When set, media URL helpers resolve `mxc://` URIs to scanner URLs, and
@@ -159,8 +162,9 @@ class Client extends MatrixApi {
   /// enable the SDK to compute some code in background.
   /// Set [timelineEventTimeout] to the preferred time the Client should retry
   /// sending events on connection problems or to `Duration.zero` to disable it.
-  /// Set [customImageResizer] to your own implementation for a more advanced
-  /// and faster image resizing experience.
+  /// Set [customImageResizer] and [customVideoThumbnailGenerator] to your own
+  /// implementations for a more advanced and faster image resizing experience
+  /// and for generating video thumbnails, which the SDK can not do itself.
   /// Set [enableDehydratedDevices] to enable experimental support for enabling MSC3814 dehydrated devices.
   Client(
     this.clientName, {
@@ -190,6 +194,7 @@ class Client extends MatrixApi {
     Duration defaultNetworkRequestTimeout = const Duration(seconds: 35),
     this.sendTimelineEventTimeout = const Duration(minutes: 1),
     this.customImageResizer,
+    this.customVideoThumbnailGenerator,
     this.shareKeysWith = ShareKeysWith.crossVerifiedIfEnabled,
     this.enableDehydratedDevices = false,
     this.receiptsPublicByDefault = true,
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -17,6 +17,7 @@ import 'package:matrix/src/utils/file_send_request_credentials.dart';
 import 'package:matrix/src/utils/markdown.dart';
 import 'package:matrix/src/utils/marked_unread.dart';
 import 'package:matrix/src/utils/space_child.dart';
+import 'package:mime/mime.dart';
 
 /// max PDU size for server to accept the event with some buffer incase the server adds unsigned data f.ex age
 /// https://spec.matrix.org/v1.9/client-server-api/#size-limits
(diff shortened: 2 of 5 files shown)


td-famedly:
@cursor review

lib/src/utils/matrix_file.dart
```diff
@@ -353,6 +353,54 @@ class MatrixVideoFile extends MatrixFile {
...
+class MatrixVideoThumbnailArguments {
+  /// The bytes of the video to generate a thumbnail for.
+  final Uint8List bytes;
+  final String fileName;
+  final String? mimeType;
+
+  /// The requested maximum width and height of the thumbnail.
+  final int maxDimension;
+
+  const MatrixVideoThumbnailArguments({
```
krille-chan, line 362: Should we use Records instead of classes here?

lib/src/client.dart
```diff
@@ -190,6 +198,7 @@ class Client extends MatrixApi {
     Duration defaultNetworkRequestTimeout = const Duration(seconds: 35),
     this.sendTimelineEventTimeout = const Duration(minutes: 1),
     this.customImageResizer,
+    this.customVideoThumbnailGenerator,
```
krille-chan, line 197: We use this method only in the `Room.sendFileEvent()` event and nowhere else. Instead of making it a parameter here, shouldn't we move it to a parameter of this method instead maybe?
Just a thought. I know this would be inconsistent to the customImageResizer. Maybe it makes sense in a later refactoring to also move the customImageResizer to this method out of the Client?

krille-chan:
Looks very good :)

lib/src/client.dart
td-famedly: let's do this in a later refactoring, don't want to cause a breaking change here

lib/src/utils/matrix_file.dart
td-famedly: given the number of fields I think records would be really cumbersome

td-famedly: They also do not support default values for named classes

krille-chan: good point

lib/src/room.dart
```diff
@@ -925,6 +927,69 @@ class Room {
       }
     }
 
+    // computing a thumbnail for video files in case the client has set a
+    // generator for this, as the SDK can not decode videos itself:
```
krille-chan: I would make the comment more short:

> "Compute video thumbnail"

Code comments are generally something which should be avoided. They are a sign that the code is not self explaining and not readable enough. Code comments only should be the last action to make unreadable code a little bit easier to understand. However this if statement is readable enough. Just marking something like a header for this section should be enough IMO

lib/src/room.dart
```diff
@@ -925,6 +927,69 @@ class Room {
...
+          MatrixVideoThumbnailArguments(
+            bytes: file.bytes,
+            fileName: file.name,
+            mimeType: file.mimeType,
+          ),
+        );
+        if (videoThumbnail != null) {
+          // The thumbnail is an image, so it must not reuse the video's
+          // file name and extension:
+          final thumbnailMimeType = videoThumbnail.mimeType;
```
krille-chan, line 953: I don't get why we add this not very useful comment there and even this abstraction. We use it only in two different places below. Just use `videoThumbnail.mimeType` directly there? Where is the benefit of these extra LoC?

krille-chan (approved):

td-famedly:
@cursor push 6531cd0774

PR #2419, opened 2026-07-23
fix: do not guess a file extension for video thumbnails with unknown mimetype

td-famedly:
### Description

Follow-up to #2416, fixing a review finding that arrived after the merge.

Pending file events often have no `thumbnail_info` in their content yet — explicit thumbnails never write it before sending. When `Event.sendAgain` restored a cached thumbnail in that state, `thumbnailMimetype` was empty, `extensionFromMime` returned `null`, and string interpolation produced a literal `file.mp4.thumbnail.null` file name that got re-uploaded. `downloadAndDecryptAttachment` had the same problem for thumbnails whose mimetype the event does not name.

Per review feedback no extension is guessed: the thumbnail file name only gets one when the mimetype is actually known. All three places that name a thumbnail now share one helper — generating it in `Room.sendFileEvent`, restoring it from the cache, and downloading it (`file.mp4.thumbnail.jpg` for `image/jpeg`, plain `file.mp4.thumbnail` otherwise).

Dropping the extension is safe for consumers: `MatrixFile` derives the mimetype from the bytes' magic numbers when none is given, so the uploaded `Content-Type`, the `thumbnail_info.mimetype` of the event and clients rendering the decoded bytes are unaffected.

### Verification

- The `sendAgain` test covers re-sending a video with a cached thumbnail but no `thumbnail_info`, asserting the thumbnail is re-uploaded exactly once as `file.mp4.thumbnail` (no guessed extension, no `.null`).
- The attachment test now asserts the downloaded thumbnail name for both a mimetype without a known extension (`image.thumbnail`) and a known one (`image.thumbnail.jpg`).
- `dart analyze`, `dart format` clean; `event_test` and `room_test` pass.

<div><a href="https://cursor.com/agents/bc-7e8c8b3b-c66f-4bc5-a031-13cc1f17d4a1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7e8c8b3b-c66f-4bc5-a031-13cc1f17d4a1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -419,6 +419,14 @@ class Event extends MatrixEvent {
     room.client.onCancelSendEvent.add(eventId);
   }
 
+  /// The name a thumbnail of the file [filename] is stored under. Only gets a
+  /// file extension if the mimetype of the thumbnail is actually known,
+  /// instead of guessing one.
+  String _thumbnailFileName(String filename) {
+    final extension = extensionFromMime(thumbnailMimetype);
+    return '$filename.thumbnail${extension == null ? '' : '.$extension'}';
+  }
+
   Future<MatrixFile?> _getCachedFile({bool getThumbnail = false}) async {
     if (transactionId == null) return null;
 
@@ -431,7 +439,7 @@ class Event extends MatrixEvent {
       if (thumbnailBytes != null) {
         return MatrixImageFile(
           bytes: thumbnailBytes,
-          name: '$filename.thumbnail.${extensionFromMime(thumbnailMimetype)}',
+          name: _thumbnailFileName(filename),
           mimeType: thumbnailMimetype,
           width: thumbnailInfoMap.tryGet<int>('w'),
           height: thumbnailInfoMap.tryGet<int>('h'),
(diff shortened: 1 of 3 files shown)


lib/src/event.dart
```diff
@@ -430,7 +430,8 @@ class Event extends MatrixEvent {
       if (thumbnailBytes != null) {
         return MatrixImageFile(
           bytes: thumbnailBytes,
-          name: '$filename.thumbnail.${extensionFromMime(thumbnailMimetype)}',
+          name:
+              '$filename.thumbnail.${extensionFromMime(thumbnailMimetype) ?? 'jpg'}',
```
krille-chan: If we have no thumbnail we should not just guess?

yash-garg (approved):

PR #2421, opened 2026-07-24
refactor: always fetch member list from server for first megolm session

krille-chan:
This workarounds that synapse sometimes seem to send a wrong joined member count which is used to check if we have all participants already cached locally.

fixes https://famedly.atlassian.net/browse/CS-1780


td-famedly (commented):
can you add a test please? :D

td-famedly:
@cursor review

td-famedly (approved):
oki

PR #2429, opened 2026-08-03
fix: read ints back as ints from IndexedDB on the web

nikzen:
`IndexedDBBox` converts values with `dartify`, which reports every JavaScript number as a double, so a value written as `1` reads back as `1.0`. dart2js maps that onto the same Dart `int`; dart2wasm does not, and loading a stored session fails:

```
Type 'double' is not a subtype of type 'int' in type cast
    at Room.fromJson
    at MatrixSdkDatabase.getRoomList
```

Every read goes through `_fromValue`, so whole-valued doubles are turned back into ints there. Fractional values are untouched.

### DIFF
--- a/lib/src/database/indexeddb_box.dart
+++ b/lib/src/database/indexeddb_box.dart
@@ -408,7 +408,8 @@ class Box<V> {
     clearQuickAccessCache();
   }
 
-  V? _fromValue(Object? value) {
+  V? _fromValue(Object? rawValue) {
+    final value = _wholeNumbersAsInt(rawValue);
     if (value == null) return null;
     switch (V) {
       case const (List<dynamic>):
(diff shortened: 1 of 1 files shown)


lib/src/database/indexeddb_box.dart
```diff
@@ -423,4 +424,26 @@ class Box<V> {
...
+  ///
+  ///     Type 'double' is not a subtype of type 'int' in type cast
+  ///         at Room.fromJson
+  ///         at MatrixSdkDatabase.getRoomList
+  ///
+  /// A genuinely fractional value keeps its type; a whole-valued double stored
+  /// as a double is read back as an int, which every model in this SDK accepts
+  /// where it accepts a number at all.
+  static Object? _wholeNumbersAsInt(Object? value) {
+    if (value is double && value == value.roundToDouble()) return value.toInt();
```
krille-chan, line 442: This could actually break in a different case. Let's assume we have a Box<int> then this would fix that it always gives us a double. But If we have a Box<double> and we put something like a `2.0` in it, then it would return us an int, while we want a double.

Edit: Okay it gets type casted below into double again, but still I'd prefer a different fix for this

td-famedly:
will close for now

PR #2430, opened 2026-08-04
fix: defer SSSS default key change until after UIA

coder-with-a-bushido:
Refer https://github.com/famedly/app/pull/7321#discussion_r3704294650

### DIFF
--- a/lib/encryption/utils/bootstrap.dart
+++ b/lib/encryption/utils/bootstrap.dart
@@ -66,6 +66,9 @@ class Bootstrap {
   OpenSSSS? newSsssKey;
   ErrorResult? errorResult;
 
+  /// Only a key created by [newSsss] may become the new default key.
+  bool _newSsssKeyNeedsDefaultUpdate = false;
+
   Bootstrap({required this.encryption, this.onUpdate}) {
     final defaultKeyId = encryption.ssss.defaultKeyId;
     final hasValidDefaultKey =
@@ -230,6 +233,7 @@ class Bootstrap {
     try {
       Logs().v('Create key...');
       newSsssKey = await encryption.ssss.createKey(passphrase, name);
+      _newSsssKeyNeedsDefaultUpdate = true;
       if (oldSsssKeys != null) {
         final existingOldKeys = oldSsssKeys!;
         if (existingOldKeys.isNotEmpty) {
@@ -244,13 +248,6 @@ class Bootstrap {
           );
         }
       }
-      await encryption.ssss.setDefaultKeyId(newSsssKey!.keyId);
-      while (encryption.ssss.defaultKeyId != newSsssKey!.keyId) {
-        Logs().v(
-          'Waiting accountData to have the correct m.secret_storage.default_key',
-        );
-        await client.oneShotSync();
-      }
     } catch (e, s) {
       Logs().e('[Bootstrapping] Error trying to migrate old secrets', e, s);
       errorResult = ErrorResult(e, s);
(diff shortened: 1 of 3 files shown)


krille-chan (changes requested):
so if I reset my chat backup but then cancel while being in UIA it would already have wiped everything but not setup something new?
okay for me it just changes the default key but yeah that's something which should be done after device keys have been uploaded.
however the old key is still there and can be used to unlock the backup but yeah it's still not nice that the default key has been changedhowever what about making this more minimal: Just pushing another key is not that bad. The problem is that the default key changes before UIA completed. Just move this to after UIA completed? Then the whole PR should become much smaller

lib/encryption/utils/bootstrap.dart
```diff
@@ -230,27 +237,22 @@ class Bootstrap {
...
               existingOldKeys[encryption.ssss.defaultKeyId] ??
               existingOldKeys.values.first;
-          await encryption.ssss.migrateSecretsToKey(
-            primaryUnlockedKey: primaryUnlockedKey,
-            destinationKey: newSsssKey!,
-            candidateOldKeys: existingOldKeys,
-            stripKeys: true,
-          );
+          _pendingMigratedSecretTypes = await encryption.ssss
+              .migrateSecretsToKey(
```
krille-chan: This doesn't look like what I have suggested. i suggested that we wait for updating the default key id but this changes how secrets are migrated which is **after** the UIA step. So this looks totally out of scope to the commit message?

coder-with-a-bushido: Thank you, I will remove it. That's an old stash I pushed wrongly without the following test when I had to rush for an emergency. It's not a realistic case in our app (but **it is** in general) at the moment because:
1. We don't migrate secrets while resetting recovery key and
2. while migrating secrets to passphrase from recovery key (by manually calling `migrateSecretsToKey`), we don't require UIA as cross signing is not wiped there

```
test(
      'keep old recovery key on migrate when cross signing uia is cancelled',
      () async {
        final testClient = await getClient();
        try {
          await testClient.initCryptoIdentity(passphrase: 'migratefoxies');
          final oldKeyId = testClient.encryption!.ssss.defaultKeyId!;
          final defaultKey = testClient.encryption!.ssss.open();
          await defaultKey.unlock(passphrase: 'migratefoxies');
          await defaultKey.store('foxes', 'floof');

          final uiaSub = testClient.onUiaRequest.stream.listen(
            (uia) => uia.cancel(),
          );

          const path = '/client/v3/keys/device_signing/upload';
          final posts = FakeMatrixApi.currentApi!.api['POST']!;
          final original = posts[path];
          posts[path] = (_) => {
            'errcode': 'M_UNAUTHORIZED',
            'error': 'Authentication required',
            'session': 'bootstrap-migrate-uia-cancel-test',
            'flows': [
              {
                'stages': ['m.login.password'],
              },
            ],
            'params': <String, dynamic>{},
          };
          try {
            Bootstrap? bootstrap;
            bootstrap = testClient.encryption!.bootstrap(
              onUpdate: (bootstrap) async {
                if (bootstrap.state == BootstrapState.askWipeSsss) {
                  bootstrap.wipeSsss(false);
                } else if (bootstrap.state ==
                    BootstrapState.askUseExistingSsss) {
                  bootstrap.useExistingSsss(false);
                } else if (bootstrap.state == BootstrapState.askUnlockSsss) {
                  await bootstrap.oldSsssKeys![oldKeyId]!.unlock(
                    passphrase: 'migratefoxies',
                  );
                  bootstrap.unlockedSsss();
                } else if (bootstrap.state == BootstrapState.askNewSsss) {
                  await bootstrap.newSsss('newmigratefoxies');
                } else if (bootstrap.state ==
                    BootstrapState.askWipeCrossSigning) {
                  await bootstrap.wipeCrossSigning(true);
                } else if (bootstrap.state ==
                    BootstrapState.askSetupCrossSigning) {
                  await bootstrap.askSetupCrossSigning(
                    setupMasterKey: true,
                    setupSelfSigningKey: true,
                    setupUserSigningKey: true,
                  );
                }
              },
            );
            while (bootstrap.state != BootstrapState.error) {
              await Future.delayed(Duration(milliseconds: 50));
            }
            expect(bootstrap.state, BootstrapState.error);
          } finally {
            if (original != null) {
              posts[path] = original;
            } else {
              posts.remove(path);
            }
          }

          await uiaSub.cancel();

          expect(testClient.encryption!.ssss.defaultKeyId, oldKeyId);
          final oldKey = testClient.encryption!.ssss.open(oldKeyId);
          await oldKey.unlock(passphrase: 'migratefoxies');
          expect(await oldKey.getStored('foxes'), 'floof');
          final encrypted = testClient.accountData['foxes']!.content
              .tryGetMap<String, Object?>('encrypted');
          expect(encrypted?.containsKey(oldKeyId), isTrue);
        } finally {
          await testClient.dispose(closeDatabase: true);
        }
      },
      timeout: Timeout(Duration(minutes: 2)),
    );
```

lib/encryption/utils/bootstrap.dart
```diff
@@ -244,22 +248,35 @@ class Bootstrap {
           );
         }
       }
+    } catch (e, s) {
```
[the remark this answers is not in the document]
coder-with-a-bushido: Already clarified here - https://github.com/famedly/matrix-dart-sdk/pull/2430#discussion_r3729434906

krille-chan (approved):

PR #2431, opened 2026-08-04
fix: workaround that [JSAny.dartify] on wasm could turn an int into double

krille-chan:


lib/src/database/indexeddb_box.dart
```diff
@@ -416,6 +416,9 @@ class Box<V> {
       case const (Map<dynamic, dynamic>):
         return Map.unmodifiable(value as Map) as V;
       case const (int):
+        // Workaround that [JSAny.dartify] on wasm could turn an int into double
+        if (value is double) return value.round() as V;
```
td-famedly, line 420: so 2.3 would become 2? isn't that wrong?

krille-chan: it would only become 2 if it is a `Box<int>` so any other value than an int would crash here. However this can not happen, as it is a `Box<int>` only 2 can be inserted, but because of wasm the 2 could be come 2.0 and this is what I want to cover

td-famedly (approved):
lg, how do I check if the package is wasm compatible now? 🤔

PR #2433, opened 2026-08-04
feat: (BREAKING) add ability to finish partial crypto identity initialization to initCryptoIdentity

coder-with-a-bushido:
Needed for https://github.com/famedly/app/pull/7321#discussion_r3704899734

### DIFF
--- a/lib/encryption/utils/crypto_setup_extension.dart
+++ b/lib/encryption/utils/crypto_setup_extension.dart
@@ -5,20 +5,37 @@
 import 'dart:async';
 
 import '../../matrix.dart';
+import '../ssss.dart';
 import 'bootstrap.dart';
 
 extension CryptoSetupExtension on Client {
   /// Returns the current state of the crypto identity.
-  /// The crypto identity is `initialized` if key backup and cross signing
-  /// are correctly set up. You can initialize a new account by using
-  /// `Client.initCryptoIdentity()`.
+  ///
+  /// [keyBackupEnabled] and [crossSigningEnabled] report each half
+  /// independently. The crypto identity is `initialized` if both are set up.
+  /// You can initialize a new account by using `Client.initCryptoIdentity()`.
+  ///
+  /// When [initialized] is false but the account already holds part of an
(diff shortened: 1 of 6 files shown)


lib/encryption/utils/crypto_setup_extension.dart
```diff
@@ -138,6 +155,121 @@ extension CryptoSetupExtension on Client {
...
+  ///
+  /// [keyOrPassphrase] unlocks the existing secret storage key. [keyIdentifier]
+  /// selects a specific key when several exist.
+  ///
+  /// Throws [BootstrapBadStateException] when the identity cannot be completed
+  /// in place — there is no usable existing secret storage key, or the existing
+  /// secrets cannot be read with [keyOrPassphrase]. Callers should fall back to
+  /// [initCryptoIdentity], which regenerates everything and returns a new
+  /// recovery key, and should treat that as a destructive operation.
+  Future<void> completePartialCryptoIdentity(
```
krille-chan: > Unlike [initCryptoIdentity], the current
  /// secret storage key is kept, so the user's existing recovery key and
  /// passphrase stay valid and nothing is wiped. Only the missing pieces are
  /// provisioned.

We can achieve the same with `initCryptoIdentity()` already by just set:

```dart
    bool wipeSecureStorage = false,
    bool wipeKeyBackup = true, // or false if already initialized
    bool wipeCrossSigning = true, // or false if already initialized
```

I don't get from the documentation why we need another method here

coder-with-a-bushido: Setting `wipeSecureStorage` to false doesn't wipe the secrets, but `initCryptoIdentity()` still creates a new SSSS key by calling `useExistingSsss(false)`. This method reuses the existing key and creates only missing identity parts. I'll improve the documentation maybe?

coder-with-a-bushido: using initCryptoIdentity now

krille-chan (approved):
LGTM

PR #2436, opened 2026-08-05
feat(l10n): add displaynameOverride hook to MatrixLocalizations

ggurdin:
### What

Add a `displaynameOverride(userId)` hook to `MatrixLocalizations` so clients can localize specific users' display names in SDK-generated text.

### Why

Addresses pangeachat/client#7487. The client PR (pangeachat/client#8157) localizes the "Pangea Bot" / "Support" display names client-side, but SDK-generated strings — state event texts ("Pangea Bot joined the chat"), `Sender: message` prefixes in `calcLocalizedBodyFallback`, and hero-based room names — resolve names via `User.calcDisplayname(i18n:)` internally, with no client-side interception point.

### How

- `MatrixLocalizations.displaynameOverride(String userId)` — new virtual method, default `null` (no behavior change for existing implementations, including `MatrixDefaultLocalizations`).
- `User.calcDisplayname` returns the override when non-null, before the profile-name/localpart logic. This single point covers sender names, target names in member-event texts, sender prefixes, and DM hero names — every path that threads `i18n` through.

The client's `MatrixLocals` can then override it to return the localized bot/support names (client-side follow-up after this merges and the `matrix` dep is refreshed).

### Testing

- New test: `calcDisplayname with displaynameOverride` in `test/user_test.dart` — covers the override, the null-default fallback, and a membership state event localizing to "Localized Bot joined the chat".
- `dart analyze lib test/user_test.dart`: no issues.
- `dart test test/user_test.dart test/matrix_localizations_test.dart test/event_test.dart test/room_test.dart`: all pass except 5 pre-existing failures (`encrypted attachments`, `downloadAndDecryptAttachment`, `calcEncryptionHealthState`, `pushRuleState`, `setPushRuleState`) which fail identically on unmodified `main` in this environment (flutter_rust_bridge/vodozemac not initialized locally).

### Deploy Notes

None — no-op until a client implements the override.


ggurdin:
Apologies — this was opened against the wrong repository (intended for the pangeachat fork). Closing.

PR #2437, opened 2026-08-06
chore: auto request profiles for leave users

krille-chan:
fixes https://famedly.atlassian.net/browse/CS-1848

tested with Famedly.


td-famedly (approved):
oki

krille-chan:
Actually the problem is, that displayname & avatar is no longer found in the prev_content because in the sdk we are handling prevContent probably not correct. I'll dive deeper into this

PR #2443, opened 2026-08-10
fix: refresh DM hero avatars in partial rooms

qaysdwekat:
Keep in-memory member state and requestUser in sync with avatar updates that previously only landed in the DB for lazy-loaded rooms.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -2058,13 +2058,10 @@ class Room {
     required bool requestProfile,
   }) async {
     // Is user already in cache?
-
-    // If not in cache, try the database
     var foundUser = getState(EventTypes.RoomMember, mxID)?.asUser(this);
 
-    // If the room is not postloaded, check the database
-    if (partial && foundUser == null) {
-      foundUser = await client.database.getUser(mxID, this);
+    if (partial) {
+      foundUser = await client.database.getUser(mxID, this) ?? foundUser;
     }
 
     // If not in the database, try fetching the member from the server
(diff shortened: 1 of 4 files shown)


yurtemre7 (approved):

ab-famedly (approved):

lib/src/user.dart
```diff
@@ -246,13 +246,24 @@ String _hash(String s) =>
...
-    content: content,
-    typeKey: type,
-    senderId: senderId,
-    room: room,
-    originServerTs: null,
-  );
+  User asUser(Room room) {
+    // Keep prevContent / originServerTs when the state is already a User or
+    // Event. Dropping prevContent breaks leave/ban displayname and avatar_url
+    // getters that fall back to prev_content.
```
krille-chan: Code comments are a bad pattern: https://www.youtube.com/watch?v=Bf7vDBBOBUA

lib/src/room.dart
```diff
@@ -2058,13 +2058,15 @@ class Room {
...
     var foundUser = getState(EventTypes.RoomMember, mxID)?.asUser(this);
 
-    // If the room is not postloaded, check the database
-    if (partial && foundUser == null) {
-      foundUser = await client.database.getUser(mxID, this);
+    // Partial rooms write member updates to the DB without always refreshing
+    // in-memory state. Prefer the DB so heroes do not stick with a stale
+    // avatar_url / displayname. Use asUser carefully: it must preserve
+    // prevContent so leave/ban getters do not disagree with the DB user and
+    // trigger setState/onUpdate on every call.
```
krille-chan: Same here

test/room_test.dart
```diff
@@ -973,6 +973,218 @@ void main() {
...
+      await Future.delayed(Duration(milliseconds: 1));
+      expect(
+        called.length,
+        updatesAfterFirst,
+        reason:
+            'requestUser must not re-emit onUpdate when leave profile lives in prev_content',
+      );
+
+      await subscription.cancel();
+    });
```
krille-chan, line 1186: I really don't like the direction this evolves into. Has anyone ever read this 212 LoC of tests here? Just ai generated and trusted? Are we sure we can't test the same (or a sufficient part of) logic with just 20 LoC or so? Test code should have the same quality like all other code

krille-chan (approved):

PR #2447, opened 2026-08-12
refactor: (BREAKING) make calcLocalizedBody sync again

krille-chan:
closes https://famedly.atlassian.net/browse/FM-690

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -985,8 +985,7 @@ class Event extends MatrixEvent {
   bool get isEventTypeKnown =>
       EventLocalizations.localizationsMap.containsKey(type);
 
-  /// Returns a localized String representation of this event. For a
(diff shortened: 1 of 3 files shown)


td-famedly (changes requested):
hm I thought we only rename it, but now we do not have a async function which also fetches the user. Do we not need this anymore? can you elaborate why?

krille-chan:
> hm I thought we only rename it, but now we do not have a async function which also fetches the user. Do we not need this anymore? can you elaborate why?

The method was only async when we needed the senderNamePrefix feature (which IMO belongs to the app anyway). By just removing this feature we no longer need this to be async at all. Only for redacted messages and for this edge case IMO metnioning it in the dart docs is better than forcing the sdk consumer to handle all the hassle with an async method here.

PR #2448, opened 2026-08-12
Krille/refactor device keys list storing

krille-chan:
> [!NOTE]
> **High Risk**
> Touches E2EE device/cross-signing key persistence, verification/blocked flags, and a schema migration (DB v12). A bad migrate or serialize path can lose trust state or break encryption.
> 
> **Overview**
> Replaces per-device / per-cross-signing Hive boxes with a single `storeDeviceKeysList` / `getDeviceKeysList` API so a user's full `DeviceKeysList` is written as one JSON document.
> 
> `updateUserDeviceKeys` now batches one persist per user after applying `/keys/query` results (including dropping unused devices and old cross-signing keys of the same usage). Verify/block/TOFU updates rewrite the whole list via `_updateInDatabase`. Last-sent Olm messages move to their own box.
> 
> Schema version is **12**. Existing stores migrate from the three legacy boxes (and last-sent fields) into the new layout, then clear the old boxes. Dump/import and tests follow the new shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 519ebfe76807e4f2fbe0fe5fc85348e2153c7db9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/src/database/matrix_sdk_database.dart
+++ b/lib/src/database/matrix_sdk_database.dart
@@ -83,6 +83,9 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
 
   late Box<Map> _deviceKeysListBox;
 
+  /// Is a Tuple(userId, deviceId) to the last sent message as a json map:
+  late Box<String> _lastSentOlmMessagesBox;
+
   @override
   final int maxFileSize;
 
@@ -146,6 +149,9 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   static const String _readReceiptsBoxName = 'box_read_receipts';
 
   static const String _deviceKeysListBoxName = 'box_device_keys_list';
+
+  static const String _lastSentOlmMessagesBoxName =
+      'box_last_sent_olm_messages';
 
   Database? database;
 
@@ -224,6 +230,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
         _userProfilesBoxName,
         _readReceiptsBoxName,
         _deviceKeysListBoxName,
+        _lastSentOlmMessagesBoxName,
       },
       sqfliteDatabase: database,
       sqfliteFactory: sqfliteFactory,
@@ -257,6 +264,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     _userProfilesBox = _collection.openBox(_userProfilesBoxName);
     _readReceiptsBox = _collection.openBox(_readReceiptsBoxName);
     _deviceKeysListBox = _collection.openBox(_deviceKeysListBoxName);
+    _lastSentOlmMessagesBox = _collection.openBox(_lastSentOlmMessagesBoxName);
 
     // Check version and check if we need a migration
     final currentVersion = int.tryParse(await _clientBox.get('version') ?? '');
@@ -334,6 +342,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     _userProfilesBox.clearQuickAccessCache();
     _readReceiptsBox.clearQuickAccessCache();
     _deviceKeysListBox.clearQuickAccessCache();
+    _lastSentOlmMessagesBox.clearQuickAccessCache();
 
     await _collection.clear();
   }
@@ -514,11 +523,10 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     String userId,
     String deviceId,
   ) async {
-    final rawKeys = await getDeviceKeysList(
-      userId,
-      Client('db', database: this),
+    final messageStr = await _lastSentOlmMessagesBox.get(
+      TupleKey(userId, deviceId).toString(),
     );
-    return [?rawKeys?.deviceKeys[deviceId]?.lastSentMessage];
+    return [?messageStr];
   }
 
   @override
@@ -728,7 +736,8 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     return Event.fromJson(copyMap(state), room).asUser;
   }
 
-  /// Only for migration. Can be removed after a certain amount of time.
+  /// Only for migration. Can be removed after a certain amount of time. Also
+  /// clears everything from the old boxes!
   Future<Map<String, DeviceKeysList>> _legacyGetUserDeviceKeys(
     Client client,
   ) async {
@@ -751,6 +760,14 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     final userDeviceKeys = await legacyUserDeviceKeysBox.getAllValues();
     final userCrossSigningKeys = await legacyUserCrossSigningKeysBox
         .getAllValues();
+
+    // Migrate lastSentMessages
+    for (final entry in userDeviceKeys.entries) {
+      final lastSentMessage = entry.value['last_sent_message'];
+      if (lastSentMessage is! String || lastSentMessage.isEmpty) continue;
+      await _lastSentOlmMessagesBox.put(entry.key, lastSentMessage);
+    }
+
     for (final userId in deviceKeysOutdated.keys) {
       final deviceKeysBoxKeys = userDeviceKeys.keys.where((tuple) {
         final tupleKey = TupleKey.fromString(tuple);
@@ -787,6 +804,9 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
         client,
       );
     }
+    legacyUserDeviceKeysBox.clear();
+    legacyUserCrossSigningKeysBox.clear();
+    legacyUserDeviceKeysOutdatedBox.clear();
     return res;
   }
 
@@ -929,49 +949,6 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   }
 
   @override
-  Future<void> removeUserCrossSigningKey(
-    String userId,
-    String publicKey,
-  ) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.crossSigningKeys.containsKey(publicKey)) return;
-    keys.crossSigningKeys.remove(publicKey);
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> removeUserDeviceKey(String userId, String deviceId) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.deviceKeys.containsKey(deviceId)) return;
-    keys.deviceKeys.remove(deviceId);
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> setBlockedUserCrossSigningKey(
-    bool blocked,
-    String userId,
-    String publicKey,
-  ) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.crossSigningKeys.containsKey(publicKey)) return;
-    keys.crossSigningKeys[publicKey]?.blocked = blocked;
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> setBlockedUserDeviceKey(
-    bool blocked,
-    String userId,
-    String deviceId,
-  ) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.deviceKeys.containsKey(deviceId)) return;
-    keys.deviceKeys[deviceId]?.blocked = blocked;
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
   Future<void> setLastActiveUserDeviceKey(
     int lastActive,
     String userId,
@@ -990,12 +967,10 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     String lastSentMessage,
     String userId,
     String deviceId,
-  ) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.deviceKeys.containsKey(deviceId)) return;
-    keys.deviceKeys[deviceId]?.lastSentMessage = lastSentMessage;
-    await storeDeviceKeysList(userId, keys);
-  }
+  ) => _lastSentOlmMessagesBox.put(
+    TupleKey(userId, deviceId).toString(),
+    lastSentMessage,
+  );
 
   @override
   Future<void> setRoomPrevBatch(
@@ -1009,37 +984,6 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     room.prev_batch = prevBatch;
     await _roomsBox.put(roomId, room.toJson());
     return;
-  }
-
-  @override
-  Future<void> setVerifiedUserCrossSigningKey(
-    bool verified,
-    String userId,
-    String publicKey, {
-    DateTime? trustOnFirstUseSince,
-  }) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.crossSigningKeys.containsKey(publicKey)) return;
-    keys.crossSigningKeys[publicKey]?.setDirectVerified(verified);
-    if (trustOnFirstUseSince != null) {
-      keys.crossSigningKeys[publicKey]?.trustOnFirstUse(
-        since: trustOnFirstUseSince,
-        updateInDatabase: false,
-      );
-    }
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> setVerifiedUserDeviceKey(
-    bool verified,
-    String userId,
-    String deviceId,
-  ) async {
-    final keys = await getDeviceKeysList(userId, Client('db', database: this));
-    if (keys == null || !keys.deviceKeys.containsKey(deviceId)) return;
-    keys.deviceKeys[deviceId]?.setDirectVerified(verified);
-    await storeDeviceKeysList(userId, keys);
   }
 
   @override
@@ -1344,67 +1288,6 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   }
 
   @override
-  Future<void> storeUserCrossSigningKey(
-    String userId,
-    String publicKey,
-    String content,
-    bool verified,
-    bool blocked, {
-    DateTime? trustOnFirstUseSince,
-  }) async {
-    final tmpClient = Client('db', database: this);
-    final keys =
-        await getDeviceKeysList(userId, tmpClient) ??
-        DeviceKeysList(userId, tmpClient);
-    keys.crossSigningKeys[publicKey] = CrossSigningKey.fromDbJson({
-      'user_id': userId,
-      'public_key': publicKey,
-      'content': content,
-      'verified': verified,
-      'blocked': blocked,
-      if (trustOnFirstUseSince != null)
-        'tofu': trustOnFirstUseSince.millisecondsSinceEpoch,
-    }, tmpClient);
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> storeUserDeviceKey(
-    String userId,
-    String deviceId,
-    String content,
-    bool verified,
-    bool blocked,
-    int lastActive,
-  ) async {
-    final tmpClient = Client('db', database: this);
-    final keys =
-        await getDeviceKeysList(userId, tmpClient) ??
-        DeviceKeysList(userId, tmpClient);
-
-    keys.deviceKeys[deviceId] = DeviceKeys.fromDb({
-      'user_id': userId,
-      'device_id': deviceId,
-      'content': content,
-      'verified': verified,
-      'blocked': blocked,
-      'last_active': lastActive,
-      'last_sent_message': keys.deviceKeys[deviceId]?.lastSentMessage ?? '',
-    }, tmpClient);
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
-  Future<void> storeUserDeviceKeysInfo(String userId, bool outdated) async {
-    final tmpClient = Client('db', database: this);
-    final keys =
-        await getDeviceKeysList(userId, tmpClient) ??
-        DeviceKeysList(userId, tmpClient);
-    keys.outdated = outdated;
-    await storeDeviceKeysList(userId, keys);
-  }
-
-  @override
   Future<void> transaction(Future<void> Function() action) =>
       _collection.transaction(action);
 
@@ -1565,6 +1448,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
           .getAllValues(),
       _olmSessionsBoxName: await _olmSessionsBox.getAllValues(),
       _deviceKeysListBoxName: await _deviceKeysListBox.getAllValues(),
+      _lastSentOlmMessagesBoxName: await _lastSentOlmMessagesBox.getAllValues(),
       _ssssCacheBoxName: await _ssssCacheBox.getAllValues(),
       _presencesBoxName: await _presencesBox.getAllValues(),
       _timelineFragmentsBoxName: await _timelineFragmentsBox.getAllValues(),
@@ -1636,6 +1520,12 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
       }
       for (final key in json[_deviceKeysListBoxName]!.keys) {
         await _deviceKeysListBox.put(key, json[_deviceKeysListBoxName]![key]);
+      }
+      for (final key in json[_lastSentOlmMessagesBoxName]!.keys) {
+        await _lastSentOlmMessagesBox.put(
+          key,
+          json[_lastSentOlmMessagesBoxName]![key],
+        );
       }
       for (final key in json[_ssssCacheBoxName]!.keys) {
         await _ssssCacheBox.put(key, json[_ssssCacheBoxName]![key]);
--- a/lib/src/database/matrix_sdk_database.dart
+++ b/lib/src/database/matrix_sdk_database.dart
@@ -34,7 +34,7 @@ import 'sqflite_box.dart' if (dart.library.js_interop) 'indexeddb_box.dart';
 /// Learn more at:
 /// https://github.com/famedly/matrix-dart-sdk/issues/1642#issuecomment-1865827227
 class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
-  static const int version = 11;
+  static const int version = 12;
   final String name;
 
   late BoxCollection _collection;
@@ -62,15 +62,6 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   late Box<String> _inboundGroupSessionsUploadQueueBox;
   late Box<Map> _outboundGroupSessionsBox;
   late Box<Map> _olmSessionsBox;
-
-  /// Key is a tuple as TupleKey(userId, deviceId)
-  late Box<Map> _userDeviceKeysBox;
-
-  /// Key is the user ID as a String
-  late Box<bool> _userDeviceKeysOutdatedBox;
-
-  /// Key is a tuple as TupleKey(userId, publicKey)
-  late Box<Map> _userCrossSigningKeysBox;
   late Box<Map> _ssssCacheBox;
   late Box<Map> _presencesBox;
 
@@ -89,6 +80,8 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   late Box<Map> _userProfilesBox;
 
   late Box<Map> _readReceiptsBox;
+
+  late Box<Map> _deviceKeysListBox;
 
   @override
   final int maxFileSize;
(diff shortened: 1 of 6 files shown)


lib/src/database/matrix_sdk_database.dart
```diff
@@ -1631,20 +1630,8 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
       for (final key in json[_olmSessionsBoxName]!.keys) {
         await _olmSessionsBox.put(key, json[_olmSessionsBoxName]![key]);
       }
-      for (final key in json[_userDeviceKeysBoxName]!.keys) {
-        await _userDeviceKeysBox.put(key, json[_userDeviceKeysBoxName]![key]);
```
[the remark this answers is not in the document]
krille-chan: Thats okay

lib/src/database/matrix_sdk_database.dart
```diff
@@ -1794,6 +1781,31 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
+    if (raw == null) return null;
+    final json = copyMap(raw);
+    return DeviceKeysList.fromJson(json, client);
+  }
+
+  @override
+  Future<void> storeDeviceKeysList(
+    String userId,
+    DeviceKeysList deviceKeysList,
+  ) => _deviceKeysListBox.put(userId, deviceKeysList.toJson());
```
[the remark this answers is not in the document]
krille-chan: That was already the case before

lib/src/utils/device_keys_list.dart
```diff
@@ -122,6 +122,37 @@ class DeviceKeysList {
...
+        crossSigningKeys: (json['cross_signing_keys'] as Map).map(
+          (k, v) => MapEntry(
+            k as String,
+            CrossSigningKey.fromDbJson(
+              Map<String, dynamic>.from(v as Map),
+              client,
+            ),
+          ),
+        ),
+      );
```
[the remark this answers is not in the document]
krille-chan: fixed

lib/src/database/matrix_sdk_database.dart
```diff
@@ -515,11 +514,11 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
     String userId,
     String deviceId,
   ) async {
-    final raw = await _userDeviceKeysBox.get(
-      TupleKey(userId, deviceId).toString(),
+    final rawKeys = await getDeviceKeysList(
+      userId,
+      Client('db', database: this),
```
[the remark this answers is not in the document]
krille-chan: I'm going to remove it right now

td-famedly:
@cursor review

yurtemre7:
The migration did not work correctly when I tested the app against this PR.

```sh
[Matrix] Migrate store database from version 11 to 12
[Matrix] Unable to build Matrix SDK Database! (attempt 1 of 3) - Exception: Illegal value type for Box: "dynamic"
  #2  MatrixSdkDatabase._legacyGetUserDeviceKeys (matrix_sdk_database.dart:744:49)
  #3  MatrixSdkDatabase._migrateFromVersion       (matrix_sdk_database.dart:284:37)
  #4  MatrixSdkDatabase.open                      (matrix_sdk_database.dart:274:13)
```

And the app crashed on startup, forcing the user to logout on the second app start.

yurtemre7 (dismissed):
fix the migration issue

lib/src/database/matrix_sdk_database.dart
```diff
@@ -281,6 +280,16 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
     Logs().i('Migrate store database from version $currentVersion to $version');
 
+    if (currentVersion <= 11) {
+      final deviceKeysLists = await _legacyGetUserDeviceKeys(
+        Client('migrationclient', database: this),
+      );
+      for (final entry in deviceKeysLists.entries) {
+        Logs().d('Migrate user keys', entry.key);
+        await storeDeviceKeysList(entry.key, entry.value);
+      }
```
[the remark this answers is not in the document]
krille-chan: This is bullshit. Vodozemac is always initiated first

yurtemre7:
What still needs to be looked at:

A crash that could happen in the middle of the migration would lose all verification state as it deleted the old storage boxes first, should be made more robust and clear the boxes only at the end.

lib/src/database/matrix_sdk_database.dart
```diff
@@ -281,6 +280,16 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   Future<void> _migrateFromVersion(int currentVersion) async {
     Logs().i('Migrate store database from version $currentVersion to $version');
 
+    if (currentVersion <= 11) {
+      final deviceKeysLists = await _legacyGetUserDeviceKeys(
+        Client('migrationclient', database: this),
```
yurtemre7, line 285: this client should be disposed, maybe also its inner httpClient too?

krille-chan: a client that isn't initialized doesn't need to be disposed I think. It's just a data structure where none of the streams ever fires something or is subscribed to. If we dispose it we could instead accidentally close the database

krille-chan:
> What still needs to be looked at:
> 
> A crash that could happen in the middle of the migration would lose all verification state as it deleted the old storage boxes first, should be made more robust and clear the boxes only at the end.

How? If it crashes we can't do anything anyway. We can only re-run it with the same data and let it then crash again. What else should we do? We can only make the migration code as robust as possible which is the case I think as it uses the same method like before to fetch the keys from the db and then a simple for loop to store them again. If you have ideas how to make it more robust, I'm open

lib/src/database/matrix_sdk_database.dart
```diff
@@ -281,6 +280,18 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
+      );
+      for (final entry in deviceKeysLists.entries) {
+        Logs().d('Migrate user keys', entry.key);
+        await storeDeviceKeysList(entry.key, entry.value);
+      }
+      await _clientBox.put('version', version.toString());
+      return;
+    }
+
     if (version == 8) {
```
td-famedly: now that I see it, this should be currentVersion

lib/src/database/matrix_sdk_database.dart
```diff
@@ -281,6 +280,18 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
+    if (currentVersion <= 11) {
+      final deviceKeysLists = await _legacyGetUserDeviceKeys(
+        Client('migrationclient', database: this),
+      );
+      for (final entry in deviceKeysLists.entries) {
+        Logs().d('Migrate user keys', entry.key);
+        await storeDeviceKeysList(entry.key, entry.value);
+      }
+      await _clientBox.put('version', version.toString());
+      return;
```
td-famedly, line 292: for starters can we have a test with tests migration from v11 to v12 please?

also the early return here looks wrong. We should migrate through each version iteratively

td-famedly (changes requested):
for starters can we have a test with tests migration from v11 to v12 please?

also looks like the default migration of clearCache when we upgrade a version now cannot be reached. Is this intentional?

lib/src/database/matrix_sdk_database.dart
krille-chan: We only migrate device keys. A clearCache does not affect device keys at all so that wouldn't be necessary and not touch the scope of the changes at all. That's why I added the return. I'll add a small test

krille-chan: We only migrate device keys. A clearCache does not affect device keys at all so that wouldn't be necessary and not touch the scope of the changes at all. That's why I added the return. I'll add a small test

krille-chan: Added a test case now :)

td-famedly (changes requested):
Looks like cursor's vod.init not called flag is correct?

We need to atleast have docs which say that vod.init should be done before databse opening. The test also is green when it shouldn't be 

Here https://github.com/famedly/matrix-dart-sdk/actions/runs/34963970408/job/104363948991#step:6:3177 you can see it flags that the keys are invalid which should not happen.

test/database_api_test.dart
```diff
@@ -721,5 +680,106 @@ void main() {
...
+
+        expect(deviceKeys?.userId, userId);
+        expect(deviceKeys?.outdated, true);
+        expect(deviceKeys?.deviceKeys, {});
+
+        expect(deviceKeys?.crossSigningKeys.keys, {publicKey});
+        final crossSigningKey = deviceKeys?.crossSigningKeys[publicKey];
+        expect(crossSigningKey?.usage, ['master']);
+        expect(crossSigningKey?.ed25519Key, publicKey);
+        await database.close();
```
td-famedly, line 812: we also need to check if the version now has successfully become 12 along with the new verification state

lib/src/database/matrix_sdk_database.dart
td-famedly: oops currentVersion 7 is inside the 8 if loop maybe we fix this

lib/src/database/matrix_sdk_database.dart
```diff
@@ -281,7 +280,7 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
   Future<void> _migrateFromVersion(int currentVersion) async {
     Logs().i('Migrate store database from version $currentVersion to $version');
 
-    if (version == 8) {
+    if (currentVersion == 8) {
```
td-famedly: This should be if version < 8 right? same for the version 7 below? or did we need this migration for version 9? 🤔

lib/src/database/matrix_sdk_database.dart
```diff
@@ -305,6 +304,18 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
 
+    if (currentVersion < 12) {
+      final deviceKeysLists = await _legacyGetUserDeviceKeys(
+        Client('migrationclient', database: this),
+      );
+      for (final entry in deviceKeysLists.entries) {
+        Logs().d('Migrate user keys', entry.key);
+        await storeDeviceKeysList(entry.key, entry.value);
+      }
+      await _clientBox.put('version', version.toString());
```
td-famedly, line 313: If we are migrating from 11 to 13, this will write 13 and therefore skip any migration we have setup for version 13? 

we should write the exact version and not just the latest version after every migration

td-famedly (changes requested):

lib/src/database/matrix_sdk_database.dart
```diff
@@ -305,6 +304,18 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
       }
     }
 
+    if (currentVersion < 12) {
```
td-famedly, line 303: The migration should be wrapped inside a transaction to make it safer

lib/src/database/matrix_sdk_database.dart
```diff
@@ -923,83 +950,29 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
-
   @override
   Future<void> setLastActiveUserDeviceKey(
     int lastActive,
     String userId,
     String deviceId,
   ) async {
-    final raw = copyMap(
-      await _userDeviceKeysBox.get(TupleKey(userId, deviceId).toString()) ?? {},
+    final keys = await getDeviceKeysList(userId, Client('db', database: this));
```
td-famedly: The client should be disposed manually. It's just memory and httpclient which are unneeded and we should dispose them

test/database_api_test.dart
krille-chan: good idea :) added

lib/src/database/matrix_sdk_database.dart
krille-chan: yes, < 8 is correct and then the currentVersion == 7 check below is also correct. Changed it :)

lib/src/database/matrix_sdk_database.dart
krille-chan: will make sense once the currentVersion check above is < 8 not == 8.

lib/src/database/matrix_sdk_database.dart
krille-chan: good catch. Changed it to `if (version == 12 && currentVersion == 11) return;`

lib/src/database/matrix_sdk_database.dart
krille-chan: We can do it. Done :)

lib/src/database/matrix_sdk_database.dart
krille-chan: Done

lib/src/database/matrix_sdk_database.dart
```diff
@@ -305,6 +304,20 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
       }
     }
 
+    if (currentVersion < 12) {
```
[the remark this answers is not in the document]
krille-chan: Fixed

lib/src/database/matrix_sdk_database.dart
```diff
@@ -923,83 +952,31 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
-    return;
-  }
-
-  @override
-  Future<void> setBlockedUserDeviceKey(
-    bool blocked,
-    String userId,
-    String deviceId,
-  ) async {
-    final raw = copyMap(
```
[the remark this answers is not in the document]
krille-chan: There are not concurrent calls like this and even if so that would have been a problem before as well

lib/src/database/matrix_sdk_database.dart
```diff
@@ -923,83 +948,31 @@ class MatrixSdkDatabase extends DatabaseApi with DatabaseFileStorage {
...
-
   @override
   Future<void> setLastActiveUserDeviceKey(
     int lastActive,
     String userId,
     String deviceId,
   ) async {
-    final raw = copyMap(
-      await _userDeviceKeysBox.get(TupleKey(userId, deviceId).toString()) ?? {},
+    final tmpClient = Client('db', database: this);
```
td-famedly, line 958: this tmp client is unused?

test/database_api_test.dart
```diff
@@ -721,5 +680,116 @@ void main() {
         expect(dump.isEmpty, true);
       });
     });
+    group('Database migrations', () {
+      test('Migrate from version 11', () async {
```
td-famedly, line 701: the ci test is still green even though it should be red here as pointed out in the comment as well https://github.com/famedly/matrix-dart-sdk/actions/runs/35584433198/job/106284473648#step:6:3195

td-famedly (changes requested):

doc/end-to-end-encryption.md
```diff
@@ -12,7 +12,7 @@ For Flutter you can use [flutter_vodozemac](https://pub.dev/packages/flutter_vod
 flutter pub add flutter_vodozemac
```
td-famedly, line 12: https://github.com/famedly/matrix-dart-sdk/pull/2482/

is another approach to this. The difference is how the data is stored

From your approach, 
`box_device_keys_list` is updated on  /keys/query, verify, block, TOFU, device_lists.changed, and every inbound olm message
`box_last_sent_olm_messages` is updated on every outbound olm message. 

I would rather split the first box into 5 different ones to make writes smaller. Everything is read into client.userDeviceKeys when we start the sdk, so we should first optimize the writes. And the split matters because the device keys are not written uniformly. (lastActive, outdated have different triggers)

I think the migration approach is also better because we can only copy and on fail we can retry copying again. We can do the delete later

https://github.com/famedly/matrix-dart-sdk/tree/techno/device-keys-bench has the benchmark btw

lib/src/utils/device_keys_list.dart
```diff
@@ -122,6 +122,37 @@ class DeviceKeysList {
...
+    'user_id': userId,
+    'outdated': outdated,
+    'device_keys': deviceKeys.map((k, v) => MapEntry(k, v.toDbJson())),
+    'cross_signing_keys': crossSigningKeys.map(
+      (k, v) => MapEntry(k, v.toDbJson()),
+    ),
+  };
+
+  factory DeviceKeysList.fromJson(Map<String, Object?> json, Client client) =>
+      DeviceKeysList(
```
td-famedly: we should have a if (!entry.isValid) throw Exception('Invalid device keys'); entry here?

td-famedly (changes requested):

lib/src/database/matrix_sdk_database.dart
krille-chan: ah thanks, fixed it

lib/src/utils/device_keys_list.dart
krille-chan: Done. I've not made it to throw because this could crash everything but filter them out and add a warning to the logs

doc/end-to-end-encryption.md
krille-chan: Thanks for your insights. For my approach I would argue with the following: We load **all** device keys at start into memory and update them from time to time. And loading them all at app start needs a very long time, this was actually the bottle neck with the performance problem I had in FluffyChat. It scales super bad having them splitted up to multiple boxes. After some weeks of usage, my matrix account needed 10s to start on Android because on every start we had to fetch stuff from 4 different boxes.

This changes it by just moving them into one box where we don't have a large logic anymore to put them together. Instead we just do one `getAllValues()` and we are done. Now with a huge account even after 8 weeks still only needs up to 500ms to load and not more. That's why I said it is a game changer in performance.

What you suggest is if I understand correctly and improvement in the performance of writing. However, writing was never the bottle neck. This is just fine how it is

test/database_api_test.dart
krille-chan: I can't see the logs there anymore in this job. Can you explain more what is the actual problem?

td-famedly: image

lib/src/utils/device_keys_list.dart
td-famedly: We now have it only for CrossSigningKey but we need it for DeviceKeys

doc/end-to-end-encryption.md
td-famedly: Correct, but I have managed to get both read and write performance down. This branch also does not have Devicekeys validation as stated in the comment below.

The major time went away on your branch because DeviceKeysList.fromJson no longer calls isValid on device keys — it's not saved, it's deferred to first use of each device, and invalid keys now load instead of being quarantined with outdated = true. (The removeWhere you added covers cross-signing keys only, and CrossSigningKey.isValid doesn't check signatures.)

lib/src/utils/device_keys_list.dart
krille-chan: Good catch. I've added the same logic there to check it

test/database_api_test.dart
krille-chan: yes but that is not a problem. We do other tests in our unit tests with invalid device keys as well and here we are actually checking that we catch it. Also it should definitely not crash here because then an invalid device key would break the whole app while it should just be skipped

doc/end-to-end-encryption.md
krille-chan: > Correct, but I have managed to get both read and write performance down.

write performance is not the bottle neck and was planned for me to put it into a follow up PR. I just don't wanted to blow up the complexity here. 

> This branch also does not have Devicekeys validation as stated in the comment below.

But this was only a super small fix.

> The major time went away on your branch because DeviceKeysList.fromJson no longer calls isValid on device keys — it's not saved

No I'm pretty sure it is because we no longer clumch together information from 4 different boxes on start but instead just call one `getAllValues()` from only one box and we are done. This now still also does check validity of keys and this doesn't really affect the performance (while we could in future think about if we really need to check signature on every app start of all keys?)

However I don't see what arguments about what this PR does not cover are actually bringing to the table. With **trunk based development** we want small incremental changes, and this is exactly what I target here, a single change which does fix a major performance leak and does nothing else yet while settling the foundation for more. but this kind of development is very hard if the PR is open that long

td-famedly: I would suggest you do the performance testing again. .isValid check according to my benchmark adds the most overhead. We can easily fix this is we just store the validity in storeDeviceKeysList DeviceKeys.fromDb can then restore it and skip .isValid completely. And this will affect app restarts not fresh logins

(yes the fix because of less TupleKey.fromStrings does work, but .isValid is also taking a big chunk) 

Also writes are now slower because setLastActiveUserDeviceKey calls getDeviceKeysList and storeDeviceKeysList, for EVERY device of that user. This can be easily fixed (or is even better in main) because it has a different row for every device id.

lib/src/utils/device_keys_list.dart
```diff
@@ -169,6 +230,14 @@ abstract class SignableKey extends MatrixSignableKey {
   bool? _verified;
   bool? _blocked;
 
+  Future<void> _updateInDatabase() {
+    final deviceKeysList = client.userDeviceKeys[userId];
+    if (deviceKeysList == null) {
+      throw 'Unable to update device keys list of $userId in database because client.userDeviceKeys["$userId"] is null!';
```
td-famedly: Please throw a exception here not a string

test/database_api_test.dart
```diff
@@ -737,6 +695,116 @@ void main() {
...
+
+        // Check if version is now 12:
+        final version = await database.database!.query(
+          'box_client',
+          columns: ['v'],
+          where: 'k = ?',
+          whereArgs: ['version'],
+        );
+        expect(version.first['v'], '12');
+
```
td-famedly, line 811: This test passed even without the skipSignature flag which means the test was inaccurate. This is probably because of the invalid key which would have been dropped anyway. 


```
diff --git a/test/database_api_test.dart b/test/database_api_test.dart
index 833e7394..dee164fd 100644
--- a/test/database_api_test.dart
+++ b/test/database_api_test.dart
@@ -9,9 +9,18 @@ import 'dart:typed_data';
 import 'package:sqflite_common_ffi/sqflite_ffi.dart';
 import 'package:test/test.dart';
+import 'package:vodozemac/vodozemac.dart' as vod;
 
 import 'fake_database.dart';
 
+Future<void>? _vodInit;
+
+/// vodozemac may only be initialised once per isolate.
+Future<void> ensureVodozemac() => _vodInit ??= vod.init(
+  wasmPath: './pkg/',
+  libraryPath: './rust/target/debug/',
+);
+
 String createLargeString(String character, int desiredSize) {
   final buffer = StringBuffer();
 
@@ -698,7 +707,46 @@ void main() {
       });
     });
     group('Database migrations', () {
-      test('Migrate from version 11', () async {
+      const userId = '@test:fakeServer.notExisting';
+      const deviceId = 'OTHERDEVICE';
+      const publicKey = 'F9ypFzgbISXCzxQhhSnXMkc1vq12Luna3Nw5rqViOJY';
+      const lastActive = 1234567890;
+      const tofu = 1600000000000;
+
+      /// A genuinely self-signed device key. An invalid fixture would be
+      /// dropped by the read path no matter what the migration did, so the
+      /// assertions below would hold even for a broken migration.
+      final deviceKeysContent = jsonEncode({
+        'user_id': userId,
+        'device_id': deviceId,
+        'algorithms': [
+          AlgorithmTypes.olmV1Curve25519AesSha2,
+          AlgorithmTypes.megolmV1AesSha2,
+        ],
+        'keys': {
+          'curve25519:$deviceId': 'R96BA0qE1+QAWLp7E1jyWSTJ1VXMLpEdiM2SZHlKMXM',
+          'ed25519:$deviceId': 'EQo9eYbSygIbOR+tVJziqAY1NI6Gga+JQOVIqJe4mr4',
+        },
+        'signatures': {
+          userId: {
+            'ed25519:$deviceId':
+                '/rT6pVRypJWxGos1QcI7jHL9HwcA83nkHLHqMcRPeLSxXHh4oHWvC0/tl0Xg06ogyiGw4NuB7TpOISvJBdt7BA',
+            'ed25519:$publicKey':
+                'qnjiLl36h/1jlLvcAgt46Igaod2T9lOSnoSVkV0KC+c7vYIjG4QBzXpH+hycfufOT/y+a/kl52dUTLQWctMKCA',
+          },
+        },
+      });
+
+      final crossSigningKeyContent = jsonEncode({
+        'user_id': userId,
+        'usage': ['master'],
+        'keys': {'ed25519:$publicKey': publicKey},
+        'signatures': <String, Object?>{},
+      });
+
+      /// Builds a version 11 database holding one verified device key and one
+      /// verified cross signing key carrying a TOFU timestamp.
+      Future<Database> buildVersion11Database() async {
         final sqliteDb = await databaseFactoryFfi.openDatabase(
           ':memory:',
           options: OpenDatabaseOptions(singleInstance: false),
@@ -713,71 +761,73 @@ void main() {
             'CREATE TABLE IF NOT EXISTS $name (k TEXT PRIMARY KEY NOT NULL, v TEXT)',
           );
         }
-
         await sqliteDb.insert('box_client', {
           'k': 'version',
           'v': '11',
         }, conflictAlgorithm: .replace);
-
-        const userId = '@alice:example.com';
-        const deviceId = 'deviceId';
-        const publicKey = 'publicKey';
-
-        final deviceKeysContent = jsonEncode({
-          'user_id': userId,
-          'device_id': deviceId,
-          'algorithms': [
-            AlgorithmTypes.olmV1Curve25519AesSha2,
-            AlgorithmTypes.megolmV1AesSha2,
-          ],
-          'keys': {
-            'curve25519:$deviceId':
-                '3C5BFWi2Y8MaVvjM8M22DBmh24PmgR0nPvJOIArzgyI',
-            'ed25519:$deviceId': 'lEuiRJBit0IG6nUf5pUzWTUEsRVVe/HJkoKuEww9ULI',
-          },
-          'signatures': {
-            userId: {
-              'ed25519:$deviceId':
-                  'dSO80A01XiigH3uBiDVx/EjzaoycHcjq9lfQX0uWsqxl2giMIiSPR8a4d291W1ihKJL/a+myXS367WT6NAIcBA',
-            },
-          },
-        });
         await sqliteDb.insert('box_user_device_keys', {
           'k': TupleKey(userId, deviceId).toString(),
           'v': jsonEncode({
             'user_id': userId,
             'device_id': deviceId,
             'content': deviceKeysContent,
-            'verified': false,
+            'verified': true,
             'blocked': false,
-            'last_active': 0,
+            'last_active': lastActive,
           }),
         });
-
-        final crossSigningKeyContent = jsonEncode({
-          'user_id': userId,
-          'usage': ['master'],
-          'keys': {'ed25519:$publicKey': publicKey},
-          'signatures': <String, Object?>{},
-        });
         await sqliteDb.insert('box_cross_signing_keys', {
           'k': TupleKey(userId, publicKey).toString(),
           'v': jsonEncode({
             'user_id': userId,
             'public_key': publicKey,
             'content': crossSigningKeyContent,
-            'verified': false,
+            'verified': true,
             'blocked': false,
+            'tofu': tofu,
           }),
         });
-
         await sqliteDb.insert('box_user_device_keys_outdated', {
           'k': userId,
-          'v': 'true',
+          'v': 'false',
         });
+        return sqliteDb;
+      }
 
+      // Runs before the vodozemac test below, because vodozemac can only be
+      // initialised once per isolate and this case needs it absent.
+      test('Migrate from version 11 without vodozemac', () async {
+        expect(
+          vod.isInitialized(),
+          false,
+          reason: 'this test must run before vodozemac is initialised',
+        );
+        final sqliteDb = await buildVersion11Database();
+        final database = await MatrixSdkDatabase.init(
+          'unit_test.${DateTime.now().microsecondsSinceEpoch}',
+          database: sqliteDb,
+          sqfliteFactory: databaseFactoryFfi,
+        );
+
+        // Signature checks cannot pass without vodozemac, so the migration
+        // must copy the keys rather than validate them. Assert on what landed
+        // on disk, since the read path does still validate.
+        final rows = await sqliteDb.query('box_device_keys_list');
+        expect(rows, hasLength(1));
+        final stored = jsonDecode(rows.single['v'] as String) as Map;
+        expect((stored['device_keys'] as Map).keys, {
+          deviceId,
+        }, reason: 'the device key must survive a migration without vodozemac');
+        expect((stored['cross_signing_keys'] as Map).keys, {publicKey});
+
+        await database.close();
+      });
+
+      test('Migrate from version 11 keeps trust state', () async {
+        await ensureVodozemac();
+        final sqliteDb = await buildVersion11Database();
         final database = await MatrixSdkDatabase.init(
-          'unit_test.${DateTime.now().millisecondsSinceEpoch}',
+          'unit_test.${DateTime.now().microsecondsSinceEpoch}',
           database: sqliteDb,
           sqfliteFactory: databaseFactoryFfi,
         );
@@ -788,13 +838,26 @@ void main() {
         );
 
         expect(deviceKeys?.userId, userId);
-        expect(deviceKeys?.outdated, true);
-        expect(deviceKeys?.deviceKeys, {});
+        expect(deviceKeys?.outdated, false);
+
+        final deviceKey = deviceKeys?.deviceKeys[deviceId];
+        expect(deviceKey, isNotNull, reason: 'the device key must survive');
+        expect(deviceKey?.directVerified, true);
+        expect(deviceKey?.blocked, false);
+        expect(
+          deviceKey?.lastActive,
+          DateTime.fromMillisecondsSinceEpoch(lastActive),
+        );
 
-        expect(deviceKeys?.crossSigningKeys.keys, {publicKey});
         final crossSigningKey = deviceKeys?.crossSigningKeys[publicKey];
+        expect(crossSigningKey, isNotNull);
         expect(crossSigningKey?.usage, ['master']);
         expect(crossSigningKey?.ed25519Key, publicKey);
+        expect(crossSigningKey?.directVerified, true);
+        expect(
+          crossSigningKey?.trustOnFirstUseSince,
+          DateTime.fromMillisecondsSinceEpoch(tofu),
+        );
 
         // Check if version is now 12:
         final version = await database.database!.query(

```


Can you apply that patch please?

td-famedly (changes requested):

lib/src/utils/device_keys_list.dart
krille-chan: okay done

PR #2477, opened 2026-09-05
feat: add method to clear free database pages

krille-chan:
After a clearCache() the database file actually does not shrink in size but has empty pages. Without the auto vacuum pragma it can never really shrink. This adds an optional method the sdk consumer can call to fix it.


td-famedly:
@cursor review

ab-famedly (approved):

PR #2479, opened 2026-09-14
fix: prev_content handling

krille-chan:
Previously prev_content was by the spec defined to be an event's top level attribute. Since 1.2 it is defined as a field of unsigned. However Synapse has always put it into unsigned which lead to these wrokarounds in the sdk. Now it looks like these workarounds are not water proof and might have caused CS-1848 so that prev_content is not correctly read. This PR removes the custom prev_content state completely, reads it from unsigned only but makes sure that in constructor also the top level json is checked so we can never have a missed prev_content object.

Probably fixes https://famedly.atlassian.net/browse/CS-1848


lib/src/event.dart
```diff
@@ -190,18 +176,25 @@ class Event extends MatrixEvent {
   factory Event.fromJson(Map<String, dynamic> jsonPayload, Room room) {
     final content = Event.getMapFromPayload(jsonPayload['content']);
     final unsigned = Event.getMapFromPayload(jsonPayload['unsigned']);
-    final prevContent = Event.getMapFromPayload(jsonPayload['prev_content']);
     final originalSource = Event.getMapFromPayload(
       jsonPayload['original_source'],
     );
+
+    // Previous implementations have defined prev_content on the top level:
```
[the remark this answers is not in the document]
krille-chan: Comments are okay if they explain WHY something has to be done to mark e.g. workarounds

lib/src/event.dart
```diff
@@ -67,6 +67,9 @@ class Event extends MatrixEvent {
 
   String? get transactionId => unsigned?.tryGet<String>('transaction_id');
 
+  Map<String, Object?>? get prevContent =>
```
ab-famedly, line 70: is this a breaking change? it's the new public API surface

krille-chan: The API stays more or less the same. Okay you have no longer writing access to it, but I'd say that shouldn't harm. However we already have breaking changes on main so the next release will be a major release anyway so everything is fine :)

ab-famedly (approved):

td-famedly:
@cursor review

PR #2481, opened 2026-09-18
fix: correctly self sign

krille-chan:
We already selfsign when we do ssss.unlock() with postUnlock=true which is the default. So we always tried to selfSign TWICE which in edge cases crashes.


td-famedly:
@cursor review

